PR #4, opened 2015-03-27
Chunk Store implemented

ustulation:


Cargo.toml
```diff
@@ -9,3 +9,20 @@ repository = "https://github.com/dirvine/maidsafe_vault"
...
+[dependencies]
+rustc-serialize = "*"
+cbor = "*"
+bchannel = "*"
+lru-cache = "*"
+time = "*"
+
+[dependencies.maidsafe_types]
+git = "https://github.com/dirvine/maidsafe_types"
+
```
maqi: need to include routing as well

src/vault.rs
```diff
@@ -0,0 +1,16 @@
...
+    directory of this project at LICENSE, COPYING and CONTRIBUTOR respectively and also
+    available at: http://www.maidsafe.net/licenses
+    Unless required by applicable law or agreed to in writing, the MaidSafe Software distributed
+    under the GPL Licence is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS
+    OF ANY KIND, either express or implied.
+    See the Licences for the specific language governing permissions and limitations relating to
+    use of the MaidSafe
+    Software.                                                                 */
+
+
```
maqi: this shall be the place implement facade as in C++ code base vault.cc

src/chunk_store.rs
```diff
@@ -0,0 +1,145 @@
...
+    let mut return_val: Vec<u8> = Vec::new();
+
+    for it in self.entries.iter() {
+      if it.name == name {
+        let mut iv: Vec<u8> = Vec::with_capacity(AES256_IVSize);
+        let mut key: Vec<u8> = Vec::with_capacity(AES256_KeySize);
+
+        for it in name.iter().take(AES256_KeySize) {
+          key.push(*it);
+        }
```
maqi: probably you can do key.push_all(name.iter().take(key_size)))

src/vault.rs
```diff
@@ -0,0 +1,68 @@
...
+use std::default::Default;
+
+
+use pmid_node::PmidNode;
+use self::Routing::Authority;
+use self::Routing::DhtIdentity;
+use self::Routing::Action;
+use self::Routing::RoutingError;
+
+impl Routing::Facade for PmidNode {
```
maqi: I think it shall be something like this : 
struct Vault {
  pmid_node : PmidNode;
  data_manager : DataManager;
  ...
  routing_node : RoutingNode
}

impl Vault {
  fn new() {
     routing_node : RoutingNode { traits : {|| self.handle_get(_)}}
  }
  fn handle_get(&self, our_authority: Authority, from_authority: Authority, from_address: DhtIdentity, data: Vec<u8>)->Result<Action, RoutingError> {
    let get_result = pmid_node.get(data);
    Ok(Action::Reply(get_result))
  }
}

src/vault.rs
```diff
@@ -0,0 +1,68 @@
...
+extern crate bchannel;
+
+use std::net::{TcpStream};
+use sodiumoxide::crypto;
+use std::sync::mpsc;
+use std::sync::mpsc::{Sender, Receiver};
+use std::default::Default;
+
+
+use pmid_node::PmidNode;
```
maqi: as this implementation will also depends on the implementation of routing_node in routing,
I will suggest leave this vault.rs as empty to make this PR only includes the work of chunk_store

src/chunk_store.rs
ustulation: didn't work (atleast not out of the box)

Cargo.toml
ustulation: done

PR #8, opened 2015-03-28
link vault with routing

maqi:


src/lib.rs
```diff
@@ -14,15 +13,36 @@ extern crate cbor;
...
+  my_facade : VaultFacade
+}
+
+impl Vault {
+  pub fn new() -> Vault {
+    Vault { my_facade: VaultFacade::new() }
+  }
+
+  pub fn start_vault(&self) {
+    let my_routing = Some(routing::RoutingNode::new(&self.my_facade as & routing::Facade));
```
dirvine: I don't think we need the cast here.

src/lib.rs
```diff
@@ -14,15 +13,36 @@ extern crate cbor;
...
+    loop {
+      always_true();
```
dirvine: This is fine, routing will supply a start() method to start listening that will block till finish so this is fine for now.

dirvine:
Looks good Qi I am happy to merge if you are. I see nighlies are all broke again (rustc-serialise)

PR #13, opened 2015-04-01
implement PmidNode and PmidManager

maqi:


dirvine:
I will wait until the routing PR is updated (apart from build error it is fine so will get merged quickly).

PR #16, opened 2015-04-01
Chunk store test integrated

krishnaIndia:
Have implemented the initial utility functions and structured it. Will continue writing the helper functions and then the unit tests.


src/chunk_store/test.rs
```diff
@@ -0,0 +1,82 @@
+extern crate maidsafe_types;
+extern crate rand;
+
+use rand::{thread_rng, Rng};
+use std::str;
+use chunk_store::ChunkStore;
+use std::path::Path;
+use std::fs::{remove_dir_all};
+
+static ONE_KB: u64 = 1024u64;
```
dirvine: I think this can be just `static ONE_KB: u64 = 1024;`

src/chunk_store/test.rs
```diff
@@ -0,0 +1,82 @@
...
+static AES_PADDING: u64 = 16u64;
```
dirvine: We do no tneed to consider encrypting these chunks in the chunk-store any more. They will be session based so we can store them as they are (already encrypted anyway). Keeps it simpler. So this becomes literally a HashMap of <NameType, Data> . Should simplify it a lot :-)

krishnaIndia:
Updated the unit tests. Two more unit tests are pending. Unused code to be removed after the tests are wired up. Implemented `get` in chunk_store.rs for writing the tests.

Cargo.toml
```diff
@@ -25,3 +25,9 @@ git = "https://github.com/dirvine/maidsafe_types"
 
 [dependencies.routing]
 git = "https://github.com/dirvine/routing"
+
+[dependencies.self_encryption]
```
dirvine: I do not think we need self_encryption as a dependency here.

src/lib.rs
```diff
@@ -10,6 +10,10 @@ extern crate sodiumoxide;
 extern crate rustc_serialize;
 extern crate cbor;
 extern crate time;
+extern crate bchannel;
```
dirvine: Not sure bchannel is required either.

Cargo.toml
krishnaIndia: It was not compiling at my end. Will check one more time, removing the dependency

src/lib.rs
krishnaIndia: I might have messed it while merging. I ll check this along with the self_encryption dependency.

krishnaIndia:
Tests have been integrated. But there are three failing tests.

krishnaIndia:
The failing tests are fixed. All tests are passing for ChunkStore

dirvine:
Nice @krishnaIndia  Thanks

PR #17, opened 2015-04-01
Fixed bug and added 'partial' test, WIP.

BSMaidSafe:
Uncommenting the last assert in test results in panic. There's a TODO in DataManager w.r.t close_nodes that when fixed the test(s) can be revisited.

### DIFF
--- a/src/data_manager/data_manager.rs
+++ b/src/data_manager/data_manager.rs
@@ -22,7 +22,7 @@ mod database;
 
 use self::routing::types;
 
-use cbor::{ Decoder};
+use cbor::{ Decoder };
 
 type CloseGroupDifference = self::routing::types::CloseGroupDifference;
 type Address = self::routing::types::Address;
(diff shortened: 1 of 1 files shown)


src/data_manager/data_manager.rs
```diff
@@ -74,4 +74,41 @@ impl DataManager {
...
+    let mut encoder = cbor::Encoder::from_memory();
+
+    encoder.encode(&[&data]);
+
+    let result = data_manager.handle_put(&array_as_vector(encoder.as_bytes()));
+    assert_eq!(result.is_err(), false);
+
+    let data_name = array_as_vector(&data.get_name().get_id());
+    let result = data_manager.handle_get(&data_name);
+    // assert_eq!(result.is_err(), false); error no pmid nodes
```
dirvine: Cool, can you please add a //FIXME here and I will merge and we can pick it up when we get the pmid addresses from routing sorted proper)

PR #23, opened 2015-04-02
handle_churn and has_account

ustulation:


src/maid_manager/maid_manager.rs
```diff
@@ -50,4 +64,28 @@ impl MaidManager {
 
     Ok(routing::Action::SendOn(destinations))
   }
+
+  pub fn handle_churn(&mut self, close_group_difference: &CloseGroupDifference) {
```
dirvine: Is it possible to also add a test here @ustulation  just to not affect coverage too much.

PR #34, opened 2015-04-08
using DhtId in vault

maqi:


dirvine:
I think this is failing as it calls a routing::DhtIdentity can you please tie up with Peter to confirm the routing::type for this.

PR #38, opened 2015-04-15
catch up the changes in types and routing

maqi:


Cargo.toml
```diff
@@ -21,7 +21,7 @@ rand = "*"
 git="https://github.com/contain-rs/lru-cache.git"
 
 [dependencies.maidsafe_types]
-git = "https://github.com/dirvine/maidsafe_types"
+path = "../maidsafe_types"
 
 [dependencies.routing]
-git = "https://github.com/dirvine/routing"
+path = "../routing"
```
dirvine: Sorry @maqi I know why you are doing this but we cannot reference local repo's can you update with this part removed. I will take the merge hit :-)

maqi: sorry, this was committed in by mistake

PR #40, opened 2015-04-16
passing cargo test

maqi:
[Review on Reviewable](https://reviewable.io/reviews/dirvine/maidsafe_vault/40)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 
 name = "maidsafe_vault"
-version = "0.0.1"
+version = "0.0.3"
 authors = ["David Irvine <david.irvine@maidsafe.net>"]
 description = "This is Pre alpha, and not useful, no code worth looking at."
 documentation = "http://dirvine.github.io/maidsafe_vault/"
@@ -16,12 +16,6 @@ cbor = "*"
 time = "*"
 sodiumoxide = "*"
 rand = "*"
-
-[dependencies.lru_time_cache]
-git = "https://github.com/dirvine/lru_time_cache"
-
-[dependencies.maidsafe_types]
-git = "https://github.com/dirvine/maidsafe_types"
-
-[dependencies.routing]
-git = "https://github.com/dirvine/routing"
+lru_time_cache = "0.1.2"
+maidsafe_types = "0.0.9"
+routing = "0.0.9"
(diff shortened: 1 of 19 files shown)


dirvine:
<sup>**[src/data_manager/data_manager.rs, line 98 \[r1\]](https://reviewable.io:443/reviews/dirvine/maidsafe_vault/40#-JnV4YBAP7vG09bUHAc9)** ([raw file](https://github.com/dirvine/maidsafe_vault/blob/fd5b189a2e9263f656b6c5b056b544f0e6994a1a/src/data_manager/data_manager.rs#L98)):</sup>
Can we remove the comment

---

<sup>**[src/maid_manager/maid_manager.rs, line 72 \[r1\]](https://reviewable.io:443/reviews/dirvine/maidsafe_vault/40#-JnV4f7AQBtKZaCxXXxu)** ([raw file](https://github.com/dirvine/maidsafe_vault/blob/fd5b189a2e9263f656b6c5b056b544f0e6994a1a/src/maid_manager/maid_manager.rs#L72)):</sup>
This should not be here, instead `use routing` will work

---

<sup>**[src/maid_manager/maid_manager.rs, line 76 \[r1\]](https://reviewable.io:443/reviews/dirvine/maidsafe_vault/40#-JnV4loP7_4evYD1_CxD)** ([raw file](https://github.com/dirvine/maidsafe_vault/blob/fd5b189a2e9263f656b6c5b056b544f0e6994a1a/src/maid_manager/maid_manager.rs#L76)):</sup>
instead of super::* use these individually and not self::

---

<sup>**[src/vault.rs, line 138 \[r1\]](https://reviewable.io:443/reviews/dirvine/maidsafe_vault/40#-JnV4yHUcItrRIxgrNhu)** ([raw file](https://github.com/dirvine/maidsafe_vault/blob/fd5b189a2e9263f656b6c5b056b544f0e6994a1a/src/vault.rs#L138)):</sup>
  `use maidsafe_types::*;
  use maidsafe_types::traits::RoutingTrait;
  use maidsafe_types::traits::RoutingTrait;
  use routing::types::Authority;
  use routing::types::{Authority, DestinationAddress, DhtId, generate_random_vec_u8};`
is fine here no need for super:: unless we want a specific item from super I believe

---


---

Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/dirvine/maidsafe_vault/40)

PR #43, opened 2015-04-20
update routing version

rossmuir:


dirvine:
Have you checked there is no breaking changes here for other libs @rossmuir ?

rossmuir:
I am doing it manually and also want to see what Travis and Appveyor do so please do not merge - should have stated that sorry

dirvine:
No worries, just don't want to get into a cycle of upgrades when we can and instead upgrade when we must or when it makes sense. The locked API's make this much easier and we don't get into release cycle cycles of breakage if you see what I mean.

PR #44, opened 2015-04-22
Documentation

benjaminbollen:
Simple cleanup.  Added license. Added minimal documentation. Added #![deny(missing_docs)].

[Review on Reviewable](https://reviewable.io/reviews/dirvine/maidsafe_vault/44)

### DIFF
--- a/src/chunk_store/chunk_store.rs
+++ b/src/chunk_store/chunk_store.rs
@@ -14,14 +14,18 @@
     Software.                                                                 */
 
 #![allow(dead_code)]
+#![deny(missing_docs)]
 
 use routing::types::DhtId;
 
+/// Data entry in the chunkstore network name and data.
 pub struct Entry {
     name: DhtId,
     data: Vec<u8>
 }
 
+/// Chunkstore is a collection for holding all data chunks.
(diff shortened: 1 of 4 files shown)


dirvine:
<sup>**[src/vault.rs, line 16 \[r2\]](https://reviewable.io:443/reviews/dirvine/maidsafe_vault/44#-JnX6V91GpswEe_jGoPS)** ([raw file](https://github.com/dirvine/maidsafe_vault/blob/ca6e1ae94d1d44c63404ac79982823d5c3eb98a5/src/vault.rs#L16)):</sup>
This should really be removed as well. We do not want to have either dead code or unused_variables allowed. So any sign of those should be purged now

---


---

Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/dirvine/maidsafe_vault/44)

dirvine:
Reviewed files:
- src/chunk_store/chunk_store.rs @ r2
- src/main.rs @ r2
- src/vault.rs @ r2

---





Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/dirvine/maidsafe_vault/44)

PR #64, opened 2015-05-08
Reflects changes in routing Error types

inetic:
Should be merged after [this PR](https://github.com/dirvine/routing/pull/232/files) is merged. Note, some of the code could be simplified if we had the `fail!` macro mentioned [here](http://lucumr.pocoo.org/2014/11/6/error-handling-in-rust/) (it's not in the standard yet). Then we could write e.g. `fail!(ResponseError::NoData)` instead of `return Err(From::from(ResponseError::NoData))`.

[Review on Reviewable](https://reviewable.io/reviews/dirvine/maidsafe_vault/64)

### DIFF
--- a/src/data_manager/mod.rs
+++ b/src/data_manager/mod.rs
@@ -25,6 +25,7 @@ use routing::NameType;
 use maidsafe_types;
 use cbor::{ Decoder };
 use routing::sendable::Sendable;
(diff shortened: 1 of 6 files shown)


dirvine:
Thanks Peter. Are you going to take a look at maidsafe_client as well with the same?

PR #66, opened 2015-05-14
Windows installer fix

krishnaIndia:
Windows installer batch will build --release and then invoke the advanced installer. The icon dependency is added to the installer folder itself


Viv-Rajkumar:
NOTE: Using this PR to test highfive integration. Do not merge.

PR #68, opened 2015-05-14
Windows Installer Fix

krishnaIndia:
Windows installer batch will build --release and then invoke the advanced installer. The icon dependency is added to the installer folder itself


maidsafe-highfive:
r? @fraser999

(rust_highfive has picked a reviewer for you, use r? to override)

dirvine:
Nice @krishnaIndia

PR #72, opened 2015-05-22
expanding vault network test

maqi:
[Review on Reviewable](https://reviewable.io/reviews/maidsafe/maidsafe_vault/72)

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -73,7 +73,7 @@ fn main () {
     let copied_vault = mutate_vault.clone();
     let thread_guard = spawn(move || {
         loop {
-            thread::sleep_ms(10);
+            thread::sleep_ms(1);
             let _ = copied_vault.lock().unwrap().routing_node.run();
         }
     });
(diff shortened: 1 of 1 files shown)


maidsafe-highfive:
r? @ustulation

dirvine:
At the moment the location of the exe maidsafe_vault is not picked up in CI. I suspect this is due to CI not doing `cargo build` before test. I will reallocate to Ross to confirm. 

---

Review status: all files reviewed, 1 unresolved discussion, some commit checks failed.
Reviewed files:
- src/main.rs @ r1

---

<sup>**[src/main.rs, line 136 \[r1\]](https://reviewable.io:443/reviews/maidsafe/maidsafe_vault/72#-JpwCtVPekcVaZCYoY5M)** ([raw file](https://github.com/maidsafe/maidsafe_vault/blob/e40a1c795d8a11cef226007e780be63af2573b9f/src/main.rs#L136)):</sup>
Very nice, when the swarm issues fix in routing we can probably remove all of these sleeps actually. Well I hope so.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/maidsafe/maidsafe_vault/72)

benjaminbollen:
Nice one Qi, would be great to have the vault exec built and run as part of the CI.  We should then do the same for CRUST and routing example.

PR #81, opened 2015-05-29
Maid-1010 put flow documentation

maqi:
put flow documentation involving Primary, Backup and Sacrificial copies of data

[Review on Reviewable](https://reviewable.io/reviews/maidsafe/maidsafe_vault/81)


maidsafe-highfive:
r? @ustulation

benjaminbollen:
Looks good to me.

Viv-Rajkumar:
Just waiting on @inetic to confirm these are fine before merging this in.

inetic:
I would like to have a link somewhere that explains the language: *->, *->>, {Delete.Sy, So}... Some of it can be guessed, but I would prefer not to.

The doc also seems to suggest that more than one sacrificial copy can be deleted and let upper layers know about. From looking at the code, this is not the case.

maqi:
here is the link for the conventions :
https://github.com/maidsafe-archive/MaidSafe-Vault/blob/next/docs/conventions.md
I will move it into rust code base

regarding the "removal of multiple sacrificial copies", the documentation is aiming for the 'proper' flow,
which may incur some work in next iteration.
The current implementation is limited by the current routing API, so has to be only one copy.

inetic:
Thanks @maqi, looks OK to me then. Sorry for making this longer.

PR #101, opened 2015-06-15
run first vault node in membrane mode

maqi:
[Review on Reviewable](https://reviewable.io/reviews/maidsafe/maidsafe_vault/101)

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -26,8 +26,11 @@
 //!
 //! The resulting executable is the Vault node for the SAFE network.
 //! Refer to https://github.com/dirvine/maidsafe_vault
-#![feature(std_misc)]
+#![feature(convert, core, std_misc)]
 
+extern crate core;
+extern crate crust;
+extern crate docopt;
 extern crate sodiumoxide;
 extern crate rustc_serialize;
 extern crate cbor;
@@ -37,8 +40,14 @@ extern crate routing;
 extern crate maidsafe_types;
 extern crate rand;
 
+use core::iter::FromIterator;
+use std::net::SocketAddr;
+use std::str::FromStr;
 use std::thread;
 use std::thread::spawn;
+
+use docopt::Docopt;
+use rustc_serialize::{Decodable, Decoder};
 
 mod data_manager;
 mod maid_manager;
@@ -49,6 +58,7 @@ mod pmid_node;
 mod vault;
 mod utils;
 
+use crust::Endpoint;
 use vault::{ VaultFacade, VaultGenerator };
 
 /// Placeholder doc test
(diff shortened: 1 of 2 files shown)


maidsafe-highfive:
r? @chandraprakash

maqi:
failing in  Travis is because the executable can not be found , please make sure the executable has been built before the test.
failing in OSX is probably due to the test has been executed in parallel with other tests.

rossmuir:
ok I will have a look @maqi

chandraprakash:
As we are not going for first node option with routing as discussed in the meeting, this PR will not be required.

src/main.rs
```diff
@@ -67,10 +77,74 @@ impl Vault {
...
+  The optional <peer>... arg(s) are a list of peer endpoints (other running vaults).
+  If these are supplied, the node will try to connect to one of these in order to
+  join the network.  If no endpoints are supplied, the node will try to connect to
+  an existing network using Crust's discovery protocol.
+";
+
+#[derive(RustcDecodable, Debug)]
+struct Args {
+    arg_peer: Vec<PeerEndpoint>,
+    flag_node: bool,
```
chandraprakash: --node option is not required.

benjaminbollen:
do the tests not need to get updated as well, to have a first routing_node start as "RoutingNode::run_zero_node"?

chandraprakash:
@benjaminbollen  Its already done here https://github.com/maidsafe/maidsafe_vault/pull/101/files#diff-639fbc4ef05b315af92b4d836c31b023R209

src/main.rs
```diff
@@ -119,9 +205,17 @@ mod test {
...
-        for i in 0..num_of_nodes {
+        // the first vault must be run in zero_membrane mode
+        println!("---------- starting node 0 --------------");
+        processes.push(match Command::new("./target/debug/maidsafe_vault").arg("-f").stdout(Stdio::piped()).spawn() {
+                    Err(why) => panic!("couldn't spawn maidsafe_vault: {}", why.description()),
+                    Ok(process) => process,
+                });
+        thread::sleep_ms(1000);
+
+        for i in 1..num_of_nodes {
```
chandraprakash: this means we are running n+1 nodes.

benjaminbollen:
okay cheers, sorry missed that !

src/main.rs
```diff
@@ -67,10 +77,74 @@ impl Vault {
...
+    flag_first: bool,
+    flag_help: bool,
+}
+
+#[derive(Debug)]
+enum PeerEndpoint {
+    Tcp(SocketAddr),
+}
+
+impl Decodable for PeerEndpoint {
```
chandraprakash: This is not required I think. We already have this in crust here : https://github.com/maidsafe/crust/blob/master/src/transport.rs#L54

src/main.rs
maqi: I prefer to have an explicit node option instead of using a default none

src/main.rs
maqi: 1..num_of_nodes only runs (num_of_nodes - 1) nodes, so it is still num_of_nodes in total

PR #110, opened 2015-06-20
Handle refresh

maqi:
[Review on Reviewable](https://reviewable.io/reviews/maidsafe/maidsafe_vault/110)


maidsafe-highfive:
r? @chandraprakash

src/vault.rs
```diff
@@ -57,6 +58,48 @@ impl Clone for VaultFacade {
     }
 }
 
+fn merge_payload(type_tag: u64, from_group: NameType, payloads: Vec<Vec<u8>>) -> Option<Payload> {
+    match type_tag {
+      200 => {
```
chandraprakash: Can we get rid of the numbers ?

maqi: this will requires a cross check with u64 serialization and other libs.
maybe better raise a separate issue and leave the work later on.

chandraprakash: ok .. please add an issue for this.

PR #139, opened 2015-07-08
MAID-1185 removing unwrap in library code

maqi:
[Review on Reviewable](https://reviewable.io/reviews/maidsafe/maidsafe_vault/139)

### DIFF
--- a/src/data_manager/database.rs
+++ b/src/data_manager/database.rs
@@ -28,6 +28,8 @@ use routing::types::GROUP_SIZE;
 
 use transfer_parser::transfer_tags::DATA_MANAGER_ACCOUNT_TAG;
 
+use utils::{encode, decode};
+
 type Identity = NameType; // name of the chunk
 type PmidNode = NameType;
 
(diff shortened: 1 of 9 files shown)


maidsafe-highfive:
r? @chandraprakash

src/data_manager/database.rs
```diff
@@ -78,9 +80,10 @@ impl Sendable for DataManagerSendable {
         if self.has_preserialised_content {
             self.preserialised_content.clone()
         } else {
-            let mut e = cbor::Encoder::from_memory();
-            e.encode(&[&self]).unwrap();
-            e.into_bytes()
+            match encode(&self) {
+                Ok(result) => result,
+                Err(_) => Vec::new()
```
chandraprakash: Returning Result() from the method will be more appropriate here.

maqi: This will involve a re-definition in the sendable traits.
As there is chance the sendable will be updated anyway,
I will recommend leave it as what it is so far.

chandraprakash: I am merging it now but add an issue to address this with `Result()`

PR #150, opened 2015-07-13
Updated Linux install scripts

Fraser999:
- the two almost identical scripts have been changed to a single one.
- the script is now safe to run from any path
- it gives useful help if a dependency is missing
- it generates the packages to a fixed path relative to the project root (`<project>/packages/linux`) which has also been added to .gitignore

[Review on Reviewable](https://reviewable.io/reviews/maidsafe/maidsafe_vault/150)


maidsafe-highfive:
r? @chandraprakash

PR #169, opened 2015-08-10
routing in channel

maqi:


maidsafe-highfive:
r? @chandraprakash

PR #188, opened 2015-08-18
use NameType and GROUP_SIZE from real new routing  WIP

maqi:


maidsafe-highfive:
r? @fraser999

PR #193, opened 2015-08-19
New routing event

maqi:


maidsafe-highfive:
r? @fraser999

src/main.rs
```diff
@@ -25,11 +25,13 @@
 #![deny(deprecated, drop_with_repr_extern, improper_ctypes, missing_docs,
         non_shorthand_field_patterns, overflowing_literals, plugin_as_library,
         private_no_mangle_fns, private_no_mangle_statics, raw_pointer_derive, stable_features,
-        unconditional_recursion, unknown_lints, unsafe_code, unused, unused_allocation,
+        unconditional_recursion, unknown_lints, unsafe_code, unused_allocation,
```
Fraser999: We shouldn't be doing this should we?

maqi: have to do it temporarily as there are much un-implemented code blocks.
mark this done and we shall put it back once integration work is done.

Fraser999: I've already reverted this locally and have the code compiling again (albeit with quite a few `#[allow(dead_code)]` instances).

I think modifying the project lint settings should be last resort, then slightly less worse are file-level overrides, then least worst are function/variable-level overrides.

PR #204, opened 2015-08-24
Enable logging

maqi:


maidsafe-highfive:
r? @fraser999

src/main.rs
```diff
@@ -29,15 +29,19 @@
         unused_attributes, unused_comparisons, unused_features, unused_parens, while_true)]
 #![warn(trivial_casts, trivial_numeric_casts, unused_extern_crates, unused_import_braces,
         unused_qualifications, unused_results, variant_size_differences)]
+#![feature(rustc_private)]
```
Fraser999: Can we avoid this?  It ties us to the Nightly channel.

src/main.rs
```diff
@@ -29,15 +29,19 @@
...
+
+#[macro_use]
+extern crate log;
```
Fraser999: I think we should add `log = "*"` to the toml if we're using the crate.

maqi: seems there is some issue with using "log = "*" "
currently crust lib using env_logger, meanwhile routing lib using log
but when I executing vault test, I only see the output from crust, nothing from routing.
so that's why following crust using env_logger here

Fraser999: I think we need both.  `log` is just a facade, and `env_logger` implements it.  There's an issue with using `env_logger` in tests.  You need to [`init`](http://rust-lang.github.io/log/env_logger/fn.init.html) the logger as early as possible.  However, I already looked at how we should handle this for tests, and it's not great: http://stackoverflow.com/q/30177845/2556117

Either way, I think if we're calling `extern crate ...` in main.rs or lib.rs we should have a corresponding entry in the toml.

PR #206, opened 2015-08-25
minor updates

maqi:


maidsafe-highfive:
r? @fraser999

src/vault.rs
```diff
@@ -618,30 +631,48 @@ pub type ResponseNotifier =
...
+
+        let name = ::routing::NameType(::routing::types::vector_as_u8_64_array(
+            ::routing::types::generate_random_vec_u8(64)));
+        let value = ::routing::types::generate_random_vec_u8(1024);
+        let sign_keys =  ::sodiumoxide::crypto::sign::gen_keypair();
+        let sd = ::routing::structured_data::StructuredData::new(0, name, 0,
+            value.clone(), vec![sign_keys.0], vec![], Some(&sign_keys.1)).ok().unwrap();
+        client_routing.put_request(::routing::authority::Authority::ClientManager(client_name),
+                                   ::routing::data::Data::StructuredData(sd.clone()));
+        ::std::thread::sleep_ms(2000);
```
Fraser999: Is there a way to avoid this sleep and the subsequent one?

src/vault.rs
```diff
@@ -435,6 +441,8 @@ impl Vault {
                 },
                 MethodCall::Reply { data } => {
                     if reply_to != None && original_data_request != None {
+                        debug!("as {:?} sending data {:?} to {:?} in responding to the ori_data_request {:?}",
+                               our_authority, data, reply_to.clone().unwrap(), original_data_request.clone().unwrap());
```
Fraser999: `unwrap` used and also in following lines.

Fraser999:
I'd have probably made the `debug!` calls `trace!` since I can't imagine anything more verbose/low-level than these (http://doc.rust-lang.org/log/log/enum.LogLevel.html)

Also, we could maybe remove some of the `#[allow(dead_code)]` statements as we implement more of the logic.

PR #207, opened 2015-08-25
remove unwraps in production code of vault.rs

benjaminbollen:


maidsafe-highfive:
r? @fraser999

src/vault.rs
```diff
@@ -429,8 +440,8 @@ impl Vault {
     fn send(&mut self, our_authority: Authority,
             actions: Vec<MethodCall>,
             response_token: Option<::routing::SignedToken>,
-            reply_to: Option<Authority>,
-            original_data_request: Option<DataRequest>) {
+            opt_reply_to: Option<Authority>,
```
Fraser999: I'm not a fan of abbreviations (should be "optional" rather than "opt") nor of injecting the variable type into the variable name.

PR #211, opened 2015-08-26
use actual routing by default

ustulation:


maidsafe-highfive:
r? @fraser999

PR #214, opened 2015-08-27
Executable test

maqi:


maidsafe-highfive:
r? @fraser999

Fraser999:
As discussed at this morning's meeting, we need to avoid using arbitrary-length sleeps in the tests and we can't use the console output as a means of determining actions done by the vaults.

These are the blockers for this PR, but I also noticed that there are a few `println!`s in the code too which should be removed or replaced by log macros if they're helpful printouts.

Fraser999:
After discussion on Slack, this PR is to be merged as is.

PR #221, opened 2015-09-01
Method call refactor

maqi:


maidsafe-highfive:
r? @Fraser999

src/maid_manager/mod.rs
```diff
@@ -36,12 +36,11 @@ impl MaidManager {
...
 
-    pub fn handle_put(&mut self, from: &NameType,
-                      data: Data) -> Result<Vec<MethodCall>, ResponseError> {
+    pub fn handle_put(&mut self, from: &NameType, data: Data) -> Vec<MethodCall> {
         if self.db_.put_data(from, data.payload_size() as u64) {
-            Ok(vec![MethodCall::Put { location: Authority::NaeManager(data.name()), content: data }])
+            vec![MethodCall::Put { location: Authority::NaeManager(data.name()), content: data }]
         } else {
-            Err(ResponseError::InvalidRequest(data))
+            vec![]
```
Fraser999: Isn't this an example of where we need to differentiate between error types - i.e. aren't there cases where we should be sending an error message back, and other cases where we silently drop the message?

PR #222, opened 2015-09-01
Temporary change - do not install crust.config file

rossmuir:


maidsafe-highfive:
r? @Fraser999

PR #223, opened 2015-09-02
disable local storage reset during churn

maqi:


maidsafe-highfive:
r? @Fraser999

PR #230, opened 2015-09-08
MAID-1324_churn_verification   WIP

maqi:


maidsafe-highfive:
r? @Fraser999

Fraser999:
The change to Routing's API with regards to the refresh was handled in https://github.com/maidsafe/safe_vault/pull/237.

PR #233, opened 2015-09-09
MAID-1335: Removed many `#![allow(...)]` statements.

Fraser999:


maidsafe-highfive:
r? @maqi

Fraser999:
Yeah - I noticed a few tests that were somewhat basic.  I absolutely agree that `handle_account_transfer` tests in the database.rs files should be extended, but I don't think that's really part of this task.

The `handle_account_transfer` tests in the mod.rs files which I removed were fairly pointless in my opinion since the functions are single line calls to the database function.  I wouldn't have removed them as part of this task though except for the fact that they were using a function which was only used by test code (`exist()`) which I removed.

I can't see a way to replace that `exist()` functionality without exposing some private member/function from database, and I prefer to not modify the production code just to accommodate tests.

PR #235, opened 2015-09-09
increase testing env nodes from 4 to 8

maqi:


maidsafe-highfive:
r? @Fraser999

PR #246, opened 2015-09-16
MAID-1331

Fraser999:
Refactored to remove `MethodCall` struct and have personas use Routing instance directly.


maidsafe-highfive:
r? @maqi

PR #275, opened 2015-10-09
Maid 1386 - Chunk store shall not panic

hintofbasil:
Hi,
I hope I'm not jumping the gun by attempting this before the sprint has officially begun.
I know my code introduces two new unwraps which I probably should handle but I was unsure on how to properly handle an error being returned.
I also checked for more panicable code but could only see the two functions.  If there are any I missed I can update them.
Any other comments would be highly welcome!

William

[Review on Reviewable](https://reviewable.io/reviews/maidsafe/safe_vault/275)

Issue #269, opened 2015-09-28
Chunkstore should clean up after itself


Fraser999:
Chunkstore should remove any added dirs/files from the local disk on destruction.

It should also remove previous chunkstores' remnants in case the previous run crashed without cleaning up.  This is not straightforward if we are to support multiple instances of vault running concurrently.  We'd maybe have to name the dir with the current process' PID and check for that PID running before erasing the dir for example.

maqi:
In network test, vaults are being executed as spawned threads, and also vault has PmidNode and StructruedDataManager using separate chunk_store instance.
So the dir structure may needs to be :   <process_id>/<thread_id>/<persona_name>

Fraser999:
Agreed, although I think the cleanup of previous runs should only be looking at PID; I don't think we need per-thread cleanup.

Also, I think I'd go for Vault ID rather than thread name (I think we only have access to [`Thread::name()`](https://doc.rust-lang.org/nightly/std/thread/struct.Thread.html#method.name)).  I'd like to see all threads named to help with debugging (I'm planning to propose it in the style RFC), but either way, all threads will have the same name for every vault, or they'll all have `None`.

PR #277, opened 2015-10-12
MAID-1432

hintofbasil:
It seems the extra terminate functions weren't needed as closing one routing thread then waiting appears to close the other.

[Review on Reviewable](https://reviewable.io/reviews/maidsafe/safe_vault/277)

### DIFF
--- a/src/vault.rs
+++ b/src/vault.rs
@@ -352,30 +352,32 @@ mod test {
     use super::*;
 
     #[cfg(feature = "use-mock-routing")]
-    fn mock_env_setup() -> (Routing, ::std::sync::mpsc::Receiver<(::routing::data::Data)>) {
+    fn mock_env_setup() -> (Routing,
+                            ::std::sync::mpsc::Receiver<(::routing::data::Data)>,
+                            ::std::thread::JoinHandle<()>) {
         ::utils::initialise_logger();
         let run_vault = |mut vault: Vault| {
-            let _ = ::std::thread::spawn(move || {
+            ::std::thread::spawn(move || {
                 vault.do_run();
-            });
+            })
         };
         let vault = Vault::new(None, None);
         let mut routing = vault.pmid_node.routing();
         let receiver = routing.get_client_receiver();
-        let _ = run_vault(vault);
+        let join_handle = run_vault(vault);
 
         let mut available_nodes = Vec::with_capacity(30);
         for _ in 0..30 {
             available_nodes.push(::utils::random_name());
         }
         routing.churn_event(available_nodes, ::utils::random_name());
-        (routing, receiver)
+        (routing, receiver, join_handle)
     }
 
     #[cfg(feature = "use-mock-routing")]
     #[test]
     fn put_get_flow() {
-        let (mut routing, receiver) = mock_env_setup();
+        let (mut routing, receiver, join_handle) = mock_env_setup();
 
         let client_name = ::utils::random_name();
         let sign_keys = ::sodiumoxide::crypto::sign::gen_keypair();
@@ -395,13 +397,13 @@ mod test {
             break;
         }
         routing.stop();
-        ::std::thread::sleep_ms(2000);
+        let _ = join_handle.join();
     }
 
     #[cfg(feature = "use-mock-routing")]
     #[test]
     fn post_flow() {
-        let (mut routing, receiver) = mock_env_setup();
+        let (mut routing, receiver, join_handle) = mock_env_setup();
 
         let name = ::utils::random_name();
         let value = ::routing::types::generate_random_vec_u8(1024);
@@ -442,7 +444,7 @@ mod test {
             break;
         }
         routing.stop();
-        ::std::thread::sleep_ms(2000);
+        let _ = join_handle.join();
     }
 
     #[cfg(not(feature = "use-mock-routing"))]
--- a/src/vault.rs
+++ b/src/vault.rs
@@ -422,7 +422,10 @@ mod test {
         loop {
             match completion_receiver.try_recv() {
                 Err(_) => {}
-                Ok(_) => break,
+                Ok(_) => {
+                    ::std::thread::sleep_ms(20);
+                    break;
+                }
             }
             ::std::thread::sleep_ms(1);
             if starting_time + time_limit < ::time::SteadyTime::now() {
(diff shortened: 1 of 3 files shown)


maidsafe-highfive:

hintofbasil:
Hello,
I'm getting a similar failed test as my other pull request (https://github.com/maidsafe/safe_vault/pull/275).  David seemed to think it was due to an error in routing which should be fixed this sprint.  Any advice on whether this is a similar issue?

Fraser999:
Hi there,

Yes, I've added [an issue for the Windows failure](https://github.com/maidsafe/safe_vault/issues/279) and flagged it as a critical bug, and Qi has been discussing the OS X/Linux failures in [a Routing issue](https://github.com/maidsafe/routing/issues/728) which is also flagged as a critical bug, so I expect your PR doesn't affect these results.

src/vault.rs
```diff
@@ -394,6 +394,8 @@ mod test {
             assert_eq!(it, ::routing::data::Data::ImmutableData(im_data));
             break;
         }
+        routing.terminate();
+        ::std::thread::sleep_ms(2000);
```
maqi: a mandatory sleep duration shall be avoided as much as possible.
If you do think a period needs to be waited for the termination process to be completed,
please consider put in an event flag to notify such completion explicitly.

Before this mandatory sleep has been replaced, I am afraid this PR can not be merged.

hintofbasil:
Hi Qi,
I think I've hit a bit of a road block and was wondering if you could help.
For the tests to guarantee clean-up the spawned thread must close properly.  The only reliable way I can think of to do this is to call join_handler.join() as in my latest commit.  This works perfectly fine for successful tests but obviously blocks if the thread can't close.  I doubt this is acceptable for the tests but I'm not sure how else to approach this issue.

Sorry to bother your with this.
William

src/vault.rs
```diff
@@ -352,30 +352,32 @@ mod test {
...
     #[cfg(feature = "use-mock-routing")]
-    fn mock_env_setup() -> (Routing, ::std::sync::mpsc::Receiver<(::routing::data::Data)>) {
+    fn mock_env_setup() -> (Routing,
+                            ::std::sync::mpsc::Receiver<(::routing::data::Data)>,
+                            ::std::thread::JoinHandle<()>) {
         ::utils::initialise_logger();
         let run_vault = |mut vault: Vault| {
-            let _ = ::std::thread::spawn(move || {
+            ::std::thread::spawn(move || {
                 vault.do_run();
```
maqi: the spawned thread here could be expanded say taking an extra event channel and notify the completion of vault.do_run() through that channel.
by doing so, a blind unlimited wait on thread.join() can be avoided by using try_recv and setup a timeout.

src/vault.rs
```diff
@@ -352,30 +352,32 @@ mod test {
     use super::*;
 
     #[cfg(feature = "use-mock-routing")]
-    fn mock_env_setup() -> (Routing, ::std::sync::mpsc::Receiver<(::routing::data::Data)>) {
+    fn mock_env_setup() -> (Routing,
```
maqi: does the env_setup for the network tests using real routing also needs the similar enhancement to ensure destructor of ChunkStore will be called, so tempdir's dropper will be executed to cleanup the used resource ?

hintofbasil:
Hi Qi,
I've updated all the tests outside of tests/ to clean up their threads properly and hence clean up TempDirs.
However the integration tests use processes instead of threads.  This means I can't send them a terminate signal to have them exit cleanly.  Is there likely a good method to deal with this?  Would simply deleting the files after the tests be the best solution here?

maqi:
Hi, There,

Regarding the integration test, this involves sending a termination signal to a process, which is not in the scope of the task 1432, neither concerned by the issue https://github.com/maidsafe/safe_vault/issues/269 .

You don't need to worry about it.

src/vault.rs
```diff
@@ -439,6 +477,22 @@ mod test {
...
+                Err(_) => {},
+                Ok(_) => {
+                    ::std::thread::sleep_ms(20);
+                    break;
+                }
+            }
+            ::std::thread::sleep_ms(1);
+            if starting_time + time_limit < ::time::SteadyTime::now() {
+                panic!("vault did not shut down in expected duration");
+            }
```
maqi: This part better to be a utility function as it is being repeatedly used in different tests

src/vault.rs
```diff
@@ -649,11 +707,40 @@ mod test {
...
+                        completion_receiver: ::std::sync::mpsc::Receiver<::routing::event::Event>) {
+        client_routing.stop();
+        for notifier in &vault_notifiers {
+            let _ = notifier.1.send(0);
+        }
+        let starting_time = ::time::SteadyTime::now();
+        let time_limit = ::time::Duration::seconds(10);
+        for _ in 0..vault_notifiers.len() {
+            loop {
+                match completion_receiver.try_recv() {
```
maqi: there is only one receiver for all the vaults, means here is only waiting for any one vault's completion,
not for ALL vaults.

src/vault.rs
```diff
@@ -108,8 +108,27 @@ impl Vault {
...
 
+    #[cfg(test)]
+    fn confirm_run_complete(&self, completion_sender: Option<::std::sync::mpsc::Sender<::routing::event::Event>>)
+    {
+        match completion_sender {
+            Some(sender) => {
+                let _ = sender.send(::routing::event::Event::Terminated);
+            },
+            None => { },
+        }
```
maqi: I think you can use
completion_sender.and_then(|sender| let _ = sender.send(::routing::event::Event::Terminated));
to simplify the code

src/vault.rs
hintofbasil: Hi Qi,
I've tested this and I believe it is verifying all vaults.  The break here breaks the loop but not the for loop.  The for loop ensures that 8 (vault_notifiers.len()) events are received from the completion_receiver.  This can be verified by adding in a println!() to the Ok match.
Let me know if you think I'm still mistaken here.

maqi:
After the work PR https://github.com/maidsafe/safe_vault/pull/294 resolving Issue https://github.com/maidsafe/safe_vault/issues/279,
vault spawned in threads during the network_test can now be properly terminated,
which calls the destructor of chunk_store and clean up the tempdir automatically.

However, PR 294 only focused on termination with real routing, the clean up of mock-routing was not concerned,
which still leave tempdir not cleaned up properly when executing mocking tests.

But this issue shall be resolved in a new PR based from the latest master head, instead of keeps working on this PR.

PR #293, opened 2015-10-19
MAID-1453_MAID-1454 DM handle replication

maqi:
[Review on Reviewable](https://reviewable.io/reviews/maidsafe/safe_vault/293)

### DIFF
--- a/src/data_manager/mod.rs
+++ b/src/data_manager/mod.rs
@@ -142,20 +142,27 @@ impl DataManager {
         // Before querying the records, first ensure all records are valid
         let ongoing_gets = self.ongoing_gets.retrieve_all();
         let mut failing_entries = Vec::new();
+        let mut fetching_list = ::std::collections::BTreeSet::new();
+        fetching_list.insert(data_name.clone());
         for ongoing_get in ongoing_gets {
             if ongoing_get.1 + ::time::Duration::seconds(10) < ::time::SteadyTime::now() {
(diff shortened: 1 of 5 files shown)


maidsafe-highfive:
r? @Fraser999

src/data_manager/mod.rs
```diff
@@ -407,10 +441,12 @@ impl DataManager {
 
     fn replicate_to(&mut self, name: &::routing::NameType) -> Option<::routing::NameType> {
         let pmid_nodes = self.database.get_pmid_nodes(name);
-        if pmid_nodes.len() < 3 && pmid_nodes.len() > 0 {
+        if pmid_nodes.len() < (PARALLELISM / 2 + 1) && pmid_nodes.len() > 0 {
```
Fraser999: I guess `PARALLELISM` isn't really an appropriate name for this variable - maybe something more like `REPLICATION_COUNT`?  Also, I think we should either have a second variable for the threshold (e.g. `pub const MIN_REPLICANTS: usize = 3;`) or for the factor (e.g. `pub const MIN_REPLICATION_FACTOR: f32 = 0.75;`)

src/utils.rs
```diff
@@ -136,6 +136,14 @@ pub fn random_name() -> ::routing::NameType {
         ::routing::types::generate_random_vec_u8(::routing::NAME_TYPE_LEN)))
 }
 
+// #[cfg(all(test, not(feature = "use-mock-routing")))]
```
Fraser999: Can we remove this commented test?

PR #299, opened 2015-10-21
Avoid consuming excessive system CPU usage.  Fixes #298.

### LINKED ISSUE #298: safe_vault uses 100% CPU no matter what
I've compiled safe_vault at 0507ad3f3f5a264e3618d914c7220fa618fb210f commit, didn't test previous ones, but when I run it on ubuntu 14.04 x86_64 machine (didn't test anywhere else) safe_vault seems to be always stuck at 100% CPU. Tried 10 vaults at once, now due to contention it's lower than 100% per each core but it's around 60-100 at all times. I don't think this is normal.

Fraser999:
[Review on Reviewable](https://reviewable.io/reviews/maidsafe/safe_vault/299)

### DIFF
--- a/src/vault.rs
+++ b/src/vault.rs
@@ -98,7 +98,7 @@ impl Vault {
                     self.pmid_node.routing().stop();
                 }
             }
-            ::std::thread::yield_now();
+            ::std::thread::sleep_ms(1);
         }
     }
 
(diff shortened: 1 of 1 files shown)


maidsafe-highfive:
r? @maqi

PR #301, opened 2015-10-22
MAID-1433_MAID-1434_Issue_269_cleanup_chunkstore_tempdir

maqi:
[Review on Reviewable](https://reviewable.io/reviews/maidsafe/safe_vault/301)

### DIFF
--- a/src/chunk_store/chunk_store.rs
+++ b/src/chunk_store/chunk_store.rs
@@ -26,7 +26,16 @@ pub struct ChunkStore {
 impl ChunkStore {
     /// Create new chunkstore with `max_disk_usage` allowed disk usage.
     pub fn new(max_disk_usage: usize) -> Result<ChunkStore, ::error::ChunkStoreError> {
-        let tempdir = try!(::tempdir::TempDir::new("safe_vault"));
+        let own_pid = format!("{}", Self::get_own_pid());
+        let folder_name = format!("safe_vault-{}/", own_pid);
+        let mut path = ::std::env::temp_dir();
+        path.push(folder_name);
+        println!("{:?}", path);
+        if Self::cleanup() {
+            let _ = ::std::fs::create_dir(&path);
+        }
+        let tempdir = try!(::tempdir::TempDir::new_in(path, "safe_vault"));
+
         Ok(ChunkStore {
             tempdir: tempdir,
             max_disk_usage: max_disk_usage,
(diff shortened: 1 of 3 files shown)


maidsafe-highfive:
r? @Fraser999

src/main.rs
```diff
@@ -36,6 +36,7 @@
         unused_qualifications, unused_results, variant_size_differences)]
 #![allow(box_pointers, fat_ptr_transmutes, missing_copy_implementations,
          missing_debug_implementations)]
+#![feature(libc)]
```
Fraser999: We can't use this - it ties us to the Nightly channel :(

src/chunk_store/chunk_store.rs
```diff
@@ -138,4 +145,65 @@ impl ChunkStore {
...
+    }
+
+    #[allow(unsafe_code)]
+    fn get_own_pid() -> u32 {
+        extern crate libc;
+        extern { fn getpid() -> libc::c_uint; }
+        unsafe { getpid() }
+    }
+
+    #[cfg(target_os="windows")]
```
Fraser999: I think we can just use `#[cfg(windows)]` here (and similar comment for the "not windows" one below).

src/chunk_store/test.rs
```diff
@@ -27,6 +27,29 @@ mod test {
         string
     }
 
+    fn has_child_dir(parent: ::std::path::PathBuf, child: &::std::path::PathBuf) -> bool {
+        ::std::fs::read_dir(&parent).ok().and_then(|mut dir_entries| {
+            dir_entries.find(|dir_entry| {
+                match dir_entry {
+                    &Ok(ref entry) =>
+                        entry.file_name().to_str() == child.file_name().unwrap().to_str(),
```
Fraser999: Can you use `evaluate_result!` or `evaluate_option!` (https://github.com/maidsafe/safe_vault/blob/master/src/macros.rs) instead of `unwrap()` please?

src/chunk_store/chunk_store.rs
```diff
@@ -138,4 +145,65 @@ impl ChunkStore {
...
+        ::std::fs::read_dir(::std::env::temp_dir()).ok().and_then(|dir_entries| {
+            dir_entries.filter(|dir_entry| {
+                match dir_entry {
+                    &Ok(ref entry) => {
+                        let line = entry.file_name().into_string().ok().unwrap_or(String::from(""));
+                        match (line.contains("safe_vault"), line.contains(&own_pid)) {
+                            (true, false) => {
+                                let v: Vec<&str> = line.split("-").collect();
+                                if v.len() > 1 && !safe_vault_pids.contains(&String::from(v[1])) {
+                                    let _ = ::std::fs::remove_dir_all(entry.path());
```
Fraser999: Can we log the error here please rather than throwing away the result?

src/chunk_store/chunk_store.rs
```diff
@@ -26,7 +26,14 @@ pub struct ChunkStore {
 impl ChunkStore {
     /// Create new chunkstore with `max_disk_usage` allowed disk usage.
     pub fn new(max_disk_usage: usize) -> Result<ChunkStore, ::error::ChunkStoreError> {
-        let tempdir = try!(::tempdir::TempDir::new("safe_vault"));
+        let folder_name = format!("safe_vault-{}/", Self::get_own_pid());
+        let mut path = ::std::env::temp_dir();
+        path.push(folder_name);
+        if Self::cleanup() {
+            let _ = ::std::fs::create_dir(&path);
```
Fraser999: Shouldn't this be a `try!` so we don't continue if it fails?

src/chunk_store/chunk_store.rs
```diff
@@ -138,4 +145,65 @@ impl ChunkStore {
...
+    fn get_own_pid() -> u32 {
+        extern crate libc;
+        extern { fn getpid() -> libc::c_uint; }
+        unsafe { getpid() }
+    }
+
+    #[cfg(target_os="windows")]
+    fn get_all_vault_pids() -> Vec<String> {
+        let output = ::std::process::Command::new("tasklist").output()
+                .unwrap_or_else(|e| { panic!("failed to execute process: {}", e) });
```
Fraser999: Should we be `panic`ing here?  If we can't get all other vaults' PIDs, we should maybe just not do any cleanup?

Same for Linux/OS X function below.

Issue #278, opened 2015-10-12
Network tests should clean up bootstrap and config files


Fraser999:
Any test which causes a Crust bootstrap file or Crust config file to be created should try and remove these at the start (and ideally at the end too) of each run.

I think the bootstrap file is the more problematic, since old files can cause tests to take too long to form a testnet or start a vault due to the old bootstrap file being populated with defunct endpoints.

maqi:
https://github.com/maidsafe/crust/pull/357
https://github.com/maidsafe/safe_vault/pull/281

maqi:
This is the PR creating empty bootstrap for test (and remove after the test)

Fraser999:
As mentioned above, this is partially addressed in https://github.com/maidsafe/safe_vault/pull/283.  However, that PR only deals with the bootstrap file for a single test; it doesn't cover the config file, nor all test cases, and it leaves behind the bootstrap file if the test fails.

maqi:
In case of test failure, will it be good to leave behind the bootstrap file ? as some investigation may rely on the bootstrap status to figure out what the problem is?

Fraser999:
I don't think so, but if this is useful, we should try and output it to the console so it'll be alongside the other output which will be more likely useful for debugging.  Then if the user wants to keep the output he can redirect to a file and it's up to him whether to clean that up or not.  At least the test will be cleaning up after itself.

PR #304, opened 2015-10-27
Issue_278 remove config file after the network_test

maqi:
Fixes https://github.com/maidsafe/safe_vault/issues/278

[Review on Reviewable](https://reviewable.io/reviews/maidsafe/safe_vault/304)

### DIFF
--- a/src/vault.rs
+++ b/src/vault.rs
@@ -471,10 +471,8 @@ mod test {
             ::utils::initialise_logger();
             Self::show_warning();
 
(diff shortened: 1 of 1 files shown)


maidsafe-highfive:
r? @Fraser999

Fraser999:
The three added functions are all pretty much duplicates of existing functions.  I'd much rather see the common code moved out to separate functions to minimise code duplication.

PR #309, opened 2015-11-11
updating libs, allowing some lints

waynenilsen:
the changes to `Cargo.toml` should be rolled back but for now they were needed to compile, also should do a minor version bump

[Review on Reviewable](https://reviewable.io/reviews/maidsafe/safe_vault/309)


maidsafe-highfive:

maqi:
Hi, there,

Thanks for the contribution.
As you can see, the PR failed CI builds when trying building tests using mock routing (cargo test --features "use-mock-routing")

Also, there is a policy that "allow" shall be avoided. Is it possible to remove the un-used parts?

One last thing is regarding the routing version to be used. Pointing to a specific commit shall be avoided, so seems we have to wait for the version pump of routing lib to be completed.

maqi:
Hi, there,

I cross checked with other stuff and seems the lint changes are due to the same issue as reported in https://github.com/maidsafe/crust/issues/410, which should be fixed in the next release of Rust nightly hopefully.

Regarding the dependency image, it is preferred not to include the svg file directly, instead of having some post-commit auto-generating hook to anyone else can view it online if interested.

So, I will close this PR at this moment.

waynenilsen:
I think that removing  the deprecated `sleep_ms` and upgrading the crust version should probably be included, do you? I can roll back the other changes.

dirvine:
:+1:

maqi:
Hi, Waynenilsen,

Could you please raise a new PR only includes the part removing deprecated `sleep_ms` and updating crust version?
thx.

PR #310, opened 2015-11-13
updated crust to 0.5.0 and fixed some changes in std

waynenilsen:
see also #309 @maqi

PR #311, opened 2015-11-16
Build fixes and remove version wildcards

rossmuir:
[Review on Reviewable](https://reviewable.io/reviews/maidsafe/safe_vault/311)

### DIFF
--- a/src/mock_routing/mock_routing_impl.rs
+++ b/src/mock_routing/mock_routing_impl.rs
@@ -169,7 +169,7 @@ impl MockRoutingImpl {
         let cloned_sender = self.sender.clone();
         let cloned_client_sender = self.client_sender.clone();
         let _ = ::std::thread::spawn(move || {
-            l::std::thread::sleep(::std::time::Duration::from_millis(delay_ms as u64));
+            ::std::thread::sleep(::std::time::Duration::from_millis(delay_ms as u64));
             match location.clone() {
                 ::routing::Authority::NaeManager(_) => {
                     let _ = cloned_sender.send(::routing::event::Event::Response {
--- a/src/mock_routing/mock_routing_impl.rs
+++ b/src/mock_routing/mock_routing_impl.rs
@@ -95,8 +95,7 @@ impl MockRoutingImpl {
         let delay_ms = self.network_delay_ms;
         let cloned_sender = self.sender.clone();
         let _ = ::std::thread::spawn(move || {
-            let duration = ::std::time::Duration::from_millis(delay_ms as u64);
-            ::std::thread::sleep(duration);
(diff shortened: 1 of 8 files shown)


maidsafe-highfive:
r? @maqi

rossmuir:
Thanks to @waynenilsen for https://github.com/maidsafe/safe_vault/pull/310 included in this PR

rossmuir:
Issue raised for failing test > https://github.com/maidsafe/safe_vault/issues/312

maqi:
The failing test seems is due to the updates in crust/routing, when setting up the test network and there are more than 2 nodes ( say the third or forth nodes ) the previous established connection starts dropping among nodes.
So the test always failed as the setup of testing network is failed.

I think this PR can be merged and we shall have a detailed look at what's happening in crust/routing

rossmuir:
@Viv-Rajkumar I have removed version change for just now, until all tests pass - see @maqi comment above. This PR now just fixes the build.

src/mock_routing/mock_routing_impl.rs
```diff
@@ -78,7 +78,8 @@ impl MockRoutingImpl {
         let delay_ms = self.network_delay_ms;
         let cloned_sender = self.sender.clone();
         let _ = ::std::thread::spawn(move || {
-            ::std::thread::sleep_ms(delay_ms);
+            let duration = ::std::time::Duration::from_millis(delay_ms as u64);
+            ::std::thread::sleep(duration);
             let _ = cloned_sender.send(::routing::event::Event::Request {
```
maqi: can we use
::std::thread::sleep(::std::time::Duration::from_millis(delay_ms as u64))
here ?

src/mock_routing/mock_routing_impl.rs
```diff
@@ -169,7 +169,7 @@ impl MockRoutingImpl {
         let cloned_sender = self.sender.clone();
         let cloned_client_sender = self.client_sender.clone();
         let _ = ::std::thread::spawn(move || {
-            ::std::thread::sleep_ms(delay_ms);
+            l::std::thread::sleep(::std::time::Duration::from_millis(delay_ms as u64));
```
maqi: isn't it a typo ?

src/mock_routing/mock_routing_impl.rs
```diff
@@ -78,7 +78,7 @@ impl MockRoutingImpl {
         let delay_ms = self.network_delay_ms;
```
dirvine: yes typo below plus if this was `let delay_ms: u64 = self.network_delay_ms;`it would maybe prevent typing all those as u64, but I feel this should be usize not u64 for 32 bit systems, perhaps

rossmuir:
Sorry @maqi yes a typo - should be fixed now.

PR #321, opened 2016-02-01
OutboxHas

brian-js:
[Review on Reviewable](https://reviewable.io/reviews/maidsafe/safe_vault/321)


maidsafe-highfive:
r? @Fraser999

src/personas/mpid_manager.rs
```diff
@@ -315,50 +318,77 @@ impl MpidManager {
...
+                if let Some(ref account) = self.accounts.get(&request.dst.get_name().clone()) {
+                    let names_in_outbox = header_names.iter()
+                                                      .filter(|h| account.has_in_outbox(h))
+                                                      .cloned()
+                                                      .collect::<Vec<XorName>>();
+                    let mut mpid_headers = vec![];
+
+                    for name in names_in_outbox.iter() {
+                        if let Ok(data) = self.chunk_store_outbox.get(name) {
+                            let mpid_message: MpidMessage = unwrap_result!(deserialise(&data));
```
maqi: the for loop will be broken as long as there is any one corrupted data.
probably we have to use a match to ensure the for loop can tolerate corrupted data.

brian-js: If it's anything other than Ok it'll loop to the next name, IIUC.

src/personas/mpid_manager.rs
```diff
@@ -315,50 +318,77 @@ impl MpidManager {
...
                                     client.clone(), Data::PlainData(data.clone()), message_id.clone());
                             }
                         }
                     }
-                    None => warn!("can not find the account {:?}", dst.get_name()),
+                    None => warn!("can not find the account {:?}", request.dst.get_name().clone()),
+                }
+            }
+            MpidMessageWrapper::OutboxHas(header_names) => {
+                if let Some(ref account) = self.accounts.get(&request.dst.get_name().clone()) {
```
maqi: do we need to check the src requester is the correct owner of the account (src being one of the registered client)?
the current implementation allows anyone query other's outbox

src/personas/mpid_manager.rs
```diff
@@ -83,7 +83,7 @@ impl MailBox {
     }
 
     #[allow(dead_code)]
```
maqi: seems the allow dead_code can be removed

PR #323, opened 2016-02-02
MpidManager sending PutFailure: outbox or inbox full

maqi:
[Review on Reviewable](https://reviewable.io/reviews/maidsafe/safe_vault/323)

### DIFF
--- a/src/personas/mpid_manager.rs
+++ b/src/personas/mpid_manager.rs
@@ -190,17 +190,17 @@ impl MpidManager {
         match mpid_message_wrapper {
             MpidMessageWrapper::PutHeader(_mpid_header) => {
                 if self.chunk_store_inbox.has_chunk(&data.name()) {
-                    return Err(InternalError::Client(ClientError::DataExists));;
+                    return Err(InternalError::Client(ClientError::DataExists));
                 }
                 // TODO: how the sender's public key get retained?
                 if self.accounts
                        .entry(request.dst.get_name().clone())
                        .or_insert(Account::default())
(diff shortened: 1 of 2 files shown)


maidsafe-highfive:
r? @Fraser999

src/personas/mpid_manager.rs
```diff
@@ -198,6 +198,9 @@ impl MpidManager {
                        .or_insert(Account::default())
                        .put_into_inbox(data.payload_size() as u64, &data.name(), &None) {
                     let _ = self.chunk_store_inbox.put(&data.name(), data.value());
+                } else {
+                    let _ = routing_node.send_put_failure(request.dst.clone(),
+                        request.src.clone(), request.clone(), Vec::new(), message_id);
```
brian-js: Prefer using something like `try` to `let _ =`

src/personas/mpid_manager.rs
```diff
@@ -237,6 +240,44 @@ impl MpidManager {
...
+                }
+            }
+            _ => unreachable!("Error in vault demuxing"),
+        }
+        Ok(())
+    }
+
+    // PutFailure only happens from receiver's MpidManager to sender's MpidManager to
+    // indicate an inbox full.
```
brian-js: PutFailure is returned from MpidManager to Client, if Clients' Outbox or Inbox is full.

maqi: This comment is about a PutFailure between mpid_managers.
the one you mentioned is to be handled by client

PR #324, opened 2016-02-02
Handle deletion for message from outbox and header from inbox.

brian-js:
Requires https://github.com/maidsafe/mpid_messaging/pull/8 for mpid_messaging merged, version changed and published.

[Review on Reviewable](https://reviewable.io/reviews/maidsafe/safe_vault/324)

### DIFF
--- a/src/personas/mpid_manager.rs
+++ b/src/personas/mpid_manager.rs
@@ -65,7 +65,6 @@ impl MailBox {
         }
     }
 
-    #[allow(dead_code)]
     fn remove(&mut self, size: u64, entry: &XorName) -> bool {
         if !self.mail_box.contains_key(entry) {
             return false;
@@ -123,12 +122,10 @@ impl Account {
         self.inbox.put(size, entry, public_key)
     }
 
-    #[allow(dead_code)]
     fn remove_from_outbox(&mut self, size: u64, entry: &XorName) -> bool {
         self.outbox.remove(size, entry)
     }
 
-    #[allow(dead_code)]
     fn remove_from_inbox(&mut self, size: u64, entry: &XorName) -> bool {
         self.inbox.remove(size, entry)
     }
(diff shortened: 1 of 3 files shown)


maidsafe-highfive:
r? @maqi

src/personas/mpid_manager.rs
```diff
@@ -419,6 +416,68 @@ impl MpidManager {
...
+
+    pub fn handle_delete(&mut self, routing_node: &RoutingNode, request: &RequestMessage)
+            -> Result<(), InternalError> {
+        let (data, message_id) = match request.content {
+            RequestContent::Delete(Data::PlainData(ref data), ref message_id) => {
+                (data.clone(), message_id.clone())
+            }
+            _ => unreachable!("Error in vault demuxing"),
+        };
+        let mpid_message_wrapper = unwrap_option!(deserialise_wrapper(data.value()),
```
maqi: the rfc defines the delete request not being wrapped in the MpidMessageWrapper.
which means the data is expected to be the MpidHeader or MpidMessage directly.

src/personas/mpid_manager.rs
```diff
@@ -419,6 +416,68 @@ impl MpidManager {
...
+                if let Some(ref mut account) = self.accounts.get_mut(&request.dst.get_name().clone()) {
+                    let mut registered = false;
+
+                    if account.registered_clients().iter().any(|authority| *authority == request.src) {
+                        registered = true;
+                    }
+
+                    if let Ok(data) = self.chunk_store_outbox.get(&message_name) {
+                        if !registered {
+                            let mpid_message: MpidMessage = unwrap_result!(deserialise(&data));
```
maqi: the current stored data in chunk_store, is a serialised MpidMessageWrapper::PutMessage(mpid_message).
refer https://github.com/maidsafe/safe_vault/blob/master/src/personas/mpid_manager.rs#L307-L312  for detail.

brian-js: The code referred to is for the MpidMessageWrapper::PutHeader and should probably be updated to store the serialised Header.

PR #325, opened 2016-02-03
Initial test and bug fixes.

brian-js:
[Review on Reviewable](https://reviewable.io/reviews/maidsafe/safe_vault/325)

### DIFF
--- a/src/personas/mpid_manager.rs
+++ b/src/personas/mpid_manager.rs
@@ -179,14 +179,14 @@ impl MpidManager {
             }
             _ => unreachable!("Error in vault demuxing"),
         };
-        let mpid_message_wrapper: MpidMessageWrapper = unwrap_result!(deserialise(&data.value()));
+        let mpid_message_wrapper: MpidMessageWrapper = try!(deserialise(&data.value()));
         match mpid_message_wrapper {
             MpidMessageWrapper::PutHeader(mpid_header) => {
                 if self.chunk_store_inbox.has_chunk(&data.name()) {
                     return Err(InternalError::Client(ClientError::DataExists));
                 }
                 // TODO: how the sender's public key get retained?
-                let serialised_header = unwrap_result!(serialise(&mpid_header));
+                let serialised_header = try!(serialise(&mpid_header));
                 if self.accounts
                        .entry(request.dst.get_name().clone())
                        .or_insert(Account::default())
@@ -202,7 +202,7 @@ impl MpidManager {
                     return Err(InternalError::Client(ClientError::DataExists));
                 }
                 // TODO: how the sender's public key get retained?
-                let serialised_message = unwrap_result!(serialise(&mpid_message));
+                let serialised_message = try!(serialise(&mpid_message));
                 if self.accounts
                        .entry(request.dst.get_name().clone())
                        .or_insert(Account::default())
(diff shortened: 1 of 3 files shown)


maidsafe-highfive:
r? @Fraser999

src/personas/mpid_manager.rs
```diff
@@ -49,29 +49,21 @@ impl MailBox {
 
 
     fn put(&mut self, size: u64, entry: &XorName, public_key: &Option<PublicKey>) -> bool {
-        if size > self.space_available {
-            return false;
-        }
-        if self.mail_box.contains_key(entry) {
+        if self.used_space + size > self.space_available {
```
maqi: MailBox already has allowance member field

src/personas/mpid_manager.rs
```diff
@@ -49,29 +49,21 @@ impl MailBox {
...
             return false;
         }
+
         match self.mail_box.insert(entry.clone(), public_key.clone()) {
-            Some(_) => {
+            Some(_) => false,
+            None => {
                 self.used_space += size;
                 self.space_available -= size;
```
maqi: this usage of space_available is not consistent with other changes

src/personas/mpid_manager.rs
```diff
@@ -49,29 +49,21 @@ impl MailBox {
...
-            None => false,
         }
     }
 
-    #[allow(dead_code)]
     fn remove(&mut self, size: u64, entry: &XorName) -> bool {
-        if !self.mail_box.contains_key(entry) {
-            return false;
-        }
-        self.used_space -= size;
```
maqi: shall used_space to be updated?

src/personas/mpid_manager.rs
```diff
@@ -49,29 +49,21 @@ impl MailBox {
...
-        self.space_available += size;
         match self.mail_box.remove(entry) {
             Some(_) => {
                 self.used_space -= size;
```
maqi: space_available shall be updated here, unless it is going to be used just as allowance.

src/personas/mpid_manager.rs
brian-js: ?

src/personas/mpid_manager.rs
brian-js: If insert returns a value the entry was already there, otherwise it's been added.

src/personas/mpid_manager.rs
brian-js: It is on the line below, is it not?

src/personas/mpid_manager.rs
maqi: I mean, in the above ( if self.used_space + size > self.space_available { ), space_available is used like a constant (allowance), but here it is being updated

src/personas/mpid_manager.rs
maqi: sorry, my fault,
yes it is.

src/personas/mpid_manager.rs
brian-js: Yeah, it should have been allowance, but since we've decided not to use that for the time being I'll fix the calculation.

Fraser999:
Reviewed 2 of 2 files at r4.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/maidsafe/safe_vault/325)*

PR #332, opened 2016-02-08
WIP - Integrating with (currently unpublished) updated Routing API.

Fraser999:
[Review on Reviewable](https://reviewable.io/reviews/maidsafe/safe_vault/332)

PR #334, opened 2016-02-11
Messaging test

brian-js:
Updates,

1. PutMessage requires client online registration prior to sending.
2. Sends put/post success for PutMessage/Online requests.
3. When PutHeader messages are received check if receiver is online and push the message.

[Review on Reviewable](https://reviewable.io/reviews/maidsafe/safe_vault/334)


maidsafe-highfive:
r? @Fraser999

src/personas/mpid_manager.rs
```diff
@@ -185,8 +184,27 @@ impl MpidManager {
...
+                    // TODO: how the sender's public key get retained?
+                    if account.put_into_inbox(serialised_header.len() as u64, &data.name(), &None) {
+                        try!(self.chunk_store_inbox.put(&data.name(), &serialised_header[..]));
+                        let dst = Authority::ClientManager(mpid_header.sender().clone());
+                        let wrapper = MpidMessageWrapper::GetMessage(mpid_header.clone());
+                        let value = try!(serialise(&wrapper));
+                        let name = try!(mpid_header.name());
+                        let data = Data::PlainData(PlainData::new(name, value));
+                        try!(routing_node.send_post_request(request.dst.clone(), dst, data, message_id.clone()));
+                        return Ok(())
```
maqi: If we do need to push the msg to the client as soon as received the header notification, I think it will be better to forward the full message directly.
the header part can still be recovered from the full message and get stored, meanwhile a GetMessage can be saved to be sent out.

src/personas/mpid_manager.rs
```diff
@@ -185,8 +184,27 @@ impl MpidManager {
...
+                    } else {
+                        try!(routing_node.send_put_failure(request.dst.clone(),
+                                request.src.clone(), request.clone(), Vec::new(), message_id));
+                        return Ok(())
+                    }
+                }
+
                 if self.accounts
```
maqi: this seems now having some kind of duplication with the code above, can these two parts get merged together?

src/personas/mpid_manager.rs
```diff
@@ -198,16 +216,19 @@ impl MpidManager {
...
-                if self.accounts
-                       .entry(request.dst.get_name().clone())
-                       .or_insert(Account::default())
-                       .put_into_outbox(serialised_message.len() as u64, &data.name(), &None) {                    
+                if let Some(ref mut account) = self.accounts.get_mut(&request.dst.get_name().clone()) {
+                    if self.chunk_store_outbox.has_chunk(&data.name()) {
+                        return Err(InternalError::Client(ClientError::DataExists));
+                    }
+                    let serialised_message = try!(serialise(&mpid_message));
                     try!(self.chunk_store_outbox.put(&data.name(), &serialised_message[..]));
```
maqi: put store this chunk after put_into_outobx got succeeded

src/personas/mpid_manager.rs
brian-js: Probably depends on user behaviour. It makes sense if we assume they're online most of the time.

src/personas/mpid_manager.rs
brian-js: We get borrowing of mutable accounts when already borrowed errors if we do that. If the client is online the header is stored and we return, otherwise the header is stored for the created account, either way it only happens once.

src/personas/mpid_manager.rs
brian-js: The order hasn't changed, it probably should though.

Fraser999:
This has been merged into PR #332 pending a Routing publish.

PR #338, opened 2016-02-24
Ci tests implement delete tests

maqi:
[Review on Reviewable](https://reviewable.io/reviews/maidsafe/safe_vault/338)


maidsafe-highfive:
r? @Fraser999

Fraser999:
I guess this needs discussed in an RFC to be honest.

I think what you have looks like it should work for now, but I'm not sure about using empty content for a deleted SD.  This can all be discussed in the RFC though.

Fraser999:
Reviewed 8 of 10 files at r1, 2 of 2 files at r2.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/maidsafe/safe_vault/338)*

PR #343, opened 2016-03-01
For Get requests in the CI tests, allow 2 failures where appropriate.

Fraser999:
[Review on Reviewable](https://reviewable.io/reviews/maidsafe/safe_vault/343)

### DIFF
--- a/tests/detail/client.rs
+++ b/tests/detail/client.rs
@@ -98,11 +98,16 @@ impl Client {
     }
 
     /// Send a `Get` request to the network and return the received response.
-    pub fn get(&mut self, request: DataRequest) -> Option<ResponseMessage> {
-        unwrap_result!(self.routing_client
-                           .send_get_request(Authority::NaeManager(request.name()),
-                                             request.clone()));
-        self.wait_for_response()
+    pub fn get(&mut self, request: DataRequest, max_attempts: u32) -> Option<ResponseMessage> {
+        for i in 0..max_attempts {
+            unwrap_result!(self.routing_client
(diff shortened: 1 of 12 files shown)


maidsafe-highfive:
r? @maqi

tests/detail/client.rs
```diff
@@ -98,11 +98,16 @@ impl Client {
...
-        unwrap_result!(self.routing_client
-                           .send_get_request(Authority::NaeManager(request.name()),
-                                             request.clone()));
-        self.wait_for_response()
+    pub fn get(&mut self, request: DataRequest, max_attempts: u32) -> Option<ResponseMessage> {
+        for i in 0..max_attempts {
+            unwrap_result!(self.routing_client
+                               .send_get_request(Authority::NaeManager(request.name()),
+                                                 request.clone()));
+            if let Some(response) = self.wait_for_response() {
```
maqi: wait_for_response will return with both GetFailure and GetSuccess, which means this max_attempts only means:
in case of time_out (no event from network), there will be max_attempts; if there is failure (get or put failure), immediately return (which may happen if PutSucess is sent from MM and we are trying to fetch from SDM/IDM immediately after).
is this mean to be like this?

Fraser999: Yeah - well spotted - you're right :)  I'll update the PR now.

maqi:
Reviewed 7 of 7 files at r2.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/maidsafe/safe_vault/343)*

PR #345, opened 2016-03-01
Move request_cache to MM Account.

brian-js:
[Review on Reviewable](https://reviewable.io/reviews/maidsafe/safe_vault/345)

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -29,6 +29,7 @@ pub enum ClientError {
     NoSuchData,
     DataExists,
     LowBalance,
+    DuplicateRequest,
 }
 
 #[derive(Debug)]
--- a/src/personas/maid_manager.rs
+++ b/src/personas/maid_manager.rs
@@ -153,7 +153,7 @@ impl MaidManager {
                               routing_node: &RoutingNode,
                               message_id: &MessageId)
                               -> Result<(), InternalError> {
-        for (_, mut account) in self.accounts.iter_mut() {
+        for (_, account) in self.accounts.iter_mut() {
             if let Some(client_request) = account.remove_cached_request(message_id) {
                 // Send success response back to client
                 let message_hash = sha512::hash(&try!(serialisation::serialise(&client_request))[..]);
--- a/src/personas/maid_manager.rs
+++ b/src/personas/maid_manager.rs
@@ -15,6 +15,7 @@
 // Please review the Licences for the specific language governing permissions and limitations
 // relating to use of the SAFE Network Software.
 
+use std::fmt;
 use rustc_serialize::{Decodable, Decoder, Encodable, Encoder};
 use error::{ClientError, InternalError};
 use lru_time_cache::LruCache;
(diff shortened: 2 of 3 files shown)


maidsafe-highfive:
r? @maqi

src/personas/maid_manager.rs
```diff
@@ -225,22 +276,26 @@ impl MaidManager {
...
+        // Update the account
+        if let Err(error) = self.accounts
+                                .get_mut(&client_name)
+                                .ok_or(ClientError::NoSuchAccount)
+                                .and_then(|account| {
+                                    if type_tag != 0 {
+                                        try!(account.cache_request(message_id.clone(), &request));
+                                        account.put_data(DEFAULT_PAYMENT /* data.payload_size() as u64 */)
+                                    } else {
+                                        account.cache_request(message_id.clone(), &request)
```
maqi: put request of SD having type_tag = 0 won't be passed to SDM, does it still required to cache the request?

src/personas/maid_manager.rs
```diff
@@ -119,26 +179,17 @@ impl MaidManager {
                               message_id: &MessageId,
                               external_error_indicator: &Vec<u8>)
                               -> Result<(), InternalError> {
-        match self.request_cache.remove(message_id) {
-            Some(client_request) => {
+        for (_, account) in self.accounts.clone().iter_mut() {
```
maqi: iterate through every account will be costly. can we do ```account = self.account.get_mut(request.dest.name())``` here?

src/error.rs
```diff
@@ -29,6 +29,7 @@ pub enum ClientError {
     NoSuchData,
     DataExists,
     LowBalance,
+    DuplicateRequest,
```
maqi: if this error is not being sent back to the client, shall it be defined under InternalError ?

src/personas/maid_manager.rs
brian-js: Don't really know what your saying here Qi. SD go to SDM and logic hasn't changed, the request was being cached.

maqi: SD with tpye_tag=0 is only used for creating account in MM, and such SD doesn't get forwarded to SDM, which means there won't be any response.

src/personas/maid_manager.rs
brian-js: What request are we talking about here?

maqi: sorry, then that shall be in vault.rs change to something like : 
```
            (&Authority::NaeManager(_),
             &Authority::ClientManager(dst),
             &ResponseContent::PutFailure{ ref id, ref external_error_indicator, .. }) => {
                self.maid_manager.handle_put_failure(routing_node, dst, id, external_error_indicator)
```
then here we can do
```
account = self.account.get_mut(dst.name())
```

src/personas/maid_manager.rs
brian-js: That's not what's implemented currently. Are you referring to some part of the documentation?

maqi: Ah, sorry, my fault.
yes, you are correct.

src/personas/maid_manager.rs
brian-js: Are we expecting the MM to be catering for a large number of client accounts at any given time.

maqi: probably yes, if the vault lives a long enough time.  :(

src/error.rs
brian-js: I wonder if we should send a failure with this error back to the client?

src/personas/maid_manager.rs
brian-js: I think we can leave this as it is here and take a look later as a potential performance improvement.

src/error.rs
maqi: probably should.
then could you please put the code in to send this error back to the client?
thx

maqi:
Reviewed 2 of 3 files at r1, 1 of 1 files at r2.
Review status: all files reviewed at latest revision, 3 unresolved discussions, some commit checks broke.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/maidsafe/safe_vault/345)*

PR #348, opened 2016-03-02
CI tests Debugging

maqi:
[Review on Reviewable](https://reviewable.io/reviews/maidsafe/safe_vault/348)


maidsafe-highfive:
r? @Fraser999

src/personas/maid_manager.rs
```diff
@@ -181,6 +181,7 @@ impl MaidManager {
                              account.put_data(DEFAULT_PAYMENT /* data.payload_size() as u64 */)
                          });
         if let Err(error) = result {
+            trace!("MM sending put failure for immutable_data {:?} with error {:?}", data.name(), error);
```
Fraser999: Just for info, `XorName` now has a `Display` implementation, so we can use `{}` rather than `{:?}`.

tests/detail/test_cases/immutable_data.rs
```diff
@@ -45,6 +45,8 @@ pub fn test(max_get_attempts: u32) {
         ResponseMessage { content: ResponseContent::PutSuccess(..), .. } => {}
         _ => panic!("Received unexpected response"),
     }
+    // an interval allows pmid_node storing data
+    ::std::thread::sleep(::std::time::Duration::from_secs(5));
```
Fraser999: Can this be removed please?

tests/detail/test_cases/immutable_data_churn.rs
```diff
@@ -26,9 +26,9 @@ pub fn test(request_count: u32, max_get_attempts: u32) {
     client.create_account();
     let mut stored_data = Vec::with_capacity(request_count as usize);
     for i in 0..request_count {
-        test_group.start_case(&format!("Put ImmutableData {}", i));
         let data = Data::Immutable(ImmutableData::new(ImmutableDataType::Normal,
                                                       generate_random_vec_u8(1024)));
+        test_group.start_case(&format!("Put ImmutableData {} - {:?}", i, data.name()));
```
Fraser999: Can you undo these changes please (the ones with `start_case`)?  If we need the data name or other details from the test, we can use logging.  I'd rather the test names weren't extended like this, and also it's best that they properly reflect the point at which the test case starts.

src/personas/pmid_node.rs
```diff
@@ -47,16 +47,27 @@ impl PmidNode {
...
+            if let Ok(decoded) = serialisation::deserialise::<ImmutableData>(&data) {
+                let immutable_data = Data::Immutable(decoded);
+                trace!("As {:?} sending data {:?} to {:?}",
+                       request.dst,
+                       immutable_data,
+                       request.src);
+                let _ = routing_node.send_get_success(request.dst.clone(),
+                                                      request.src.clone(),
+                                                      immutable_data,
+                                                      message_id.clone());
```
Fraser999: Isn't there a `return` missing here?

tests/detail/test_cases/immutable_data.rs
maqi: not for my machine to get test passing.  :(

src/personas/pmid_node.rs
maqi: sorry, and added.

tests/detail/test_cases/immutable_data.rs
Fraser999: We can set `RUST_LOG=ci=info` without affecting the running Vaults though.

tests/detail/test_cases/immutable_data_churn.rs
```diff
@@ -49,4 +51,4 @@ pub fn test(request_count: u32, max_get_attempts: u32) {
     }
 
     test_group.release();
-}
+}
```
Fraser999: Is your editor set up to strip trailing newline at the end of file?  There are two files like this.

maqi: nope.
 it was raw file from github I used to restore the change, and seems the raw file will trailing .   :(

tests/detail/test_cases/immutable_data.rs
maqi: sorry I don't get it. 
you mean using `RUST_LOG=ci=info` only so the execution of vault can be speed up?
But I think the purpose of the CI test is to simulate the real scenario.
The Get request is one hop quicker than Put requests, and the timing gap does exists.
I think the sleep here will expose and highlight this.

Fraser999: I thought you meant logging in Vaults caused the tests to fail on your machine, so I was suggesting avoiding logging at too high a level and/or log for the ci target only, not Vaults and Routing too.

As for the sleep, we're going to be returning failure messages from the network once https://github.com/maidsafe/safe_vault/pull/350 gets merged, so there's no need for a sleep here along with the client retry mechanism we've put in place I don't think?

maqi: tried setting `RUST_LOG=ci=info` still get test failed, if this sleep removed.

Regarding the retry mechanism, it won't work for CI test in this scenario. When the failure happens on my machine, it is almost all of the picked-up nodes will sending back get_failure, this will trigger the majority of the nodes in the network to be marked as BAD dataholder for that data, and also no where to replicate and restore. When the re-try attempts comes in, there is no Good data-holder to visit.

this might be an extreme case, but it does happen with the CI test, which only have tiny number of nodes.  :(

Fraser999: OK - so the logging thing wasn't meant as a comment about the test passing/failing - only a request to avoid changing the test names when we can do this with log messages in the test code itself.  I've added the comment to the wrong thread - sorry!

As for the sleep, I really think it should be removed, since we're not expecting users of vaults to have to do a sleep.  The issues that are being exposed are valid failure cases, and I think we should fix them in the Vault code.  We can perhaps discuss this in the Vault slack channel if we need more input from others.

maqi:
replaced by the PR https://github.com/maidsafe/safe_vault/pull/351 to squash down the number of commits.

PR #357, opened 2016-03-06
Cache immutable data on put until desired redundancy is met.

brian-js:
[Review on Reviewable](https://reviewable.io/reviews/maidsafe/safe_vault/357)

### DIFF
--- a/src/personas/immutable_data_manager.rs
+++ b/src/personas/immutable_data_manager.rs
@@ -201,8 +201,15 @@ impl ImmutableDataManager {
     pub fn handle_put(&mut self,
                       routing_node: &RoutingNode,
                       data: &ImmutableData,
-                      message_id: &MessageId)
+                      message_id: &MessageId,
+                      request: &RequestMessage)
                       -> Result<(), InternalError> {
+        // Send success on receipt.
+        let src = request.dst.clone();
+        let dst = request.src.clone();
+        let message_hash = sha512::hash(&try!(serialisation::serialise(&request))[..]);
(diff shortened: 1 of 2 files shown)


maidsafe-highfive:
r? @Fraser999

src/personas/immutable_data_manager.rs
```diff
@@ -139,6 +140,7 @@ pub struct ImmutableDataManager {
     accounts: HashMap<XorName, Account>,
     // key is chunk_name
     ongoing_gets: LruCache<XorName, MetadataForGetRequest>,
+    ongoing_puts: HashMap<MessageId, ImmutableData>,
```
maqi: In case of churn, shall this ongoing_puts to be filterred as well for none management ?
otherwise there will be staled list keeps growing up.

Viv-Rajkumar: soz @maqi, what do you mean by `ongoing_puts to be filterred as well for none management`. We can't use a LruCache here even if we wanted to I'd guess as if we just drop these on a timeout without confirming the store from the next Persona, we will possibly be loosing data. Clearing of this list I'll assume needs to be glued in when we have confirmation from the redundancy number of groups that the data is stored with the ManagedNode they are representing.

maqi: When this IDM is no longer in the close group of the data, it will never receives any confirmation (success / failure) from PM.

Fraser999: Good point @maqi - I think that's correct.

Fraser999:
Now incorporated into https://github.com/maidsafe/safe_vault/pull/358.

PR #358, opened 2016-03-07
Cache immutable data on put until desired redundancy is met.

Fraser999:
[Review on Reviewable](https://reviewable.io/reviews/maidsafe/safe_vault/358)

### DIFF
--- a/src/personas/immutable_data_manager.rs
+++ b/src/personas/immutable_data_manager.rs
@@ -47,7 +47,9 @@ pub enum DataHolder {
 impl DataHolder {
     pub fn name(&self) -> &XorName {
         match *self {
-            DataHolder::Good(ref name) | DataHolder::Failed(ref name) | DataHolder::Pending(ref name) => name,
+            DataHolder::Good(ref name) |
+            DataHolder::Failed(ref name) |
+            DataHolder::Pending(ref name) => name,
         }
     }
 }
@@ -67,7 +69,10 @@ impl MetadataForGetRequest {
         Self::construct(vec![], pmid_nodes)
     }
 
-    pub fn with_message(message_id: &MessageId, request: &RequestMessage, pmid_nodes: &Account) -> MetadataForGetRequest {
+    pub fn with_message(message_id: &MessageId,
+                        request: &RequestMessage,
+                        pmid_nodes: &Account)
+                        -> MetadataForGetRequest {
         Self::construct(vec![(message_id.clone(), request.clone()); 1], pmid_nodes)
     }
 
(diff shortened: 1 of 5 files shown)


maidsafe-highfive:
r? @maqi

src/personas/immutable_data_manager.rs
```diff
@@ -139,6 +127,7 @@ pub struct ImmutableDataManager {
     accounts: HashMap<XorName, Account>,
     // key is chunk_name
     ongoing_gets: LruCache<XorName, MetadataForGetRequest>,
+    ongoing_puts: HashMap<MessageId, ImmutableData>,
```
maqi: As discussed in the PR https://github.com/maidsafe/safe_vault/pull/357
shall we use LruCache, or filter in churn, or just ignore this currently?

src/personas/immutable_data_manager.rs
```diff
@@ -333,6 +350,86 @@ impl ImmutableDataManager {
...
 
+    pub fn handle_put_success(&mut self,
+                              pmid_node: &XorName,
+                              message_id: &MessageId)
+                              -> Result<(), InternalError> {
+        let mut replicants_stored = 0;
+
+        {
+            if let Some(immutable_data) = self.ongoing_puts.get(message_id) {
+                if let Some(pmid_nodes) = self.accounts.get_mut(&immutable_data.name()) {
```
maqi: the status of data_holders in account is now depends on the on_going_puts cache heavily.
However, on_going_puts cache is not refreshed in/out during churn.
This will causing issues that IDM lost access to update the status of data_holders in account : 
when a new vault joining, it received account (having pending nodes), however, as it doesn't receive the on_going_put cache, when it received PutSuccess/PutFailure from PM, it can not figure out which record it relates to.

Fraser999: Yeah - we're not wanting these short-term items refreshed for now until we find out if they need to be.

src/personas/immutable_data_manager.rs
Fraser999: I think we're ignoring this for now.  To fix it, we'll leave it as a `HashMap`, but we'll prune invalid entries when we get a `NodeAdded` or `NodeLost` event.

maqi:
Reviewed 5 of 5 files at r2.
Review status: all files reviewed at latest revision, 2 unresolved discussions.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/maidsafe/safe_vault/358)*

PR #361, opened 2016-03-08
Changed to remove accounts for which personas are no longer responsible after churn event.

Fraser999:
[Review on Reviewable](https://reviewable.io/reviews/maidsafe/safe_vault/361)

### DIFF
--- a/src/personas/immutable_data_manager.rs
+++ b/src/personas/immutable_data_manager.rs
@@ -429,15 +429,17 @@ impl ImmutableDataManager {
         let _ = self.accounts.insert(data_name, account);
     }
 
-    pub fn handle_node_added(&mut self, routing_node: &RoutingNode, _node_added: XorName) {
-        self.handle_churn(routing_node)
+    pub fn handle_node_added(&mut self, routing_node: &RoutingNode, node_added: XorName) {
+        let churn_message_id = MessageId::from_added_node(node_added);
+        self.handle_churn(routing_node, churn_message_id)
     }
 
-    pub fn handle_node_lost(&mut self, routing_node: &RoutingNode, _node_lost: XorName) {
-        self.handle_churn(routing_node)
+    pub fn handle_node_lost(&mut self, routing_node: &RoutingNode, node_lost: XorName) {
+        let churn_message_id = MessageId::from_lost_node(node_lost);
+        self.handle_churn(routing_node, churn_message_id)
     }
 
-    pub fn handle_churn(&mut self, routing_node: &RoutingNode) {
+    fn handle_churn(&mut self, routing_node: &RoutingNode, churn_message_id: MessageId) {
         // Only retain accounts for which we're still in the close group
         let accounts = mem::replace(&mut self.accounts, HashMap::new());
         self.accounts =
@@ -513,8 +515,7 @@ impl ImmutableDataManager {
                 // Create a new entry and send Get requests to each of the current holders
                 let entry = MetadataForGetRequest::new(pmid_nodes);
                 trace!("Created ongoing get entry for {} - {:?}", data_name, entry);
-                let message_id = MessageId::new();
-                entry.send_get_requests(routing_node, data_name, &message_id);
+                entry.send_get_requests(routing_node, data_name, &churn_message_id);
                 let _ = self.ongoing_gets.insert(data_name.clone(), entry);
             }
             self.send_refresh(routing_node, data_name, pmid_nodes);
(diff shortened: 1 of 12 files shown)


maidsafe-highfive:
r? @maqi

src/personas/pmid_node.rs
```diff
@@ -145,20 +153,45 @@ impl PmidNode {
...
-                                      data_name: &XorName,
-                                      message_id: &MessageId,
-                                      request: &RequestMessage) -> Result<(), InternalError> {
+    pub fn handle_churn(&mut self, routing_node: &RoutingNode) {
+        // Only retain chunks for which we're still in the close group
+        let chunk_names = self.chunk_store.names();
+        for chunk_name in chunk_names {
+            match routing_node.close_group(chunk_name) {
+                Ok(None) => {
+                    trace!("No longer a PN for {}", chunk_name);
```
maqi: when IDM picking up the PNs, it is not based on the rule of close_group.
I think on PN, such removal on churn shall not be carried out.
also, this cause additional chance of in-consistency with DM's records.

Fraser999: Can you bring this up in the Vault channel please?  I think what's here is what we want, but it'd be good to involve the Routing guys too.

src/personas/pmid_node.rs
```diff
@@ -145,20 +153,45 @@ impl PmidNode {
...
+                    let _ = self.chunk_store.delete(&chunk_name);
```
maqi: In case a deletion does need to be carryied out here, shall a notification of removal being sent to the PM and IDM ?

src/personas/pmid_node.rs
maqi: I notice your change in IDM that will filter out non-closegropu pmid_nodes from the record in case of churn.
This helps reduce IDM's dependency on client's get request to detecting failed node.
regarding the non-consistency of the view of close_group, I think some solid data from tests need to be collected first to make it convincing.

src/personas/immutable_data_manager.rs
```diff
@@ -434,62 +429,65 @@ impl ImmutableDataManager {
...
-                        })
-                        .collect();
-        }
+                            Ok(Some(close_group)) => close_group.into_iter().collect(),
+                            Err(error) => {
+                                error!("Failed to get close group: {:?} for {}", error, data_name);
+                                return None;
+                            }
+                        };
+                        pmid_nodes = pmid_nodes.into_iter()
```
maqi: for those nodes that have now been filtered out, shall a notification to be sent to PM to update its account?
As, when PN removing such chunk, there won't be any notification sent out either.
This will leave PM's record inconsistent with the real storage.

src/personas/pmid_node.rs
Fraser999: So I guess we'll discuss here then :)

So far, the ImmutableData churn tests have passed 3 out of 3 times for me.  We need to be able to remove chunks for which we're no longer responsible, and while this exact implementation might need tweaked (we could possibly be removing the chunk even sooner than this) we do need to handle removing stale chunks.

What's implemented in this PR is what's described in the Vault flow document, and this is also what's implemented in the Routing example I believe (where it also works fine as far as I know).

If you still disagree, I really think we should involve more people than just the two of us.

src/personas/pmid_node.rs
Fraser999: No - I don't think that's needed - at least not yet.  If it is needed, we should discuss on the Vault slack channel and update the Vault flow doc first.

src/personas/immutable_data_manager.rs
Fraser999: I agree with this point, but again it should be discussed on the Slack channel and updated in the doc before we change the flow.

src/personas/pmid_node.rs
maqi: I pulled your PR to local and the ImmutableData churn tests have failed 3 out of 3 times for me. :(
Meanwhile, the churn test for SD and messaging get passed 3 out of 3 times.
which indicating a too radical movement towards removing ID related records in churn handling.

I am ok to have more people involved in the discussion. probably you can put a msg in slack and tag the guys interested?

maqi:
Reviewed 11 of 12 files at r1, 1 of 1 files at r2.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/maidsafe/safe_vault/361)*

PR #363, opened 2016-03-09
some logs for debugging

maqi:
This change is [Review on Reviewable](https://reviewable.io/reviews/maidsafe/safe_vault/363)

### DIFF
--- a/src/personas/maid_manager.rs
+++ b/src/personas/maid_manager.rs
@@ -189,6 +189,7 @@ impl MaidManager {
             unreachable!("Logic error")
         };
         let client_name = utils::client_name(&request.src);
+        trace!("MM received put request of data {} from client {}", data.name(), client_name);
         self.forward_put_request(routing_node, client_name, data, *message_id, request)
     }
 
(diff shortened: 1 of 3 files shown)


maidsafe-highfive:
r? @Fraser999

Fraser999:
I'm not sure about a lot of these Qi.  It looks like many will be pretty much just duplicating the trace message from the event loop in `vault::do_run`.  The only use I could see for these would be to check the demuxing is working properly in `vault::on_...`, so they'd be temporary only - not committed to the repo.

I think the ones which are showing how we're responding are potentially useful.

Can you keep the format consistent though please - we have some that have "SDM" and others have "StructuredDataManager" - I guess shorter is better for logging?

maqi:
removed some duplicated logging and making the pattern consistent (MM, SDM and PM)

Fraser999:
Reviewed 3 of 3 files at r2.
Review status: all files reviewed at latest revision, all discussions resolved, some commit checks broke.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/maidsafe/safe_vault/363)*

PR #364, opened 2016-03-09
Reduced cyclomatic complexity for some functions to remove final Clippy warnings.

Fraser999:
This change is [Review on Reviewable](https://reviewable.io/reviews/maidsafe/safe_vault/364)

### DIFF
--- a/src/personas/mpid_manager.rs
+++ b/src/personas/mpid_manager.rs
@@ -22,7 +22,7 @@ use default_chunk_store;
 use error::{ClientError, InternalError};
 use maidsafe_utilities::serialisation::{deserialise, serialise};
 use mpid_messaging::{MAX_INBOX_SIZE, MAX_OUTBOX_SIZE, MpidHeader, MpidMessage, MpidMessageWrapper};
-use routing::{Authority, Data, PlainData, RequestContent, RequestMessage};
+use routing::{Authority, Data, MessageId, PlainData, RequestContent, RequestMessage};
 use sodiumoxide::crypto::sign::PublicKey;
 use sodiumoxide::crypto::hash::sha512;
 use types::{Refresh, RefreshValue};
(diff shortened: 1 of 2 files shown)


maidsafe-highfive:
r? @maqi

src/personas/mpid_manager.rs
```diff
@@ -299,257 +213,95 @@ impl MpidManager {
...
-                                                                  client.clone(),
-                                                                  request.clone(),
-                                                                  Vec::new(),
-                                                                  *message_id);
-                        }
-                    }
-                }
-            }
+        let wrapper: MpidMessageWrapper = try!(deserialise(&data.value()));
+        let mpid_header = if let MpidMessageWrapper::PutHeader(mpid_header) = wrapper {
```
maqi: well, i am not sure, but seems to me changing to this new format is more painful to read. :(

maqi:
Reviewed 2 of 2 files at r1.
Review status: all files reviewed at latest revision, 1 unresolved discussion, some commit checks broke.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/maidsafe/safe_vault/364)*

src/personas/mpid_manager.rs
Fraser999: Well, I tend to think that more than two or three levels of nesting are poor form.  I'm sure Viv likes returning early from functions at every opportunity too, but maybe others disagree.

I'm generally in favour of returning early, but I'm more in favour of consistency, and I've certainly been trying to return early rather than create new levels of nesting wherever possible throughout the rest of the code I've worked on.

I guess if it's a blocker to getting the PR merged, we can take the discussion to the rust Slack channel?

maqi: regarding early return, the current nesting is also early returned (though not calling return explicitly).

However, if nesting is a concern, then that's another story.

Fraser999: Yeah - it's the explicit calling of return to avoid nesting I meant.

PR #366, opened 2016-03-10
Minor bug fix and tests.

brian-js:
This change is [Review on Reviewable](https://reviewable.io/reviews/maidsafe/safe_vault/366)

### DIFF
--- a/src/personas/maid_manager.rs
+++ b/src/personas/maid_manager.rs
@@ -124,13 +124,12 @@ impl MaidManager {
         match self.request_cache.remove(message_id) {
             Some(client_request) => {
                 // Refund account
-                match self.accounts.get_mut(client_request.dst.name()) {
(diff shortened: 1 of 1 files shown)


maidsafe-highfive:
r? @maqi

maqi:
Reviewed 1 of 1 files at r2.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/maidsafe/safe_vault/366)*

maqi:
Put of normal SD not included will be covered in another PR.
And, MM don't expect a put_failure from IDM, shall be only from SDM.
However, not sure whether put_failure from IDM shall be blocked explicitly (handle_put_failure shall return with internal error), or like what current tests exposed that it is to be tolerated.
This better to be left to a separate PR to cover.

PR #367, opened 2016-03-11
Completed churn-handling for IDM.

Fraser999:
This change is [Review on Reviewable](https://reviewable.io/reviews/maidsafe/safe_vault/367)

### DIFF
--- a/src/personas/immutable_data_manager.rs
+++ b/src/personas/immutable_data_manager.rs
@@ -486,10 +486,8 @@ impl ImmutableDataManager {
                     .collect();
 
         for (data_name, pmid_nodes) in &self.accounts {
-            if pmid_nodes.len() < MIN_REPLICANTS {
-                trace!("Need to replicate {} as only {} holders left",
-                       data_name,
-                       pmid_nodes.len());
+            if Self::should_replicate(pmid_nodes) {
+                trace!("Need to replicate {}", data_name);
                 {
                     if let Some(mut metadata) = self.ongoing_gets.get_mut(&data_name) {
                         trace!("Already getting {} - {:?}", data_name, metadata);
@@ -515,6 +513,17 @@ impl ImmutableDataManager {
             }
             self.send_refresh(routing_node, data_name, pmid_nodes);
         }
+    }
+
+    fn should_replicate(account: &Account) -> bool {
+        let mut holder_count = 0;
+        for pmid_node in account {
+            match *pmid_node {
+                DataHolder::Pending(_) | DataHolder::Good(_) => holder_count += 1,
+                DataHolder::Failed(_) => (),
+            }
+        }
+        holder_count < MIN_REPLICANTS
     }
 
     fn send_refresh(&self, routing_node: &RoutingNode, data_name: &XorName, pmid_nodes: &Account) {
--- a/src/personas/immutable_data_manager.rs
+++ b/src/personas/immutable_data_manager.rs
@@ -707,6 +707,12 @@ impl ImmutableDataManager {
     }
 }
 
+impl Default for ImmutableDataManager {
+    fn default() -> ImmutableDataManager {
+        ImmutableDataManager::new()
+    }
+}
+
 
 
 #[cfg(all(test, feature = "use-mock-routing"))]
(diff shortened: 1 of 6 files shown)


maidsafe-highfive:
r? @maqi

src/personas/immutable_data_manager.rs
```diff
@@ -426,94 +431,174 @@ impl ImmutableDataManager {
...
+               data_name,
+               pmid_nodes);
+        // This function is used to filter accounts for which this node is no longer responsible, so
+        // return `None` in this case
+        let close_group = if let Some(group) = self.close_group_to(routing_node, &data_name) {
+            group
+        } else {
+            // Remove entry from `ongoing_puts`, as we're not part of the IDM group any more
+            let ongoing_puts = mem::replace(&mut self.ongoing_puts, HashMap::new());
+            self.ongoing_puts = ongoing_puts.into_iter()
```
maqi: will this be more efficient than `self.ongoing_puts.remove(data_name)`  ?

src/personas/immutable_data_manager.rs
```diff
@@ -426,94 +431,174 @@ impl ImmutableDataManager {
...
+
+    fn handle_churn_for_ongoing_gets(&mut self,
+                                     data_name: &XorName,
+                                     close_group: &[XorName])
+                                     -> bool {
+        if let Some(mut metadata) = self.ongoing_gets.get_mut(&data_name) {
+            trace!("Already getting {} - {:?}", data_name, metadata);
+            // Remove any holders which no longer belong in the cache entry
+            metadata.pmid_nodes
+                    .retain(|pmid_node| close_group.contains(pmid_node.name()));
```
maqi: It was observed in the previous test that there is chance the pmid_node list here to be empty.
As get_mut call will refresh the timestamp of that entry in lru_time_cache, this empty entry will sit in the cache forever.
Though, I am not sure whether this situation will happen with the new code or not?
shall a check shall be considered to be put?

src/personas/maid_manager.rs
```diff
@@ -238,7 +238,9 @@ impl MaidManager {
                              account.put_data(DEFAULT_PAYMENT /* data.payload_size() as u64 */)
                          });
         if let Err(error) = result {
-            trace!("MM responds put_failure of data {}, due to error {:?}", data.name(), error);
+            trace!("MM responds put_failure of data {}, due to error {:?}",
```
maqi: why clippy complains about this line not the line after?
the next line is even longer.   :(

src/personas/immutable_data_manager.rs
Fraser999: `ongoing_puts` is a `HashMap` indexed by `MessageId`, so we can't use the `data_name` like that.

src/personas/maid_manager.rs
Fraser999: Yeah - I know (although it's rustfmt that does this - not Clippy)  It doesn't seem to handle logging macros at all well.

src/personas/immutable_data_manager.rs
Fraser999: I don't think so - that was a bug previously.  We shouldn't have the situation where `metadata.pmid_nodes` is empty unless all holders churn out before we can replicate.

src/personas/immutable_data_manager.rs
brian-js: What about using the `MessageId` though?

Fraser999: It's a different ID.  The one passed into this function is derived from the name of the node which caused the churn event, whereas the one in the `ongoing_puts` is the original client-generated one.

brian-js:
Reviewed 5 of 6 files at r1, 1 of 1 files at r2.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/maidsafe/safe_vault/367)*

brian-js:
Reviewed 5 of 6 files at r1, 1 of 1 files at r2.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/maidsafe/safe_vault/367)*

brian-js:
Reviewed 1 of 6 files at r1.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/maidsafe/safe_vault/367)*

PR #371, opened 2016-03-14
WIP IDM only prune lost data_holder

maqi:
This change is [Review on Reviewable](https://reviewable.io/reviews/maidsafe/safe_vault/371)

### DIFF
--- a/src/personas/immutable_data_manager.rs
+++ b/src/personas/immutable_data_manager.rs
@@ -431,14 +431,17 @@ impl ImmutableDataManager {
     }
 
     pub fn handle_node_added(&mut self, routing_node: &RoutingNode, node_added: XorName) {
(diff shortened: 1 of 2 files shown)


maidsafe-highfive:
r? @Fraser999

Fraser999:
This doesn't cater for PNs which move out of the close group for the data because of new nodes being added.  It also removes the deletion of chunks from PNs which are no longer valid chunk holders.

maqi:
The aim of this PR is to reduce the complexity on the vault level, so a better performance can be achieved.
The profiling test proved a performance gain of 100% with this approach.

As PN is no longer checked against the close group, so the above concern is no longer valid.
i.e. a PN is always a valid data_holder as long as the IDM having the connection to it.

However, I am not going to re-open this PR, unless the performance starts being a real concern.

PR #375, opened 2016-03-16
Cleanup

Fraser999:
This change is [Reviewable](https://reviewable.io/reviews/maidsafe/safe_vault/375)


maidsafe-highfive:
r? @maqi

src/personas/maid_manager.rs
```diff
@@ -487,7 +495,8 @@ mod test {
         // put_requests[0] - account creation.
         assert_eq!(put_requests.len(), 2);
         assert_eq!(put_requests[1].src, env.our_authority);
-        assert_eq!(put_requests[1].dst, Authority::NaeManager(immutable_data.name()));
```
maqi: is this complained by rustfmt or clippy?
I think it is definitely less than 100 characters.
does it meant we are switching to 80 characters rule,
or just for assert_eq macro, it must be below 80 characters?

Fraser999: It's rustfmt.  It seems to have a random approach to handling macros.

maqi:
Reviewed 2 of 5 files at r1, 3 of 3 files at r3.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/maidsafe/safe_vault/375)*

PR #377, opened 2016-03-18
Integrating with common

ustulation:
This change is [Reviewable](https://reviewable.io/reviews/maidsafe/safe_vault/377)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,19 +11,22 @@ version = "0.4.0"
 
 [dependencies]
 chunk_store = "~0.3.0"
-clippy = {version = "~0.0.51", optional = true}
 config_file_handler = "~0.2.0"
 ctrlc = "~1.1.1"
 docopt = "~0.6.78"
 log = "~0.3.5"
 lru_time_cache = "~0.2.7"
 maidsafe_utilities = "~0.4.0"
-mpid_messaging = "~0.2.0"
 routing = "~0.11.1"
 rustc-serialize = "~0.3.18"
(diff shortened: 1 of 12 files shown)


maidsafe-highfive:
r? @maqi

Cargo.toml
```diff
@@ -11,20 +11,23 @@ version = "0.4.0"
...
 maidsafe_utilities = "~0.4.0"
-mpid_messaging = "~0.2.0"
 routing = "~0.11.1"
 rustc-serialize = "~0.3.18"
+safe_network_common = "~0.0.1"
 sodiumoxide = "~0.0.9"
 time = "~0.1.34"
 xor_name = "~0.1.0"
 
+[dependencies.clippy]
```
Fraser999: Can you revert this style please?  All other repos are using the single line style.  Also I think we're up to 0.0.54 (although that might have changed by the time I write this :)  ).

src/main.rs
```diff
@@ -61,6 +60,7 @@ extern crate rustc_serialize;
 extern crate sodiumoxide;
 extern crate time;
 extern crate xor_name;
+extern crate safe_network_common;
```
maqi: i think it is preferred that crates listed in an alphabet order?

maqi:
Reviewed 1 of 12 files at r1, 8 of 11 files at r2.
Review status: all files reviewed at latest revision, 2 unresolved discussions.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/maidsafe/safe_vault/377)*

PR #379, opened 2016-03-18
Reduced timeout for PM put_request cache in test.

Fraser999:
This change is [Reviewable](https://reviewable.io/reviews/maidsafe/safe_vault/379)


maidsafe-highfive:
r? @maqi

dirvine:
Why does this timeout exist? Are we not guaranteed a response otherwise something else is wrong?

Fraser999:
No, this is the timeout which was agreed to in the PMs for Put requests.  We need it in case the PN doesn't respond to a Put.  It's also used to clear entries out of the `ongoing_puts` map which won't resolve since we've been pushed out of the close group for that PN due to churn.

maqi:
Reviewed 1 of 1 files at r1.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/maidsafe/safe_vault/379)*

PR #386, opened 2016-03-25
CI tests: Use mock crust as feature

brian-js:
Enable running tests with use-mock-crust feature on.

+@dirvine +@Fraser999


maidsafe-highfive:
r? @maqi

PR #387, opened 2016-03-25
config_file for vault

maqi:
put in new files config_handler.rs and sample.config for testing purpost, but also minor impacts on other files.

This allows vault to be personalized using a config file.
So far, it is just account_size in the ClientManager enabled.

---
This change is [Reviewable](https://reviewable.io/reviews/maidsafe/safe_vault/387) &emsp; Multiple assignees:&emsp;@Fraser999&nbsp;<a href="/maidsafe/safe_vault/pulls/assigned/Fraser999">Fraser999</a>,&emsp;@brian-js&nbsp;<a href="/maidsafe/safe_vault/pulls/assigned/brian-js">brian-js</a>,&emsp;@dirvine&nbsp;<a href="/maidsafe/safe_vault/pulls/assigned/dirvine">dirvine</a>


maidsafe-highfive:
r? @Fraser999

maqi:
+@dirvine  +@brian-js  +@fraser999 

---







*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/maidsafe/safe_vault/387#-:-KDhvXuCHw4JB6rO9Pob:-2113424343)*

dirvine:
Review status: 0 of 6 files reviewed at latest revision, 1 unresolved discussion.

---

*[installer/sample.config, line 4 \[r1\]](https://reviewable.io:443/reviews/maidsafe/safe_vault/387#-KDhxAraG6T7ruRBAQaD:-KDhxArbDRWcEbeeD0gF:1486392089) ([raw file](https://github.com/maidsafe/safe_vault/blob/7795e6b6f848a27cce24faa8d39829db4c72ef80/installer/sample.config#L4)):*
Can you please describe what entries limit is and also chunk_store_limit. The more we have here the harder it is to use. 

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/maidsafe/safe_vault/387)*

maqi:
Review status: 0 of 6 files reviewed at latest revision, 1 unresolved discussion.

---

*[installer/sample.config, line 4 \[r1\]](https://reviewable.io:443/reviews/maidsafe/safe_vault/387#-KDhxAraG6T7ruRBAQaD:-KDhxoi3PY8Zg4UY-nhK:-161399410) ([raw file](https://github.com/maidsafe/safe_vault/blob/7795e6b6f848a27cce24faa8d39829db4c72ef80/installer/sample.config#L4)):*
there is task `Test filling the testnet with chunks and seeing Clients are properly denied write-access`, which will need such limit to be available.
i am ok to remove them now, and leave them to be added in the later on tasks.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/maidsafe/safe_vault/387)*

Fraser999:
Review status: 0 of 6 files reviewed at latest revision, 6 unresolved discussions.

---

*[src/config_handler.rs, line 22 \[r1\]](https://reviewable.io:443/reviews/maidsafe/safe_vault/387#-KDhzNIyPClNt1VhjFMm:-KDhzNIyPClNt1VhjFMn:-1841325670) ([raw file](https://github.com/maidsafe/safe_vault/blob/7795e6b6f848a27cce24faa8d39829db4c72ef80/src/config_handler.rs#L22)):*
Why do we need `PartialEq`, `Eq` and `Clone`?

---

*[src/config_handler.rs, line 24 \[r1\]](https://reviewable.io:443/reviews/maidsafe/safe_vault/387#-KDhzek1yhPCUUMH-kuv:-KDhzek1yhPCUUMH-kuw:1264580723) ([raw file](https://github.com/maidsafe/safe_vault/blob/7795e6b6f848a27cce24faa8d39829db4c72ef80/src/config_handler.rs#L24)):*
This should be removed

---

*[src/config_handler.rs, line 26 \[r1\]](https://reviewable.io:443/reviews/maidsafe/safe_vault/387#-KDhzhWQsbGEIeRuX_Rl:-KDhzhWQsbGEIeRuX_Rm:1264580723) ([raw file](https://github.com/maidsafe/safe_vault/blob/7795e6b6f848a27cce24faa8d39829db4c72ef80/src/config_handler.rs#L26)):*
This should be removed

---

*[src/config_handler.rs, line 27 \[r1\]](https://reviewable.io:443/reviews/maidsafe/safe_vault/387#-KDhzm2dlwHrQDzyvzIM:-KDhzm2dlwHrQDzyvzIN:-1465475506) ([raw file](https://github.com/maidsafe/safe_vault/blob/7795e6b6f848a27cce24faa8d39829db4c72ef80/src/config_handler.rs#L27)):*
This might be better called `max_disk_space` or something like that - more understandable by user with no knowledge of Vault internals.

---

*[src/config_handler.rs, line 49 \[r1\]](https://reviewable.io:443/reviews/maidsafe/safe_vault/387#-KDi-Bnp9pgJusk3wjwP:-KDi-BnqOjjlYjC1UXiQ:2053427840) ([raw file](https://github.com/maidsafe/safe_vault/blob/7795e6b6f848a27cce24faa8d39829db4c72ef80/src/config_handler.rs#L49)):*
If this is for use by tests only, shouldn't it be in a `#[cfg(test)]` block?

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/maidsafe/safe_vault/387)*

Fraser999:
Review status: 0 of 6 files reviewed at latest revision, 7 unresolved discussions.

---

*[src/main.rs, line 125 \[r1\]](https://reviewable.io:443/reviews/maidsafe/safe_vault/387#-KDi-akRlMvrPoUYZq8B:-KDi-akRlMvrPoUYZq8C:869933238) ([raw file](https://github.com/maidsafe/safe_vault/blob/7795e6b6f848a27cce24faa8d39829db4c72ef80/src/main.rs#L125)):*
I feel it would be better to limit the scope of the config file handling to just vault.rs.  So leave `Vault::run()` as taking no params and handle inside there.  That way, I think we can limit "knowledge" of the config file to just that single source file.  We parse it in there, and have the c'tors of SDM and PN take a `u64` specifying their chunkstore limits.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/maidsafe/safe_vault/387)*

Fraser999:
Review status: 0 of 6 files reviewed at latest revision, 7 unresolved discussions.

---

*[installer/sample.config, line 4 \[r1\]](https://reviewable.io:443/reviews/maidsafe/safe_vault/387#-KDhxAraG6T7ruRBAQaD:-KDi0NFzm-LdVWVhCycC:390623148) ([raw file](https://github.com/maidsafe/safe_vault/blob/7795e6b6f848a27cce24faa8d39829db4c72ef80/installer/sample.config#L4)):*
We should be able to achieve that using only the chunkstore limit though shouldn't we?

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/maidsafe/safe_vault/387)*

dirvine:
Review status: 0 of 6 files reviewed at latest revision, 7 unresolved discussions.

---

*[installer/sample.config, line 4 \[r1\]](https://reviewable.io:443/reviews/maidsafe/safe_vault/387#-KDhxAraG6T7ruRBAQaD:-KDi28qt4lDLwkkv1Kh-:-806087159) ([raw file](https://github.com/maidsafe/safe_vault/blob/7795e6b6f848a27cce24faa8d39829db4c72ef80/installer/sample.config#L4)):*
I would leave on the the wallet and chunkstore limit, but call chunkstore limit - `vault max capacity` We can comment this is in Mb (max).

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/maidsafe/safe_vault/387)*

dirvine:
Review status: 0 of 6 files reviewed at latest revision, 8 unresolved discussions.

---

*[src/error.rs, line 43 \[r1\]](https://reviewable.io:443/reviews/maidsafe/safe_vault/387#-KDi2kGGm04t_6D0EkEx:-KDi2kGHYvba_bJseXdv:1386680489) ([raw file](https://github.com/maidsafe/safe_vault/blob/7795e6b6f848a27cce24faa8d39829db4c72ef80/src/error.rs#L43)):*
As we are OK with `quick_error` crate this can be simplified now. 

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/maidsafe/safe_vault/387)*

maqi:
*[src/error.rs, line 43 \[r1\]](https://reviewable.io:443/reviews/maidsafe/safe_vault/387#-KDi2kGGm04t_6D0EkEx:-KDi7lYbh9kKVQP32HOe:1900836624) ([raw file](https://github.com/maidsafe/safe_vault/blob/7795e6b6f848a27cce24faa8d39829db4c72ef80/src/error.rs#L43)):*
prefer leaving it as a separate PR

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/maidsafe/safe_vault/387)*

Fraser999:
Review status: 0 of 6 files reviewed at latest revision, 8 unresolved discussions.

---

*[installer/sample.config, line 4 \[r1\]](https://reviewable.io:443/reviews/maidsafe/safe_vault/387#-KDhxAraG6T7ruRBAQaD:-KDi82mNxACfqyJiM7b8:-1247777611) ([raw file](https://github.com/maidsafe/safe_vault/blob/7795e6b6f848a27cce24faa8d39829db4c72ef80/installer/sample.config#L4)):*
I'd suggested `max_disk_space` to be even more explicit to users.  We can't comment in JSON files afaik, but the file itself will be called `safe_vault.config` I think, so there's maybe no need for `vault_` in the variable name too?  I suppose eventually users won't be expected to modify this by hand, so these vars will be hidden from them anyway.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/maidsafe/safe_vault/387)*

maqi:
*[installer/sample.config, line 4 \[r1\]](https://reviewable.io:443/reviews/maidsafe/safe_vault/387#-KDhxAraG6T7ruRBAQaD:-KDi93Toy5HwLnWhGKV_:1602035369) ([raw file](https://github.com/maidsafe/safe_vault/blob/7795e6b6f848a27cce24faa8d39829db4c72ef80/installer/sample.config#L4)):*
changed to max_capacity

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/maidsafe/safe_vault/387)*

dirvine:
Review status: 0 of 9 files reviewed at latest revision, 7 unresolved discussions.

---

*[installer/sample.config, line 4 \[r1\]](https://reviewable.io:443/reviews/maidsafe/safe_vault/387#-KDhxAraG6T7ruRBAQaD:-KDi9kbx0kvVCCf5ZsU5:561370578) ([raw file](https://github.com/maidsafe/safe_vault/blob/7795e6b6f848a27cce24faa8d39829db4c72ef80/installer/sample.config#L4)):*
Good, yes we cannot (easily) comment in json files, so we will need to make a separate user available doc for this. Perhaps in vault main with `-- help` or similar? This is for another PR though, just to keep in mind. 

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/maidsafe/safe_vault/387)*

Fraser999:
Review status: 0 of 9 files reviewed at latest revision, 10 unresolved discussions.

---

*[src/config_handler.rs, line 51 \[r1\]](https://reviewable.io:443/reviews/maidsafe/safe_vault/387#-KDiDZmE1iaHeWss4nl0:-KDiDZmE1iaHeWss4nl1:-964422382) ([raw file](https://github.com/maidsafe/safe_vault/blob/7795e6b6f848a27cce24faa8d39829db4c72ef80/src/config_handler.rs#L51)):*
That URL won't work.  You'll need to specify http://maidsafe.net/config_file_handler/master/config_file_handler/fn.current_bin_dir.html or remove the link.

---

*[src/config_handler.rs, line 55 \[r1\]](https://reviewable.io:443/reviews/maidsafe/safe_vault/387#-KDiDzT41gAySW3WR5X8:-KDiDzT5XcO4Zys22hxJ:1190222170) ([raw file](https://github.com/maidsafe/safe_vault/blob/7795e6b6f848a27cce24faa8d39829db4c72ef80/src/config_handler.rs#L55)):*
Should maybe rephrase "installer for the dependent application" to "Vault's installer"

---

*[src/config_handler.rs, line 58 \[r1\]](https://reviewable.io:443/reviews/maidsafe/safe_vault/387#-KDiECy7IQX7tr1ydh0p:-KDiECy8W-YX1PYYlCdM:924054283) ([raw file](https://github.com/maidsafe/safe_vault/blob/7795e6b6f848a27cce24faa8d39829db4c72ef80/src/config_handler.rs#L58)):*
We have a few places using fully-qualified names.

---

*[src/default_chunk_store.rs, line 27 \[r2\]](https://reviewable.io:443/reviews/maidsafe/safe_vault/387#-KDiCQKwskNHW1Ljh_iY:-KDiCQKxg_T6QWs6rr96:1525436326) ([raw file](https://github.com/maidsafe/safe_vault/blob/4bc6ec0991f95ff8c44930f490191c8fc5239dc0/src/default_chunk_store.rs#L27)):*
Have you used tabs here rather than spaces?

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/maidsafe/safe_vault/387)*

Fraser999:
Review status: 0 of 9 files reviewed at latest revision, 11 unresolved discussions.

---

*[src/vault.rs, line 80 \[r2\]](https://reviewable.io:443/reviews/maidsafe/safe_vault/387#-KDiF6nleLBq_uAKhqma:-KDiF6nleLBq_uAKhqmb:2068686583) ([raw file](https://github.com/maidsafe/safe_vault/blob/4bc6ec0991f95ff8c44930f490191c8fc5239dc0/src/vault.rs#L80)):*
Shouldn't the total allocated == `config.max_capacity`?  This way, we're allowing up to 3x `config.max_capacity`.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/maidsafe/safe_vault/387)*

maqi:
*[src/vault.rs, line 80 \[r2\]](https://reviewable.io:443/reviews/maidsafe/safe_vault/387#-KDiF6nleLBq_uAKhqma:-KDiIheMGFDK_ZvnJ8Lh:-1163432747) ([raw file](https://github.com/maidsafe/safe_vault/blob/4bc6ec0991f95ff8c44930f490191c8fc5239dc0/src/vault.rs#L80)):*
i remember there was a discussion on this long time ago. it will be hard to predict the pattern of usage (how much to be used for each persona). So it was agreed previously that we just put a max cap and leave chunk_store to figure out whether to extremely using upto the cap or be gracefully to sharing with other.
This also applies to the situation when there is multiple vaults running on the same machine.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/maidsafe/safe_vault/387)*

Fraser999:
Review status: 0 of 9 files reviewed at latest revision, 11 unresolved discussions.

---

*[src/vault.rs, line 80 \[r2\]](https://reviewable.io:443/reviews/maidsafe/safe_vault/387#-KDiF6nleLBq_uAKhqma:-KDiLqedwONW22lWMsAX:-1129441829) ([raw file](https://github.com/maidsafe/safe_vault/blob/4bc6ec0991f95ff8c44930f490191c8fc5239dc0/src/vault.rs#L80)):*
Yes, but the chunkstore doesn't do that now.  I think we should just hard code a split for now.  Perhaps take this discussion to Vault Slack channel?

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/maidsafe/safe_vault/387)*

Fraser999:
Review status: 0 of 9 files reviewed at latest revision, 12 unresolved discussions.

---

*[src/config_handler.rs, line 1 \[r2\]](https://reviewable.io:443/reviews/maidsafe/safe_vault/387#-KDiXNHKlwlASQp5pnHn:-KDiXNHKlwlASQp5pnHo:414314622) ([raw file](https://github.com/maidsafe/safe_vault/blob/4bc6ec0991f95ff8c44930f490191c8fc5239dc0/src/config_handler.rs#L1)):*
This should be 2016.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/maidsafe/safe_vault/387)*

dirvine:
Reviewed 1 of 9 files at r2.
Review status: 0 of 9 files reviewed at latest revision, 12 unresolved discussions.

---

*[src/vault.rs, line 78 \[r4\]](https://reviewable.io:443/reviews/maidsafe/safe_vault/387#-KDjUjGVLwkKZD_Omd1F:-KDjUjGWAVSZYuA8qgUM:-1723967279) ([raw file](https://github.com/maidsafe/safe_vault/blob/b26ee786b7c4c802239b1c036c41664c54fd1e21/src/vault.rs#L78)):*
I am not sure of this. Can we estimate capacity better. Seems data manager will be a huge % of total. This will reduce it a lot. 

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/maidsafe/safe_vault/387)*

maqi:
*[src/vault.rs, line 78 \[r4\]](https://reviewable.io:443/reviews/maidsafe/safe_vault/387#-KDjUjGVLwkKZD_Omd1F:-KDjm-BUEoGVCm3KZU4y:-1419862691) ([raw file](https://github.com/maidsafe/safe_vault/blob/b26ee786b7c4c802239b1c036c41664c54fd1e21/src/vault.rs#L78)):*
then shall we have a separate config parameter for each persona?
or any calculation you have in mind?
say : 
max_idm_space = max_capacity/2
max_sdm_space = max_capacity/3
max_mpid_space = max_capacity/6  (it has two chunk_store inside, each will have max_capacity/12)

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/maidsafe/safe_vault/387)*

dirvine:
Review status: 0 of 9 files reviewed at latest revision, 12 unresolved discussions.

---

*[src/vault.rs, line 78 \[r4\]](https://reviewable.io:443/reviews/maidsafe/safe_vault/387#-KDjUjGVLwkKZD_Omd1F:-KDmvEmaJRzOhM1zvyWB:1230797429) ([raw file](https://github.com/maidsafe/safe_vault/blob/b26ee786b7c4c802239b1c036c41664c54fd1e21/src/vault.rs#L78)):*
I think we are only going to aske the user for one parameter. So we need to intelligently work this out. For now we could just take 
5% == DM
5% == MM
5% == PM
15% = Messaging 
70% == PN

Then work these out properly where these can steal from each other up to the total. So for now a breakdown like this, measure and then set a runtime ability to balance this out. Maybe an overall figure that is updated as we go along (a ChunkStore object may require to be shared across persona's). 

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/maidsafe/safe_vault/387)*

maqi:
*[src/vault.rs, line 78 \[r4\]](https://reviewable.io:443/reviews/maidsafe/safe_vault/387#-KDjUjGVLwkKZD_Omd1F:-KDsG8DLNmbjsyoJBzgd:1799064022) ([raw file](https://github.com/maidsafe/safe_vault/blob/b26ee786b7c4c802239b1c036c41664c54fd1e21/src/vault.rs#L78)):*
currently, IDM,MM,PM doesn't use churn_store, but SDM uses. so maybe a breakdown shall be set as : 
10% = Messaging (split equally for two internal chunk_store)
30% == SDM
60% == PN

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/maidsafe/safe_vault/387)*

dirvine:
Review status: 0 of 9 files reviewed at latest revision, 12 unresolved discussions.

---

*[src/vault.rs, line 78 \[r4\]](https://reviewable.io:443/reviews/maidsafe/safe_vault/387#-KDjUjGVLwkKZD_Omd1F:-KDsHMFJrYmb86dJWDM-:290748733) ([raw file](https://github.com/maidsafe/safe_vault/blob/b26ee786b7c4c802239b1c036c41664c54fd1e21/src/vault.rs#L78)):*
As a start during test then this will work. If we can log sizes to check the stats here. Later we will share a single chunk store object I think. 

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/maidsafe/safe_vault/387)*

dirvine:
Reviewed 1 of 6 files at r1, 3 of 9 files at r2, 1 of 2 files at r3, 2 of 3 files at r4, 2 of 2 files at r5.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/maidsafe/safe_vault/387)*

Fraser999:
Review status: all files reviewed at latest revision, 2 unresolved discussions, all commit checks successful.

---

*[src/vault.rs, line 73 \[r6\]](https://reviewable.io:443/reviews/maidsafe/safe_vault/387#-KDxoAuaisSzZxcrjyJU:-KDxoAuaisSzZxcrjyJV:354049935) ([raw file](https://github.com/maidsafe/safe_vault/blob/e82514507ebd97c6a863404370b87f4c2c1e6442/src/vault.rs#L73)):*
This could just be `let config = try!(config_handler::read_config_file());`

---

*[src/vault.rs, line 77 \[r6\]](https://reviewable.io:443/reviews/maidsafe/safe_vault/387#-KDxoGRbOR3iVt0lZIgx:-KDxoGRchiZMuMBvksx8:1649033461) ([raw file](https://github.com/maidsafe/safe_vault/blob/e82514507ebd97c6a863404370b87f4c2c1e6442/src/vault.rs#L77)):*
I think a better approach would be to define the default value for the max capacity in config_handler.rs.  Then the same ratios for chunkstore allocations could be done.  With the approach here, each chunkstore gets the same space if the config file doesn't specify a size.

There are serialisation/parsing/duplication issues with handling the default size in config_handler.rs, so an alternative would be to just handle this here in vault.rs, like:

```rust
let max_capacity = config.max_capacity.unwrap_or(DEFAULT_MAX_CAPACITY);
let pn_capacity = 3 * max_capacity / 5;
let sdm_capacity = 3 * max_capacity / 10;
let mpid_capacity = max_capacity / 10;
```

where `DEFAULT_MAX_CAPACITY` is `1_073_741_824`.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/maidsafe/safe_vault/387)*

PR #389, opened 2016-03-28
MaidManager update for ImmutableData.

brian-js:
MaidManager update for ImmutableData.

MaidManager should accept Put from Client for Normal ImmutableData only.
Also caters for a few requirements/tweaks to TimedBuffer.

1. Accepts Put for ImmutableData iff the data is Normal.
2. Updated TimedBuffer, in particular to suit current use case returns keys for expired entries.
2. Use TimedBuffer for MaidManager request_cache.
4. Adds check_timeout fn to MaidManager and calls from vault main loop.

Requires version change and publish for https://github.com/maidsafe/safe_core/pull/229 and https://github.com/maidsafe/safe_network_common/pull/5.

---
This change is [Reviewable](https://reviewable.io/reviews/maidsafe/safe_vault/389) &emsp; Multiple assignees:&emsp;@Fraser999&nbsp;<a href="/maidsafe/safe_vault/pulls/assigned/Fraser999">Fraser999</a>,&emsp;@dirvine&nbsp;<a href="/maidsafe/safe_vault/pulls/assigned/dirvine">dirvine</a>,&emsp;@maqi&nbsp;<a href="/maidsafe/safe_vault/pulls/assigned/maqi">maqi</a>


maidsafe-highfive:
r? @maqi

brian-js:
+@dirvine +@fraser999 

---







*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/maidsafe/safe_vault/389#-:-KDxBt3HF77GsRjkwsB_:-1865337078)*

dirvine:
Reviewed 1 of 3 files at r1.
Review status: 1 of 3 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[src/personas/maid_manager.rs, line 80 \[r1\]](https://reviewable.io:443/reviews/maidsafe/safe_vault/389#-KDxS91K9eNU39VcZF1c:-KDxS91L8jJCc6elKWdy:-47212769) ([raw file](https://github.com/maidsafe/safe_vault/blob/4e2aea967e729b44012eab909c024ba8f511a20c/src/personas/maid_manager.rs#L80)):*
I suspect this should not be a cache at all but a standard container. I doubt this should ever fail to return a success or error, unless network lost etc. Is it even worth putting this in a timer situation when it should never fail ?

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/maidsafe/safe_vault/389)*

Fraser999:
Review status: 1 of 3 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[src/personas/maid_manager.rs, line 80 \[r1\]](https://reviewable.io:443/reviews/maidsafe/safe_vault/389#-KDxS91K9eNU39VcZF1c:-KDxTNOQZFnSvD6N3C9I:-1514193662) ([raw file](https://github.com/maidsafe/safe_vault/blob/4e2aea967e729b44012eab909c024ba8f511a20c/src/personas/maid_manager.rs#L80)):*
A timed buffer was chosen since this holds a list of Client requests which must be answered (Vault is trying to provide a guarantee that every Client request will get a response).  The existing LRU cache just silently dropped these with no real way of monitoring failures.  If we use a standard container, we'd have a similar situation although slightly better since we would at least still have unanswered requests forever.

If we do decide that the Crust/Routing/Vault code is reliable enough to assume that a MM request to DM will always return properly, and we want to use a standard container, we'd have to prune entries which we're no longer responsible for during churn events.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/maidsafe/safe_vault/389)*

dirvine:
Review status: 1 of 3 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[src/personas/maid_manager.rs, line 80 \[r1\]](https://reviewable.io:443/reviews/maidsafe/safe_vault/389#-KDxS91K9eNU39VcZF1c:-KDxWZSlRnU0G7gMI6xw:1127935039) ([raw file](https://github.com/maidsafe/safe_vault/blob/4e2aea967e729b44012eab909c024ba8f511a20c/src/personas/maid_manager.rs#L80)):*
>If we do decide that the Crust/Routing/Vault code is reliable enough to assume that a MM request to DM will always return properly
We should not code around other libraries deficiencies though. If these libs do not carry out their API call properly we should complain to them. IF we do a workaround then we should mark as such IMHO

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/maidsafe/safe_vault/389)*

dirvine:
Reviewed 1 of 3 files at r2, 1 of 2 files at r3.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/maidsafe/safe_vault/389)*

PR #403, opened 2016-04-06
Allow network to report `NetworkFull` to Clients.

Fraser999:
Affects mainly immutable_data_manager.rs.

This implements a list of "full" PmidNodes at the IDM, so that they can block Put requests if too many close nodes are full.  No tests are in place for this at the moment.  The remaining changes are due to dependency updates and running rustfmt.

---
This change is [Reviewable](https://reviewable.io/reviews/maidsafe/safe_vault/403) &emsp; Multiple assignees:&emsp;@brian-js&nbsp;<a href="/maidsafe/safe_vault/pulls/assigned/brian-js">brian-js</a>,&emsp;@dirvine&nbsp;<a href="/maidsafe/safe_vault/pulls/assigned/dirvine">dirvine</a>,&emsp;@maqi&nbsp;<a href="/maidsafe/safe_vault/pulls/assigned/maqi">maqi</a>

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,23 +11,23 @@ version = "0.5.0"
 
 [dependencies]
 chunk_store = "~0.3.0"
-clippy = {version = "~0.0.59", optional = true}
+clippy = {version = "~0.0.61", optional = true}
 config_file_handler = "~0.3.0"
 ctrlc = "~1.1.1"
 docopt = "~0.6.78"
 log = "~0.3.5"
-maidsafe_utilities = "~0.5.0"
+maidsafe_utilities = "~0.5.1"
 routing = "~0.12.0"
-rustc-serialize = "~0.3.18"
-safe_network_common = "~0.0.1"
-sodiumoxide = "~0.0.9"
-time = "~0.1.34"
+rustc-serialize = "~0.3.19"
+safe_network_common = "~0.1.1"
+sodiumoxide = "~0.0.10"
+time = "~0.1.35"
 xor_name = "~0.1.0"
 
 [dev-dependencies]
 kademlia_routing_table = "~0.4.0"
 rand = "~0.3.14"
-safe_core = "~0.13.0"
+safe_core = "~0.13.1"
 
 [features]
 use-mock-routing = []
--- a/src/main.rs
+++ b/src/main.rs
@@ -126,7 +126,7 @@ pub fn main() {
     let underline = unwrap_result!(String::from_utf8(vec!['=' as u8; message.len()]));
     info!("\n\n{}\n{}", message, underline);
 
-    let mut vault = unwrap_result!(Vault::new(None));
+    let mut vault = unwrap_result!(Vault::new());
     unwrap_result!(vault.run());
 }
 
--- a/src/main.rs
+++ b/src/main.rs
@@ -133,5 +133,5 @@ pub fn main() {
 #[cfg(feature = "use-mock-crust")]
 #[allow(missing_docs)]
 pub fn main() {
-  println!("Error: mock crust not supported");
+    println!("Error: mock crust not supported");
 }
--- a/src/personas/immutable_data_manager.rs
+++ b/src/personas/immutable_data_manager.rs
@@ -238,8 +238,6 @@ pub struct ImmutableDataManager {
     // key is chunk_name
     ongoing_gets: TimedBuffer<XorName, MetadataForGetRequest>,
     ongoing_puts: HashSet<(ImmutableData, MessageId)>,
-    // key is PmidNode name; value is whether OK to store on (true) or Full (false)
-    known_pmid_nodes: HashMap<XorName, bool>,
 }
 
 impl ImmutableDataManager {
@@ -248,7 +246,6 @@ impl ImmutableDataManager {
             accounts: HashMap::new(),
             ongoing_gets: TimedBuffer::new(Duration::minutes(5)),
             ongoing_puts: HashSet::new(),
-            known_pmid_nodes: HashMap::new(),
         }
     }
 
@@ -312,6 +309,7 @@ impl ImmutableDataManager {
 
     pub fn handle_put(&mut self,
                       routing_node: &RoutingNode,
+                      full_pmid_nodes: &HashSet<XorName>,
                       request: &RequestMessage)
                       -> Result<(), InternalError> {
         let (data, message_id) = if let RequestContent::Put(Data::Immutable(ref data),
@@ -336,7 +334,9 @@ impl ImmutableDataManager {
         }
 
         // Choose the PmidNodes to store the data on, and add them in a new database entry.
-        let target_pmid_nodes = match self.choose_initial_pmid_nodes(routing_node, &data_name) {
+        let target_pmid_nodes = match self.choose_initial_pmid_nodes(routing_node,
+                                                                     full_pmid_nodes,
+                                                                     &data_name) {
             Ok(pmid_nodes) => pmid_nodes,
             Err(InternalError::UnableToAllocateNewPmidNode) => {
                 // Send failure if src is MaidManager.
@@ -538,7 +538,6 @@ impl ImmutableDataManager {
                                              .iter()
                                              .find(|&entry| entry.1 == *message_id) {
             if let Some(account) = self.accounts.get_mut(&entry.0.name()) {
-                let _ = self.known_pmid_nodes.entry(*pmid_node).or_insert(true);
                 if !account.pmid_nodes_mut().remove(&DataHolder::Pending(*pmid_node)) {
                     return Err(InternalError::InvalidResponse);
                 }
@@ -572,9 +571,6 @@ impl ImmutableDataManager {
                                                     .iter()
                                                     .find(|&entry| entry.1 == *message_id) {
             if let Some(account) = self.accounts.get_mut(&immutable_data.name()) {
-                let known_pmid_node = self.known_pmid_nodes.entry(*pmid_node).or_insert(false);
-                *known_pmid_node = false;
-
                 // Mark the holder as Failed
                 if !account.pmid_nodes_mut().remove(&DataHolder::Pending(*pmid_node)) {
                     return Err(InternalError::InvalidResponse);
@@ -651,7 +647,6 @@ impl ImmutableDataManager {
     }
 
     pub fn handle_node_lost(&mut self, routing_node: &RoutingNode, node_lost: XorName) {
-        let _ = self.known_pmid_nodes.remove(&node_lost);
         self.handle_churn(routing_node, MessageId::from_lost_node(node_lost));
     }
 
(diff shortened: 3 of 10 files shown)


maidsafe-highfive:
r? @maqi

Fraser999:
+@dirvine +@maqi +@brian-js

---

Review status: 0 of 10 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/safe_vault/403#-:-KEet-asRfb2qAqYDu48:1580881569)*

dirvine:
Reviewed 3 of 10 files at r1.
Review status: 3 of 10 files reviewed at latest revision, 1 unresolved discussion.

---

*[src/personas/immutable_data_manager.rs, line 340 \[r1\]](https://reviewable.io:443/reviews/maidsafe/safe_vault/403#-KEeuUj4DXjuCRS-QY_i:-KEeuUj5k7tTgNS5Hl3D:-264924159) ([raw file](https://github.com/maidsafe/safe_vault/blob/08c40996afb180f0d9689267a7412de8af5ebb92/src/personas/immutable_data_manager.rs#L340)):*
If a PmidNode cannot store for any reason we should mark them full I think. IS this done and then do we confirm we are not trying to store on a full node?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/safe_vault/403)*

Fraser999:
Review status: 3 of 10 files reviewed at latest revision, 1 unresolved discussion.

---

*[src/personas/immutable_data_manager.rs, line 340 \[r1\]](https://reviewable.io:443/reviews/maidsafe/safe_vault/403#-KEeuUj4DXjuCRS-QY_i:-KEexH1kNG36_G7ey-5J:-1974124043) ([raw file](https://github.com/maidsafe/safe_vault/blob/08c40996afb180f0d9689267a7412de8af5ebb92/src/personas/immutable_data_manager.rs#L340)):*
PmidNodes which fail to store are flagged as full [here](https://github.com/Fraser999/safe_vault/blob/full/src/personas/immutable_data_manager.rs#L575-L576).  Avoiding selecting full nodes is done in [`choose_initial_pmid_nodes`](https://github.com/Fraser999/safe_vault/blob/full/src/personas/immutable_data_manager.rs#L1143-L1176) (but untested).

Not sure what you mean by confirming we are not trying to store on a full node. Do you mean in tests, or are you saying that the `target_pmid_nodes` returned by `choose_initial_pmid_nodes` function should be checked here in the production code before sending the requests?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/safe_vault/403)*

maqi:
*[a discussion](https://reviewable.io:443/reviews/maidsafe/safe_vault/403#-KEf-nsUY0lf8Bp3MFWv:-KEf-nsUY0lf8Bp3MFWw:650057949) (no related file):*
I think it has one issue that: the knowledge of `known_pmid_nodes` will depends on the time the node joins the network, as such list is only get updated on put or get hits. This will cause in-consistency among the nodes and may prevent achieving a group consensus.
As mentioned in the slack channel, I will suggest to update this list whenever receiving a refreshed in account as well.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/safe_vault/403)*

maqi:
*[a discussion](https://reviewable.io:443/reviews/maidsafe/safe_vault/403#-KEf0p9kWzcATPf0CoNl:-KEf0p9kWzcATPf0CoNm:1784696300) (no related file):*
regarding the SMD concern, I think an easy way to do is: make `choose_initial_pmid_nodes` public, so vault can call it for a sd as well to tell whether the group around sd.name() is full.
this shall be accurate enough to reach a group consensus.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/safe_vault/403)*

maqi:
*[a discussion](https://reviewable.io:443/reviews/maidsafe/safe_vault/403#-KEf0p9kWzcATPf0CoNl:-KEf1zpaf6iqSpU_34lm:-98586694) (no related file):*
however, as SDM is using a separate storage space, so this might not be required.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/safe_vault/403)*

Fraser999:
Review status: 3 of 10 files reviewed at latest revision, 3 unresolved discussions.

---

*[a discussion](https://reviewable.io:443/reviews/maidsafe/safe_vault/403#-KEf-nsUY0lf8Bp3MFWv:-KEf2Jfn05h1kNM9uXMy:-1954268626) (no related file):*
Similar to refreshing the ongoing caches, I don't think this should be required.  Testing could well prove me wrong on this, and at that stage we can add the list of known full nodes to the refresh protocol.  This would need a slight break from the existing refresh messages which are all related to specific accounts - this would be a general message unrelated to any account.

---

*[a discussion](https://reviewable.io:443/reviews/maidsafe/safe_vault/403#-KEf0p9kWzcATPf0CoNl:-KEf339RoXKcxMEIhg_4:1185162802) (no related file):*
I was thinking that `vault` owning the `known_pmid_nodes` variable and passing this into the SDM and IDM functions that need to access or mutate it would be the way forward.  Alternatively, just have another similar variable in SDM and let it manage its own - keeping ImmutableData and SD totally separate.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/safe_vault/403)*

maqi:
*[a discussion](https://reviewable.io:443/reviews/maidsafe/safe_vault/403#-KEf-nsUY0lf8Bp3MFWv:-KEf4woworE55INYY3Xz:82614243) (no related file):*
I don't mean to refresh this list out during churn. what i mean is:
when receive a refreshed in account ( currently we just insert it)
we do an extra to: peek the data_holders in that account and update the local `known_pmid_nodes` correspondingly.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/safe_vault/403)*

Fraser999:
Review status: 3 of 10 files reviewed at latest revision, 3 unresolved discussions.

---

*[a discussion](https://reviewable.io:443/reviews/maidsafe/safe_vault/403#-KEf-nsUY0lf8Bp3MFWv:-KEf5TQSkw_5AGl87wN9:1320175497) (no related file):*
That's not the same though - it contains failed Get responders as well as failed Puts.  As previously discussed, I don't really want to complicate this unless tests show we need to.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/safe_vault/403)*

maqi:
*[a discussion](https://reviewable.io:443/reviews/maidsafe/safe_vault/403#-KEf0p9kWzcATPf0CoNl:-KEf5X4IguTEccRyBFxl:1440391171) (no related file):*
SDM decides to store an sd chunk by itself, so keeping a separate list is meaningless. :(

I am ok with making `known_pmid_nodes`  to be owned by `vault`, but you may have to pass it to IDM quite often.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/safe_vault/403)*

maqi:
*[a discussion](https://reviewable.io:443/reviews/maidsafe/safe_vault/403#-KEf-nsUY0lf8Bp3MFWv:-KEf6MzsEC7nUTpYdLvQ:-1538494864) (no related file):*
i think a failed get means that node is not available anymore, i.e. you can't put to it. right? especially we have already said ANY put_failure means that node is full.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/safe_vault/403)*

maqi:
*[a discussion](https://reviewable.io:443/reviews/maidsafe/safe_vault/403#-KEf-nsUY0lf8Bp3MFWv:-KEf6MzsEC7nUTpYdLvQ:-1538494864) (no related file):*
i think a failed get means that node is not available anymore, i.e. you can't put to it. right? especially we have already said ANY put_failure means that node is full.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/safe_vault/403)*

Fraser999:
Review status: 3 of 10 files reviewed at latest revision, 3 unresolved discussions.

---

*[a discussion](https://reviewable.io:443/reviews/maidsafe/safe_vault/403#-KEf0p9kWzcATPf0CoNl:-KEf6E8Va1-4t0e4Lmrw:-76084745) (no related file):*
Good point :)

Yeah - I'm thinking the list would be passed by ref to `IDM::handle_put` and `SDM::handle_put` and by mut ref to `IDM::handle_put_success` and `IDM::handle_put_failure`.  The vault could handle pruning entries during node lost churn events itself.  That would be about it I think.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/safe_vault/403)*

Fraser999:
Review status: 3 of 10 files reviewed at latest revision, 2 unresolved discussions.

---

*[a discussion](https://reviewable.io:443/reviews/maidsafe/safe_vault/403#-KEf-nsUY0lf8Bp3MFWv:-KEf7GdYTTvQmWQSO94p:-1645011916) (no related file):*
That may not be the only reason a node can't reply to a Get.  That chunk could be corrupted in its local chunkstore, or it could have lied about Putting, or it's get request could have timed out.  I don't want to add more complexity here and confuse the meaning of the existing structures without tests showing we need to.

If we're going to use the existing structures for this, I'd much rather keep things clean and have a separate variant like `FailedPut` and change the existing `Failed` to `FailedGet`.

We really don't know that this issue is actually a problem.  I'm not saying we should never do this - I just don't see the need to add it now.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/safe_vault/403)*

maqi:
*[a discussion](https://reviewable.io:443/reviews/maidsafe/safe_vault/403#-KEf0p9kWzcATPf0CoNl:-KEf9tf_wx4ofdb0KbqV:5174689) (no related file):*
ok

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/safe_vault/403)*

maqi:
*[src/personas/immutable_data_manager.rs, line 541 \[r1\]](https://reviewable.io:443/reviews/maidsafe/safe_vault/403#-KEfCWSZCBMKHFXfA9z0:-KEfCWSZCBMKHFXfA9z1:2067459894) ([raw file](https://github.com/maidsafe/safe_vault/blob/08c40996afb180f0d9689267a7412de8af5ebb92/src/personas/immutable_data_manager.rs#L541)):*
I think we are only interested in the FAILED pmid_nodes.
so actually the list can be just a list of nodes, and only insert into that list when a node is failed.
If this list may grows to large, this will improve the performance a bit.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/safe_vault/403)*

Fraser999:
Review status: 3 of 10 files reviewed at latest revision, 3 unresolved discussions.

---

*[src/personas/immutable_data_manager.rs, line 541 \[r1\]](https://reviewable.io:443/reviews/maidsafe/safe_vault/403#-KEfCWSZCBMKHFXfA9z0:-KEfEJBwka4wnV0hNRzC:1902010410) ([raw file](https://github.com/maidsafe/safe_vault/blob/08c40996afb180f0d9689267a7412de8af5ebb92/src/personas/immutable_data_manager.rs#L541)):*
Yeah, agreed.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/safe_vault/403)*

maqi:
*[a discussion](https://reviewable.io:443/reviews/maidsafe/safe_vault/403#-KEf-nsUY0lf8Bp3MFWv:-KEfAIrUI975vUhVm3mc:-2132201349) (no related file):*

The reasons you listed can all make the pmid_node to be dis-qualified as a data_holder:
get request could have timed out :                              a put_request can also get timed_out
chunk could be corrupted in its local chunkstore :    a corruptable disk shall not be trusted to put any more data on
lied about Putting :                                                        then shall not be trusted anymore, right?

So I don't think it will be necessary to separate FailedGet from FailedPut, i.e. it won't add more complexity.

Regarding your point of not to put it now and put it later once spot failure, my concern is:
it will be very hard to debug and pin down the reason of a failure (lost of data) during test is due to this reason, or something else.
especially when the unit tests won't be able to simulate it.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/safe_vault/403)*

Fraser999:
Review status: 1 of 10 files reviewed at latest revision, 2 unresolved discussions.

---

*[a discussion](https://reviewable.io:443/reviews/maidsafe/safe_vault/403#-KEf-nsUY0lf8Bp3MFWv:-KEfU4RgPbIcM-vHjdgh:-1429923062) (no related file):*
A corrupt chunk doesn't imply a corrupt disk.  All the other files could be fine.  A timed-out get request isn't the same thing as a timed-put request, and yes a node which lied about putting probably shouldn't be trusted.  We haven't even started to handle that case though.  If we decide to just disconnect from a node node which does any of those things, then this issue isn't a problem.  We haven't planned for that yet though - so there's no reason to group Get and Put failure types into a single case now.  We may well end up deciding that's OK, but we haven't planned for it here, so I don't think we should implement it.

As for your rationale of adding the extra code now - that's the same argument as refreshing the ongoing cache data and that was deemed not a strong enough argument to justify the extra work.  I believe this is a similar case.

Also, adding the extra code could well cause just as difficult-to-spot bugs as what you're describing here.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/safe_vault/403)*

maqi:
Reviewed 5 of 10 files at r1, 5 of 5 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/safe_vault/403)*

PR #422, opened 2016-04-17
WIP - restructuring 

dirvine:
Scope: Data/Get/Put flow 

Mainly changes to Data and account handling. `DataRequest` -> `DataIdentifier`, use this for containers (Account and ongoing_gets etc.) to ensure more type safety. Make accounts an individual container and handle individual Get's instead of optimising this process. Change DataHolder to struct with state for simpler handling. Change DataHolders to HashSet.

Unfinished yet, but well on way for discussion and completion (hopefully). This is a large change as small changes were not working out  easily at all.  

Some initial changes to structure to decouple code and move to split into 3 main areas 

1. Put/Get + requests/responses 

2. Replication

3. Churn handling

---
This change is [Reviewable](https://reviewable.io/reviews/maidsafe/safe_vault/422)


maidsafe-highfive:
r? @Fraser999

afck:
Review status: 0 of 13 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[src/personas/immutable_data_manager.rs, line 81 \[r7\]](https://reviewable.io:443/reviews/maidsafe/safe_vault/422#-KFcNN3YMYTKFMY4xgAS:-KFcNN3YMYTKFMY4xgAT:1899903981) ([raw file](https://github.com/maidsafe/safe_vault/blob/87dbb598851ef5db5ebae71beb7ce894ff8233a0/src/personas/immutable_data_manager.rs#L81)):*
To avoid ambiguity, maybe we should always use "identifier" instead of "name" for `DataIdentifiers`.

---

*[src/personas/immutable_data_manager.rs, line 234 \[r7\]](https://reviewable.io:443/reviews/maidsafe/safe_vault/422#-KFcQk0Q2Xjpyqm27Cu-:-KFcQk0Q2Xjpyqm27Cu0:677121513) ([raw file](https://github.com/maidsafe/safe_vault/blob/87dbb598851ef5db5ebae71beb7ce894ff8233a0/src/personas/immutable_data_manager.rs#L234)):*
Typo

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/safe_vault/422)*

afck:
Review status: 0 of 13 files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[src/vault.rs, line 242 \[r7\]](https://reviewable.io:443/reviews/maidsafe/safe_vault/422#-KFcVOzF8tt__jSl1T98:-KFcVOzGbn8QpcMjz57O:-880315373) ([raw file](https://github.com/maidsafe/safe_vault/blob/87dbb598851ef5db5ebae71beb7ce894ff8233a0/src/vault.rs#L242)):*
Instead of a match guard, maybe the `DataRequest` should be destructured as well, and only the identifier passed in to `handle_get`? (Also below.)

---

*[src/personas/immutable_data_manager.rs, line 120 \[r7\]](https://reviewable.io:443/reviews/maidsafe/safe_vault/422#-KFcUyS-pUeO13UpiS3c:-KFcUyS0QSgF89MBxO7a:-1296027184) ([raw file](https://github.com/maidsafe/safe_vault/blob/87dbb598851ef5db5ebae71beb7ce894ff8233a0/src/personas/immutable_data_manager.rs#L120)):*
Probably the `data_cache` should have identifiers as keys instead of names. (And the `data_holders`, too?)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/safe_vault/422)*

PR #436, opened 2016-04-22
WIP: fix/refresh: fixes refresh at DataManager

Fraser999:
This is an inital commit to implement the new paradigm of refreshing chunk names only on churn, and having the recipient make individual Get requests for the relevant data.

From brief task list:

in vault.rs:
- [x] handle in demuxer: MN->MN for Refresh, Get, GetSuccess and GetFailure

in handle_node_added:
- [x] iterate all data identities
- [x] gather those relevant to new node
- [x] for ImmData, just use name.  for SD use name and SHA512 of full chunk.
- [x] send single refresh msg to new node

in handle_node_lost:
- [ ] iterate all data identities
- [ ] gather those relevant to new node
- [ ] for ImmData, just use name.  for SD use name and SHA512 of full chunk.
- [ ] send single refresh msg to all members (more than GROUP_SIZE) to work around routing issue

in handle_refresh:
- [x] accumulate all peers' msgs.  This means create a list of any entry which is sent by at least QUORUM peers.  For any SDs which reach QUORUM but aren't all identical (e.g. different version) flag these so we can ask all group members for the data.  Don't add an entry if we already hold the data.
- [x] do Get MN->MN for all entries to a single member (chosen at random)
(TODO) - this is brute force, we should look at throttling the rate of requests being sent, and also being able to target a single peer rather than the group

in handle_get_success:
- [ ] For SD, do hash check to confirm data is valid.  If not, treat as GetFailure.
- [ ] store data
- [ ] remove entry

in handle_get_failure (or if entry timed out from timed buffer):
- [x] pick new peer at random (don't worry about picking previously-failed guy for now, but don't use the peer who sent us this msg)
- [x] do Get MN->MN for this chunk again.

---
This change is [Reviewable](https://reviewable.io/reviews/maidsafe/safe_vault/436)


maidsafe-highfive:
r? @maqi

PR #437, opened 2016-04-24
Fraser999 refresh

dirvine:
fix/handle_lost_node Code sketch for lost node handling 

This code will send to all nodes that are close to the data we know about. It does not (yet) try and filter this list in any way and will spam all nodes with list of data we know they should have. Many will have the data already and work very hard to get from chunk store and confirm this data hash. At the moment this is very wasteful and far to cpu intensive.

---
This change is [Reviewable](https://reviewable.io/reviews/maidsafe/safe_vault/437)


maidsafe-highfive:
r? @maqi

PR #438, opened 2016-04-25
WIP refactor/all 

dirvine:
refactor/all 
feature/dm handle_node_lost

Many updates as per commit messages. Refactor of much layout and also tidy up of routing_node being passed in all methods. 

NOT YET MERGEABLE - To Be Completed

From brief task list:

in vault.rs:
- [x] handle in demuxer: MN->MN for Refresh, Get, GetSuccess and GetFailure

in handle_node_added:
- [x] iterate all data identities
- [x] gather those relevant to new node
- [x] for ImmData, just use name.  for SD use name and SHA512 of full chunk.
- [x] send single refresh msg to new node

in handle_node_lost:
- [x] iterate all data identities
- [x] gather those relevant to new node
- [x] for ImmData, just use name.  for SD use name and SHA512 of full chunk.
- [x] send single refresh msg to all members (more than GROUP_SIZE) to work around routing issue

in handle_refresh:
- [x] accumulate all peers' msgs.  This means create a list of any entry which is sent by at least QUORUM peers.  For any SDs which reach QUORUM but aren't all identical (e.g. different version) flag these so we can ask all group members for the data.  Don't add an entry if we already hold the data.
- [x] do Get MN->MN for all entries to a single member (chosen at random)
(TODO) - this is brute force, we should look at throttling the rate of requests being sent, and also being able to target a single peer rather than the group

in handle_get_success:
- [ ] For SD, do hash check to confirm data is valid.  If not, treat as GetFailure.
- [ ] store data
- [ ] remove entry

in handle_get_failure (or if entry timed out from timed buffer):
- [x] pick new peer at random (don't worry about picking previously-failed guy for now, but don't use the peer who sent us this msg)
- [x] do Get MN->MN for this chunk again.

---
This change is [Reviewable](https://reviewable.io/reviews/maidsafe/safe_vault/438)


maidsafe-highfive:
r? @Fraser999

Fraser999:
Review status: 0 of 10 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[src/personas/data_manager.rs, line 350 \[r3\]](https://reviewable.io:443/reviews/maidsafe/safe_vault/438#-KGBbxiqrJ8l-gDVfP0M:-KGBbxir1IXriiAQLJLg:1911028183) ([raw file](https://github.com/maidsafe/safe_vault/blob/0a29643a5c386442be7822f81f5c1c5e89cf33e3/src/personas/data_manager.rs#L350)):*
We want the sum here, not the count.  We could have just 2 entries, but each with a value of `3`, meaning 6 peers have sent responses.

---

*[src/personas/data_manager.rs, line 368 \[r3\]](https://reviewable.io:443/reviews/maidsafe/safe_vault/438#-KGBf4iHcejtRheEaWs3:-KGBf4iHcejtRheEaWs4:-1645709045) ([raw file](https://github.com/maidsafe/safe_vault/blob/0a29643a5c386442be7822f81f5c1c5e89cf33e3/src/personas/data_manager.rs#L368)):*
If the `try` fails we return here rather than continuing with the other entries.  I think this should be reverted to `let _ = `.

---

*[src/personas/data_manager.rs, line 371 \[r3\]](https://reviewable.io:443/reviews/maidsafe/safe_vault/438#-KGBfGyfbIA4eFFL7fUY:-KGBfGyfbIA4eFFL7fUZ:-270281442) ([raw file](https://github.com/maidsafe/safe_vault/blob/0a29643a5c386442be7822f81f5c1c5e89cf33e3/src/personas/data_manager.rs#L371)):*
As per comment above.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/safe_vault/438)*

maqi:
*[src/personas/data_manager.rs, line 41 \[r3\]](https://reviewable.io:443/reviews/maidsafe/safe_vault/438#-KGC13Qjd31BLgm8wYMM:-KGC13Qjd31BLgm8wYMN:-1634493886) ([raw file](https://github.com/maidsafe/safe_vault/blob/0a29643a5c386442be7822f81f5c1c5e89cf33e3/src/personas/data_manager.rs#L41)):*
just put as a remind: the u8 (hit counter) better to be vector of senders so we can pick a better candidate for get requester.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/safe_vault/438)*

maqi:
*[src/personas/data_manager.rs, line 179 \[r3\]](https://reviewable.io:443/reviews/maidsafe/safe_vault/438#-KGC1ucVi8tRuFJGnnaV:-KGC1ucWTgoeS9h63ejO:-1574264638) ([raw file](https://github.com/maidsafe/safe_vault/blob/0a29643a5c386442be7822f81f5c1c5e89cf33e3/src/personas/data_manager.rs#L179)):*
we may still need to keep this group refresh, due to the strict mock_crust churn test.
anyway, the test will tell us whether it will be required or not.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/safe_vault/438)*

maqi:
*[src/personas/data_manager.rs, line 286 \[r3\]](https://reviewable.io:443/reviews/maidsafe/safe_vault/438#-KGC2KaaRend5gj8pRoW:-KGC2Kab5MpJzMhRZ0rN:-482343140) ([raw file](https://github.com/maidsafe/safe_vault/blob/0a29643a5c386442be7822f81f5c1c5e89cf33e3/src/personas/data_manager.rs#L286)):*
1, if we don't have the data, the first incoming get_success will be trusted

2, any incoming data having greater version will be trusted first.

3, although in refresh_accumulator, there is `majority` hash to be accumulated, the send_single_refresh might still be fired to the one holding different hash data, and will be trusted.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/safe_vault/438)*

maqi:
*[src/personas/data_manager.rs, line 317 \[r3\]](https://reviewable.io:443/reviews/maidsafe/safe_vault/438#-KGC34Ilfx6AYFYAK2FD:-KGC34ImrYeaBqTyqAVR:276779119) ([raw file](https://github.com/maidsafe/safe_vault/blob/0a29643a5c386442be7822f81f5c1c5e89cf33e3/src/personas/data_manager.rs#L317)):*
for structured_data, we need to ensure the hash is consistent with the local copy, otherwise, we are supposed to assume our local copy may needs an update.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/safe_vault/438)*

maqi:
*[src/personas/data_manager.rs, line 451 \[r3\]](https://reviewable.io:443/reviews/maidsafe/safe_vault/438#-KGC5dToulmeQR23X_Wc:-KGC5dTpXNratzW29l0I:1029399055) ([raw file](https://github.com/maidsafe/safe_vault/blob/0a29643a5c386442be7822f81f5c1c5e89cf33e3/src/personas/data_manager.rs#L451)):*
with cloned routing_table, a `new_group_member` for a data can be calculated out accurately.
Shall not use this `in_range_data_keys` function anymore, and the code block here needs to be refactored.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/safe_vault/438)*

PR #446, opened 2016-04-28
test/fill_network: re-enable integration test

Fraser999:
This re-enables the integration test which checks that Clients are sent appropriate failure messages
when the network's storage limit is reached.

---
This change is [Reviewable](https://reviewable.io/reviews/maidsafe/safe_vault/446)


maidsafe-highfive:
r? @maqi

PR #460, opened 2016-05-03
test/maid_manager: keeps putting data till client account is full

### LINKED ISSUE #453: Add a CI test which stores up to allowed Client limit
We should add a test similar to the existing CI one `fill_network` which, rather than filling the DMs, instead fills the Client's account.  This should assert that the Client is consistently notified with failure once its account is full.

maqi:
Closes #453

---
This change is [Reviewable](https://reviewable.io/reviews/maidsafe/safe_vault/460)


maidsafe-highfive:
r? @Fraser999

afck:
Review status: 0 of 2 files reviewed at latest revision, 1 unresolved discussion.

---

*[tests/maid_manager.rs, line 97 \[r2\]](https://reviewable.io:443/reviews/maidsafe/safe_vault/460#-KGvoJv4yZBK04iUiGIp:-KGvoJv5mYYi83JnUqy6:199148920) ([raw file](https://github.com/maidsafe/safe_vault/blob/5ddf7304bfcbe4a427e8670d7759e4163f2c629d/tests/maid_manager.rs#L97)):*
This needs to be kept in sync with maid_manager.rs, right?
I wonder whether it would make sense to put a `#[cfg(test)] pub fn get_default_account_size()` there? Or make it configurable in the first place ...?
Not sure, but I'd at least put a comment here for now.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/safe_vault/460)*

maqi:
*[tests/maid_manager.rs, line 97 \[r2\]](https://reviewable.io:443/reviews/maidsafe/safe_vault/460#-KGvoJv4yZBK04iUiGIp:-KGvtN0gu1wxUDWxW75k:1448359554) ([raw file](https://github.com/maidsafe/safe_vault/blob/5ddf7304bfcbe4a427e8670d7759e4163f2c629d/tests/maid_manager.rs#L97)):*
yes. don't want to mess up with the production code with too much #[cfg(test)].
ideally, a `setter` is preferred, so the test can be completed quicker.
will put a comment here for now.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/safe_vault/460)*

afck:
Reviewed 1 of 2 files at r2, 1 of 1 files at r3.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/safe_vault/460)*

PR #465, opened 2016-05-06
fix/test_client: Cope new Tick event

maqi:
The new routing now sends a Tick event to Vault

---
This change is [Reviewable](https://reviewable.io/reviews/maidsafe/safe_vault/465)


maidsafe-highfive:
r? @Fraser999

afck:
Yes, but it shouldn't actually send it for clients. This should fix the issue:
https://github.com/maidsafe/routing/pull/1012

PR #478, opened 2016-05-22
feat/all: upgrade to new routing with acks and 256-bit names

afck:
This change is [Reviewable](https://reviewable.io/reviews/maidsafe/safe_vault/478)


maidsafe-highfive:
r? @Fraser999

PR #513, opened 2016-06-28
removal of apply_config; cleanup temp directory once tests completed

maqi:
This change is [Reviewable](https://reviewable.io/reviews/maidsafe/safe_vault/513)

### DIFF
--- a/src/chunk_store/test.rs
+++ b/src/chunk_store/test.rs
@@ -26,9 +26,6 @@ macro_rules! assert_err {
 
 #[cfg(test)]
 mod test {
-    use std::env;
-    use std::path::PathBuf;
-
     use chunk_store::{ChunkStore, Error};
     use maidsafe_utilities::serialisation;
     use rand::{self, Rng};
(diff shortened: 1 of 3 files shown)


maidsafe-highfive:
r? @Fraser999

afck:
Reviewed 2 of 3 files at r1, 1 of 1 files at r2.

---

*[src/vault.rs, line 52 \[r2\]](https://reviewable.io:443/reviews/maidsafe/safe_vault/513#-KLQ8DI2m-yccRZI2Yx-:-KLQ8DI2m-yccRZI2Yx0:-1023224245) ([raw file](https://github.com/maidsafe/safe_vault/blob/4dc10b80005a1145de18d25276f52fd0a9f4166d/src/vault.rs#L52)):*
> ```Rust
>     /// Creates a network Vault instance.
>     pub fn new(first_vault: bool, use_cache: bool) -> Result<Self, InternalError> {
>         let config = try!(config_handler::read_config_file());
> ```

We need to distinguish here:
If the config file doesn't exist, we should use the default.
If it exists but can't be parsed, we should terminate and print a readable error message, like: `Config file could not be parsed: JsonParserError(SyntaxError("invalid syntax", 4, 24))`.
(Feel free to do this in a separate PR, if you prefer.)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/safe_vault/513)*

PR #520, opened 2016-06-30
fix/logfile: remove old logfile on startup

Fraser999:
This is a temporary workaround for a bug in log4rs whereby the contents of a previous logfile aren't
cleared before writing to the file in a new run.

---
This change is [Reviewable](https://reviewable.io/reviews/maidsafe/safe_vault/520)

### DIFF
--- a/src/bin/safe_vault.rs
+++ b/src/bin/safe_vault.rs
@@ -52,6 +52,7 @@ extern crate rustc_serialize;
 extern crate safe_vault;
 
 use std::ffi::OsString;
+use std::fs;
 use std::process;
 use docopt::Docopt;
 use safe_vault::Vault;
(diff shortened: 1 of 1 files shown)


maidsafe-highfive:
r? @maqi

src/bin/safe_vault.rs
```diff
@@ -77,9 +78,12 @@ struct Args {
...
+    // TODO - remove the following line once maidsafe_utilities is updated to use log4rs v4.
+    let _ = fs::remove_file("Node.log");
+
     let args: Args = Docopt::new(USAGE)
-                         .and_then(|docopt| docopt.decode())
-                         .unwrap_or_else(|error| error.exit());
+        .and_then(|docopt| docopt.decode())
+        .unwrap_or_else(|error| error.exit());
 
     let name = config_file_handler::exe_file_stem().unwrap_or_else(|_| OsString::new());
```
dirvine: I wonder should the config file handler be using `create(true)` to provide this functionality (know it's not part of this PR though). If we want read or create then that would be the way I think.

PR #523, opened 2016-06-30
fix/vault: terminate if chunk store cannot be reset

afck:
Also verify chunk store root allows creating files.

---
This change is [Reviewable](https://reviewable.io/reviews/maidsafe/safe_vault/523)


maidsafe-highfive:
r? @Fraser999

PR #545, opened 2016-09-05
fix/data_manager: fix Put and Post handling in data manager

afck:
This prevents a group from going out of sync in the rare case where
half of the nodes receive one Put/Post request first, and the other
half receive a different Put/Post request for the same data chunk first.
Previously, every node would handle and put in the chunk store the data
it got first. With the new approach, this situation would simply lead to
both requests being refused, leaving the data intact and in its previous
state:

When the nodes receive a request, they don't immediately write to the
chunk store, and instead cache the request in memory. Each node sends,
as a message from the group to itself, the _first_ version of the
data that it received. Only if that message accumulates, i.e. if a
quorum of nodes agrees on which request to choose, they write it to
disk.

---
This change is [Reviewable](https://reviewable.io/reviews/maidsafe/safe_vault/545)


maidsafe-highfive:
r? @Fraser999

Fraser999:
Review status: 0 of 20 files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[src/personas/data_manager.rs, line 335 \[r1\]](https://reviewable.io:443/reviews/maidsafe/safe_vault/545#-KQuQHuUeelgeV6KnriK:-KQuQHuUeelgeV6KnriL:b-nsrwkm) ([raw file](https://github.com/maidsafe/safe_vault/blob/78fc4ebfa75d315c4ddd7041591b524546376d2a/src/personas/data_manager.rs#L335)):*
> ```Rust
>         self.pending_writes
>             .remove(data_id)
>             .and_then(|posts| posts.into_iter().find(|post| post.hash == hash))
> ```

Would `post` be better named as `pending_write`?

---

*[src/personas/data_manager.rs, line 646 \[r1\]](https://reviewable.io:443/reviews/maidsafe/safe_vault/545#-KQuX-IwWqdlrjdHnTy8:-KQuX-Ixc64eSld339un:biqtqh5) ([raw file](https://github.com/maidsafe/safe_vault/blob/78fc4ebfa75d315c4ddd7041591b524546376d2a/src/personas/data_manager.rs#L646)):*
> ```Rust
>                     self.cache.add_records((data_id, version), group.into_iter().collect());
>                 }
>                 return Ok(());
> ```

Shouldn't we be calling `self.send_gets_for_needed_data()` before returning here, since we should be a holder of this chunk?

---

*[tests/data_manager.rs, line 111 \[r1\]](https://reviewable.io:443/reviews/maidsafe/safe_vault/545#-KQuXZPALVH09ddPOk4i:-KQuXZPALVH09ddPOk4j:b-qp8bxb) ([raw file](https://github.com/maidsafe/safe_vault/blob/78fc4ebfa75d315c4ddd7041591b524546376d2a/tests/data_manager.rs#L111)):*
> ```Rust
> #[test]
> fn structured_data_parallel_posts() {
>     // let _ = ::maidsafe_utilities::log::init(false); // TODO: remove!
> ```

This looks like it could be removed now.

---

*[tests/data_manager.rs, line 119 \[r1\]](https://reviewable.io:443/reviews/maidsafe/safe_vault/545#-KQuXepfL30GfMqXMEJT:-KQuXepgkRPNxcMNz9xr:b8beews) ([raw file](https://github.com/maidsafe/safe_vault/blob/78fc4ebfa75d315c4ddd7041591b524546376d2a/tests/data_manager.rs#L119)):*
> ```Rust
>     let mut clients: Vec<_> = (0..3)
>         .map(|_| {
>             let endpoint = rng.choose(&nodes).unwrap().endpoint();
> ```

Shouldn't we prefer the `unwrap!` macro?  There are also some usages of the deprecated `unwrap_result!` macro.  If these are all because of copy-paste and they exist elsewhere in the file, maybe best to wait and clean them all up in a single PR though?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/safe_vault/545)*

afck:
Review status: 0 of 20 files reviewed at latest revision, 4 unresolved discussions.

---

*[src/personas/data_manager.rs, line 335 \[r1\]](https://reviewable.io:443/reviews/maidsafe/safe_vault/545#-KQuQHuUeelgeV6KnriK:-KQuiSjk4sSwdVVZ6I-7:b-o8ogoa) ([raw file](https://github.com/maidsafe/safe_vault/blob/78fc4ebfa75d315c4ddd7041591b524546376d2a/src/personas/data_manager.rs#L335)):*
<details><summary><i>Previously, Fraser999 (Fraser Hutchison) wrote…</i></summary>
> Would `post` be better named as `pending_write`?
</details>
Done - well, almost. But if you strongly prefer that, I'll change it to `pending_write`. (And the other one to `writes_queue`?)

---

*[src/personas/data_manager.rs, line 646 \[r1\]](https://reviewable.io:443/reviews/maidsafe/safe_vault/545#-KQuX-IwWqdlrjdHnTy8:-KQuirJAaFGLF_Sr4RXf:b-896fix) ([raw file](https://github.com/maidsafe/safe_vault/blob/78fc4ebfa75d315c4ddd7041591b524546376d2a/src/personas/data_manager.rs#L646)):*
<details><summary><i>Previously, Fraser999 (Fraser Hutchison) wrote…</i></summary>
> Shouldn't we be calling `self.send_gets_for_needed_data()` before returning here, since we should be a holder of this chunk?
</details>
Done.

---

*[tests/data_manager.rs, line 111 \[r1\]](https://reviewable.io:443/reviews/maidsafe/safe_vault/545#-KQuXZPALVH09ddPOk4i:-KQuiy8wjA7O909LhiSE:b-896fix) ([raw file](https://github.com/maidsafe/safe_vault/blob/78fc4ebfa75d315c4ddd7041591b524546376d2a/tests/data_manager.rs#L111)):*
<details><summary><i>Previously, Fraser999 (Fraser Hutchison) wrote…</i></summary>
> This looks like it could be removed now.
</details>
Done.

---

*[tests/data_manager.rs, line 119 \[r1\]](https://reviewable.io:443/reviews/maidsafe/safe_vault/545#-KQuXepfL30GfMqXMEJT:-KQuizO455KrxF91KYrF:bmik2zo) ([raw file](https://github.com/maidsafe/safe_vault/blob/78fc4ebfa75d315c4ddd7041591b524546376d2a/tests/data_manager.rs#L119)):*
<details><summary><i>Previously, Fraser999 (Fraser Hutchison) wrote…</i></summary>
> Shouldn't we prefer the `unwrap!` macro?  There are also some usages of the deprecated `unwrap_result!` macro.  If these are all because of copy-paste and they exist elsewhere in the file, maybe best to wait and clean them all up in a single PR though?
</details>
Done - almost.
We're still using `unwrap_option!` and `unwrap_result!` in safe_vault, and maybe moving to the unwrap crate should be a separate PR?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/safe_vault/545)*

Fraser999:
Review status: 0 of 20 files reviewed at latest revision, all discussions resolved.

---

*[src/personas/data_manager.rs, line 335 \[r1\]](https://reviewable.io:443/reviews/maidsafe/safe_vault/545#-KQuQHuUeelgeV6KnriK:-KQulMd56fSWn7ru_Llt:bp0624f) ([raw file](https://github.com/maidsafe/safe_vault/blob/78fc4ebfa75d315c4ddd7041591b524546376d2a/src/personas/data_manager.rs#L335)):*
<details><summary><i>Previously, afck (Andreas Fackler) wrote…</i></summary>
> Done - well, almost. But if you strongly prefer that, I'll change it to `pending_write`. (And the other one to `writes_queue`?)
</details>
No - what you have now is fine with me.  It was just the use of `post` which was niggling :)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/safe_vault/545)*

Fraser999:
Reviewed 17 of 20 files at r1, 3 of 3 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/safe_vault/545)*

Fraser999:
Reviewed 1 of 1 files at r4.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/safe_vault/545)*

vtnerd:
Isn't this just another data race? Each vault node is computing a quorum based on the message receiving order from the other vaults. The mock Network, Client, and Node all appear single threaded. Did I miss a thread somewhere? If the entire test is single threaded, this is not simulating the environment in a meaningful way.

afck:
It's single-threaded but it randomises the order in which messages are processed. This commit addresses the case where 50% of nodes handle one message first, and 50% handle the other message first. The test _did_ reproduce that problem (using https://github.com/maidsafe/routing/pull/1134).

vtnerd:
But what about the data race question? This is still fundamentally a race, is it not? There is no reason to suggest that every node receives the same message a quorum number of times first.

What happens when the two nodes receive that message roughly the same time? They are both broadcasting this new message to the entire group. Isn't this the same case as before, but with the addition of accumulation? One of them is going to win at each node, but it will not necessarily be the same one at each node.

Since there are no threads, when the first PUT message is sent, all of the dummy nodes process this new message type when handling this new broadcast immediately. They all put it into the accumulator in the same way. This happens for each PUT message. When the first quorum sized messages have been received, every node processes the accumulation the same way by using the message that reached quorum. Even with a fixed accmulator, I do not see how this is simulating the real environment.

afck:
Yes, the timeout is irrelevant in the test, and it's long enough that the messages should either have accumulated in that timeframe or not accumulate at all, so it should be irrelevant in production, too.

Now each node handles one of the two requests first and "votes" for that one: It will _only_ send the refresh message for the request it received first; the second one is just silently put into the cache. That's why in total, only `GROUP_SIZE` refresh messages are sent at all. Either there's a quorum for request A, or for request B or for neither. It cannot happen that both refresh messages accumulate.

If neither accumulates, no node will write anything to the chunk store.

If the refresh triggered by request A reaches quorum, _all_ nodes in the group (even those that received request B first) will accumulate only that one refresh message, and put data chunk A into the chunk store.

vtnerd:
Its the dynamic group membership that's really confusing here. It seems like there be a case where it would trigger a vote count higher than the expected group membership. I would have to think about it, but its really hairy.

vtnerd:
I'm guessing that this would be considered out of scope right now. Just remember that placing machine under load can do funny things at times, and a machine can be placed under load by giving it more messages. Its this constant strange manipulation that has always bothered me about this design.

vtnerd:
I don't know why I was hesitant before, the group count isn't locked during this stage, so this is certainly possible. Its just a matter of getting to happen on command.

afck:
It's "locked" in the sense that the nodes that already sent a refresh message have an entry in their cache and therefore won't send another one, even if they add another request to their cache. I don't see how two contradictory refresh messages could accumulate in the scenario that this commit addresses.

Yes, we need to be careful: No matter how the quorum or group size definitions change in the future, it must always be ensured that there cannot be two contradictory quora if every group member votes only once. (Not only for this particular commit, but for group consensus to work in general.)

Currently, every group always has exactly 8 members, and the quorum size is 5. With RFC 37, the group size will be dynamic and the quorum will, for now, be defined as a strict, absolute majority. These two definitions, at least, should be fine there.

vtnerd:
There is no synchronization on the decision to include a machine in the group. I would try to put a machine under load (the cousin of DDoS), with the goal of getting the heartbeat processing time to increase until it reaches near the cutoff. This affects the group size calculation by every member of the group, including the one under load. The machine under load could determine that the  group size is halved, and process only messages that voted in the "minority". This would still pass the quorum check, because the number of votes needed to reach quorum is lower on that machine. The important thing to remember is that under load, not everyone will agree on the status of that machine without synchronization. And the inverse case is true as well, the machine under load could think a different set of machines are "down".

I admit that this technique is a little more difficult. But I have never seen anyone attempt to do quorum this way, because it becomes really hard to reason about. From a  security perspective in particular, this is scary. I hope someone tries to understand the total behavior of the system, and then tests the hell out of it.

vtnerd:
I forgot to mention the case where a single malicious node in the group sends a different vote to each half of the group. What happens then? Should a single group be able to exert so much influence? Just consider getting  the team to read some papers related to byzantine, consensus, etc., to make sure everyone thinks about comparable systems. I promise I am done on this issue, unless otherwise probed by the team directly.

afck:
Yes, during churn it's still possible for a group to go out of sync at the moment. This change is only about the case of simultaneous Put or Post requests, as described in the commit message.

In the long run the plan is that each node signs each event and keeps a recorded history of the signatures in the group, so that these histories can be synchronised even after the fact. The goal is always only eventual consistency.

Scenarios like the ones you are pointing out need to be considered and tested, of course!
But yes, that's beyond the scope of this PR. :-)

PR #553, opened 2016-09-28
change maid_manager_account_decrease_with_churn to use duplicated put

maqi:
This change is [Reviewable](https://reviewable.io/reviews/maidsafe/safe_vault/553)

### DIFF
--- a/tests/maid_manager.rs
+++ b/tests/maid_manager.rs
@@ -24,7 +24,6 @@ use rand::distributions::{IndependentSample, Range};
 use routing::{Data, GROUP_SIZE, ImmutableData, XorName};
 use routing::client_errors::{GetError, MutationError};
 use routing::mock_crust::{self, Network};
-use safe_vault::Config;
 use safe_vault::mock_crust_detail::{self, poll, test_node};
 use safe_vault::mock_crust_detail::test_client::TestClient;
 use safe_vault::test_utils;
@@ -200,14 +199,9 @@ fn maid_manager_account_adding_with_churn() {
 
 #[test]
 fn maid_manager_account_decrease_with_churn() {
-    let config = Config {
-        wallet_address: None,
-        max_capacity: Some(3000),
-        chunk_store_root: None,
-    };
     let network = Network::new(None);
     let node_count = 15;
-    let mut nodes = test_node::create_nodes(&network, node_count, Some(config.clone()), false);
+    let mut nodes = test_node::create_nodes(&network, node_count, None, false);
     let client_config = mock_crust::Config::with_contacts(&[nodes[0].endpoint()]);
     let mut client = TestClient::new(&network, Some(client_config));
 
@@ -218,13 +212,15 @@ fn maid_manager_account_decrease_with_churn() {
 
     let full_id = client.full_id().clone();
     let mut event_count = 0;
+    let chunks_per_iter = 4;
+    let mut data_list = Vec::new();
 
     for i in 0..10 {
         trace!("Churning on {} nodes, iteration {}", nodes.len(), i);
         if nodes.len() <= GROUP_SIZE + 2 || rng.gen() {
             let index = Range::new(1, nodes.len()).ind_sample(&mut rng);
             trace!("Adding node with bootstrap node {}.", index);
-            test_node::add_node_with_config(&network, &mut nodes, config.clone(), index, false);
+            test_node::add_node(&network, &mut nodes, index, false);
         } else {
             let number = Range::new(1, 4).ind_sample(&mut rng);
             trace!("Removing {} node(s).", number);
@@ -233,10 +229,20 @@ fn maid_manager_account_decrease_with_churn() {
                 test_node::drop_node(&mut nodes, node_index);
             }
         }
-        for data in (0..4).map(|_| {
-            Data::Structured(test_utils::random_structured_data(100000, &full_id, &mut rng))
-        }) {
-            client.put(data.clone());
+        if i % 2 == 0 {
+            data_list.clear();
+            for data in (0..chunks_per_iter).map(|_| {
+                Data::Structured(test_utils::random_structured_data(100000, &full_id, &mut rng))
+            }) {
+                client.put(data.clone());
+                data_list.push(data);
+            }
+        } else {
+            for data in &data_list {
+                // Expect to be failed in DM. MM acount is increased first but be decreased back
+                // due to the put failure response from DM.
+                client.put(data.clone());
+            }
         }
         event_count += poll::poll_and_resend_unacknowledged(&mut nodes, &mut client);
 
@@ -250,9 +256,8 @@ fn maid_manager_account_decrease_with_churn() {
         let node_count_stats: Vec<(XorName, Option<u64>)> = sorted_maid_managers.into_iter()
             .map(|x| (x.name(), x.get_maid_manager_put_count(client.name())))
             .collect();
-        let expect_count = node_count_stats[0].1;
         for &(_, count) in &node_count_stats {
-            assert_eq!(count, expect_count);
+            assert_eq!(count, Some(chunks_per_iter * (i / 2 + 1) + 1));
         }
     }
 }


maidsafe-highfive:
r? @Fraser999

tests/maid_manager.rs
```diff
@@ -233,10 +228,17 @@ fn maid_manager_account_decrease_with_churn() {
...
-            client.put(data.clone());
+        if i % 2 == 0 {
+            for data in (0..4).map(|_| {
+                Data::Structured(test_utils::random_structured_data(100000, &full_id, &mut rng))
+            }) {
+                client.put(data.clone());
+                data_list.push(data.clone());
+            }
+        } else {
+            for data in &data_list {
```
Viv-Rajkumar: Any specific reason we're trying to repeat failure puts cumulatively? we don't clear `data_list`, so in every odd iteration, we're going to try to PUT all items in `data_list` again, so iteration 9 would be trying 20 PUTs?

We could maybe just try with the last X num of items from `data_list` as was added in the previous iteration?

Viv-Rajkumar:
> So if all puts attempted in this test had succeeded(as in the code this test is testing was functioning wrong)

should be:

So if all MM ignored the put failures from DM to actually decrease store count and returned success/failure to client(as in the code this test is testing was functioning wrong)

maqi:
Review status: 0 of 1 files reviewed at latest revision, 1 unresolved discussion.

---

*[tests/maid_manager.rs, line 239 at r1](https://reviewable.io:443/reviews/maidsafe/safe_vault/553#-KSlt135Zlz9lXhBhNGD-r1-239:-KSowqe8uoSDshU--vEg:bcmvn9o) ([raw file](https://github.com/maidsafe/safe_vault/blob/b52884606f49aec9248f6d9494f772e3f9a4a6f7/tests/maid_manager.rs#L239)):*
<details><summary><i>Previously, Viv-Rajkumar (Viv Rajkumar) wrote…</i></summary>
> Any specific reason we're trying to repeat failure puts cumulatively? we don't clear `data_list`, so in every odd iteration, we're going to try to PUT all items in `data_list` again, so iteration 9 would be trying 20 PUTs?
> 
> 
> 
> We could maybe just try with the last X num of items from `data_list` as was added in the previous iteration?
> 
>  
</details>
in each iteration, different nodes will added/removed. so the cumulatively repeat of failure puts is just to ensure those old data has been properly relocated, so the account can be updated properly/consistently among all nodes.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/safe_vault/553)*

maqi:
what is this about?

---

Review status: 0 of 1 files reviewed at latest revision, 1 unresolved discussion.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/safe_vault/553#-:-KSozEhME5E-pU2Biwb2:b-9wyecr)*

Fraser999:
Review status: 0 of 1 files reviewed at latest revision, 2 unresolved discussions.

---

*[tests/maid_manager.rs, line 239 at r1](https://reviewable.io:443/reviews/maidsafe/safe_vault/553#-KSlt135Zlz9lXhBhNGD-r1-239:-KSpsXuWwi2GET-1Q0_p:b-6zwe88) ([raw file](https://github.com/maidsafe/safe_vault/blob/b52884606f49aec9248f6d9494f772e3f9a4a6f7/tests/maid_manager.rs#L239)):*
<details><summary><i>Previously, maqi wrote…</i></summary>
> in each iteration, different nodes will added/removed. so the cumulatively repeat of failure puts is just to ensure those old data has been properly relocated, so the account can be updated properly/consistently among all nodes.
</details>
That's probably something that's out of scope for this test though.  We shouldn't have to Put the old data again to work round a problem which should be getting tested in a different test, and hopefully doesn't exist as a problem anyway?

---

*[tests/maid_manager.rs, line 216 at r2](https://reviewable.io:443/reviews/maidsafe/safe_vault/553#-KSpt9hqarG0Y2Y4tytf:-KSpt9hqarG0Y2Y4tytg:b6ie3vg) ([raw file](https://github.com/maidsafe/safe_vault/blob/345535886e42ba51401c6181867e27ad127428f3/tests/maid_manager.rs#L216)):*
> ```Rust
>     let mut event_count = 0;
>     let mut data_list = Vec::new();
>     let mut expect_count = 1;
> ```

We don't need this variable at all, do we?  It always ends up as `41` when we finally assert on it.  I think it would be good to use named variables for the two loop counts and use those in the assertion (e.g. `assert_eq!(count, Some(iterations * chunks_per_iter + 1));`)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/safe_vault/553)*

Fraser999:
Review status: 0 of 1 files reviewed at latest revision, 3 unresolved discussions.

---

*[tests/maid_manager.rs, line 215 at r2](https://reviewable.io:443/reviews/maidsafe/safe_vault/553#-KSpvA_4pV-B7eIFDwfj:-KSpvA_5EYgM5C1ITRrp:b-fowgtk) ([raw file](https://github.com/maidsafe/safe_vault/blob/345535886e42ba51401c6181867e27ad127428f3/tests/maid_manager.rs#L215)):*
> ```Rust
>     let full_id = client.full_id().clone();
>     let mut event_count = 0;
>     let mut data_list = Vec::new();
> ```

I guess we could reduce the scope of this variable by just moving this line inside the loop, then we can get rid of the line calling `clear()` each time.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/safe_vault/553)*

maqi:
Review status: 0 of 1 files reviewed at latest revision, 3 unresolved discussions.

---

*[tests/maid_manager.rs, line 215 at r2](https://reviewable.io:443/reviews/maidsafe/safe_vault/553#-KSpvA_4pV-B7eIFDwfj:-KSqrU_L0TWbkltabpub:b-wbcyw) ([raw file](https://github.com/maidsafe/safe_vault/blob/345535886e42ba51401c6181867e27ad127428f3/tests/maid_manager.rs#L215)):*
<details><summary><i>Previously, Fraser999 (Fraser Hutchison) wrote…</i></summary>
> I guess we could reduce the scope of this variable by just moving this line inside the loop, then we can get rid of the line calling `clear()` each time.
</details>
the scope cannot be reduced, because data_list is being populated in iteration 1, then being read in iteration 2.
so basically: in one iteration we put new data, then in next iteration we put the same data again (MM account will be increased then decreased, as received put_failure from DM)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/safe_vault/553)*

Fraser999:
Review status: 0 of 1 files reviewed at latest revision, 3 unresolved discussions.

---

*[tests/maid_manager.rs, line 215 at r2](https://reviewable.io:443/reviews/maidsafe/safe_vault/553#-KSpvA_4pV-B7eIFDwfj:-KSuXCb3zEdhn7LVS6VR:b-d7o9am) ([raw file](https://github.com/maidsafe/safe_vault/blob/345535886e42ba51401c6181867e27ad127428f3/tests/maid_manager.rs#L215)):*
<details><summary><i>Previously, maqi wrote…</i></summary>
> the scope cannot be reduced, because data_list is being populated in iteration 1, then being read in iteration 2.
> so basically: in one iteration we put new data, then in next iteration we put the same data again (MM account will be increased then decreased, as received put_failure from DM)
</details>
Yeah - you're right - sorry.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/safe_vault/553)*

afck:
Reviewed 1 of 1 files at r3.
Review status: all files reviewed at latest revision, 5 unresolved discussions.

---

*[tests/maid_manager.rs, line 238 at r3](https://reviewable.io:443/reviews/maidsafe/safe_vault/553#-KSxqnXB6YhaP4hYCSu8:-KSxqnXB6YhaP4hYCSu9:b-x093r5) ([raw file](https://github.com/maidsafe/safe_vault/blob/b16aafa50e28e41887bd25c09538230035d5875a/tests/maid_manager.rs#L238)):*
> ```Rust
>             }) {
>                 client.put(data.clone());
>                 data_list.push(data.clone());
> ```

That second `clone()` is probably not needed anymore. The `data` can be moved into the `data_list`.

---

*[tests/maid_manager.rs, line 242 at r3](https://reviewable.io:443/reviews/maidsafe/safe_vault/553#-KSxrRM3k5azsu3DQ4Yc:-KSxrRM3k5azsu3DQ4Yd:bvb9ds6) ([raw file](https://github.com/maidsafe/safe_vault/blob/b16aafa50e28e41887bd25c09538230035d5875a/tests/maid_manager.rs#L242)):*
> ```Rust
>         } else {
>             for data in &data_list {
>                 client.put(data.clone());
> ```

Please add a comment that this is expected to fail and that therefore it is not expected to increase the `Put` count in the client manager.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/safe_vault/553)*

afck:
Reviewed 1 of 1 files at r4.
Review status: all files reviewed at latest revision, 3 unresolved discussions, some commit checks broke.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/safe_vault/553)*

Fraser999:
Reviewed 1 of 1 files at r4.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/maidsafe/safe_vault/553)*

maqi:
The maid_manager::maid_manager_account_decrease_with_churn test fails occasionally, for example with RNG seed: [1907805060, 4035237522, 603661178, 3919140052].

This is due to: 
the new joined close node received account refresh AFTER the put_request, and other existing node doesn't send out any refresh of that account due to missing put_success from DM.
the missing put_success from DM is due to : among 8 DM nodes, 4 has full space, so responded with put_failure, 4 has free space so put in pending_list and wait for NAE group_refresh. However, none of them will get enough to accumulate.
Regarding the mis-match among nodes of full status, it may because: only in DM::handle_put, we check only use half of the chunk_store's max_space, in other places (pending writes and get_response for refreshed in data), we don't carry out such check. This opens couple of gaps:
1, when receiving multiple put_requests, they may all pass the 50% check, but actually taking too much
   If some nodes completes pending writes procedure quicker, it may reject some put_requests, meanwhile other still accept it.
2, Refreshed in data has the same issue, and it totally by-passed the 50% check.

This PR only change the testing code to use `duplicate put` to trigger `put_failure response` from DM to MM, which will decrease MM account. However, this doesn't resolve the issue described above.


Currently pending this PR to highlight the issue.

PR #555, opened 2016-10-06
feat/data_manager: support deletion

maqi:


maidsafe-highfive:
r? @Fraser999

src/personas/data_manager.rs
```diff
@@ -530,25 +544,22 @@ impl DataManager {
...
-        //    }
-        // }
-        // trace!("DM sending delete_failure for {:?}", new_data.identifier());
-        trace!("DM currently doesn't support Delete. Sending delete_failure for {:?}",
-               new_data.identifier());
+
+        if let Ok(Data::Structured(mut data)) = self.chunk_store.get(&data_id) {
+           if data.delete_self_against_successor(&new_data).is_ok() {
+                // chunk_store::put() deletes the old data automatically.
+                try!(self.chunk_store.put(&data_id, &Data::Structured(data)));
```
afck: This should send a `GroupRefresh` message instead of writing to the chunk store immediately, like the other data mutations do.

Also, it should still `self.count_removed_data(&data_id)`, as the `Put` will still count the re-added data as new, I think? (If it doesn't, that's okay, too, but they need to match and we should document what exactly the counter counts.)

maqi: which means we will need a pending_deletes and a new PendingWriteType::Delete   (probably rename PendingWriteType to PendingMutationType ?)

I don't think we shall count_removed_data as the data is still a copy managed by us (though it is almost null)
maybe we shall have another statistic say count_deleted_data, but that's another story

afck: Yes, we'll need a `Delete` variant. I'm happy with renaming it to `PendingMutationType`. (That's the term we also use in the errors, right?)

Good point: Let's still count data that's deleted but managed by us. In that case, we need to count the data on `Put` only if it didn't exist before.

afck:
This was included in #559.

PR #556, opened 2016-10-06
Conform to utilities v0.10.0

ustulation:
Re-run this after publishing routing 0.27.0 to crates.


maidsafe-highfive:
r? @Fraser999

ustulation:
Routing is not going to be published to crates just now. Needs to be re-raised when routing gets a version publish.

PR #557, opened 2016-10-13
chore/all: fix Clippy lints and update dependencies

afck:


maidsafe-highfive:
r? @maqi

PR #561, opened 2016-10-21
fix/maid_manager: handle different clients creating the same account

afck:


afck:
@maqi

maidsafe-highfive:
r? @maqi

src/personas/maid_manager.rs
```diff
@@ -162,8 +165,25 @@ impl MaidManager {
...
+                        let refresh = Refresh::Delete(client_name);
+                        if let Ok(serialised_refresh) = serialisation::serialise(&refresh) {
+                            trace!("MM sending delete refresh for account {}", src.name());
+                            let _ = self.routing_node
+                                .send_refresh_request(dst.clone(),
+                                                      dst.clone(),
+                                                      serialised_refresh,
+                                                      msg_id);
+                        }
+                        MutationError::AccountExists
```
maqi: probably it has been discussed, but to me `AccountExists` means the `name` of the client exists.
but here it is actually the `AccountPacket` exists.
the `account packet` maybe to be regenerated a new one easily, but having a `new name` of client means the client has to disconnect then reconnect.

afck: Starting a client with the same private key as someone else is cryptographically hard, so this needs no special handling.
The case where the same client is trying to create an account packet again is already handled (and `AccountExists` is returned).
This caters in addition for the case where a different client is trying to create account data with the same data name.

PR #562, opened 2016-10-24
MAID-1616 chore/persona:remove plain data

maqi:


maidsafe-highfive:
r? @Fraser999

PR #564, opened 2016-10-24
MAID-1617 Fix build with routing master.

afck:


maidsafe-highfive:
r? @maqi

src/vault.rs
```diff
@@ -193,7 +180,12 @@ impl Vault {
                 ret = Some(true);
                 Ok(())
             }
-            Event::Connected | Event::Tick => Ok(()),
+            Event::NodeAdded(_, _) |
+            Event::NodeLost(_, _) |
+            Event::GroupSplit(_) |
+            Event::GroupMerge(_) |
+            Event::Connected |
+            Event::Tick => Ok(()),
```
afck: You're right, I changed that.

PR #566, opened 2016-10-26
MAID-1644 chore/vault: vault events dispatch to support disjoint group

maqi:


maidsafe-highfive:
r? @Fraser999

PR #567, opened 2016-10-26
MAID-1646 chore/data_manager: handle_node_add for disjoint group

maqi:


maidsafe-highfive:
r? @Fraser999

PR #569, opened 2016-10-27
MAID-1649 chore/data_manager: handle_group_split for disjoint group

maqi:


maidsafe-highfive:
r? @Fraser999

src/personas/data_manager.rs
```diff
@@ -162,17 +162,17 @@ impl Cache {
...
-    //     }
-    //     pruned_unneeded_chunks.len() as u64
-    // }
+    fn prune_unneeded_chunks_for_split(&mut self, prefix: &Prefix<XorName>) -> u64 {
+        let mut n_pruned = 0;
+        self.unneeded_chunks.retain(|data_id|
+            if prefix.matches(data_id.name()) {
+                false
+            } else {
+                n_pruned+=1;
```
afck: Please apply rustfmt.

src/personas/data_manager.rs
```diff
@@ -162,17 +162,17 @@ impl Cache {
...
-    //         .iter()
-    //         .filter(|data_id| routing_table.is_close(data_id.name(), GROUP_SIZE))
-    //         .cloned()
-    //         .collect();
-    //     if !pruned_unneeded_chunks.is_empty() {
-    //         self.unneeded_chunks.retain(|data_id| !pruned_unneeded_chunks.contains(data_id));
-    //     }
-    //     pruned_unneeded_chunks.len() as u64
-    // }
+    fn prune_unneeded_chunks_for_split(&mut self, prefix: &Prefix<XorName>) -> u64 {
```
afck: If I remember correctly, the `unneeded_chunks` were those that _didn't_ belong to our group anymore? We don't delete such immutable data if we have space, and just mark it as unneeded, so that when it moves back into our group again, we don't need to retrieve it again.
That means, `unneeded_chunks` needs to be pruned on merge, not on split, I think?

src/personas/data_manager.rs
```diff
@@ -921,8 +926,44 @@ impl DataManager {
         unimplemented!()
     }
 
-    pub fn handle_group_split(&mut self, _prefix: &Prefix<XorName>) {
-        unimplemented!()
+    pub fn handle_group_split(&mut self, prefix: &Prefix<XorName>) {
+        let pruned_unneeded_chunks = self.cache.prune_unneeded_chunks_for_split(prefix);
+        if pruned_unneeded_chunks != 0 {
+            self.immutable_data_count += pruned_unneeded_chunks;
+            if self.logging_time.elapsed().as_secs() > STATUS_LOG_INTERVAL {
```
Fraser999: I think these four lines should be moved to a separate function - they're duplicated five times in this file.

PR #570, opened 2016-10-27
MAID-1650 chore/data_manager: handle_group_merge for disjoint group

maqi:


maidsafe-highfive:
r? @Fraser999

src/personas/data_manager.rs
```diff
@@ -915,8 +909,20 @@ impl DataManager {
         }
     }
 
-    pub fn handle_group_merge(&mut self) {
-        unimplemented!()
+    pub fn handle_group_merge(&mut self, prefix: &Prefix<XorName>) {
+        if self.cache.prune_unneeded_chunks(prefix) > 0 {
+            self.cache.print_stats();
```
Fraser999: Don't we need to increment `self.immutable_data_count` by the number of pruned unneeded chunks?

PR #571, opened 2016-10-28
MAID-1643 chore/maid_manager:MM refactor for disjoint group

maqi:


maidsafe-highfive:
r? @Fraser999

src/personas/maid_manager.rs
```diff
@@ -229,44 +230,59 @@ impl MaidManager {
     }
 
     pub fn handle_group_merge(&mut self) {
-        unimplemented!()
+        for (maid_name, account) in &self.accounts {
+            self.send_refresh(maid_name, account, MessageId::zero());
```
afck: Also, half the nodes will send the accounts with `0` and half the nodes the ones with `1`, so if we send this as the `ClientManager`, neither will accumulate (if e.g. quorum means majority).

Maybe add a TODO about this for now. Depending on how message accumulation will change in routing, we might have to manually accumulate these in vault.

maqi: Yes, I agree if `quorum` means majority, there will be issue in accumulation.
will put a TODO in.

PR #572, opened 2016-11-02
MAID-1679 chore/maid_manager_tests: re-enable MM network tests

maqi:


maidsafe-highfive:
r? @Fraser999

src/mock_crust_detail/mod.rs
```diff
@@ -75,6 +75,7 @@ pub fn check_data(all_data: Vec<Data>, nodes: &[TestNode]) {
             .sorted_by(|left, right| data_id.name().cmp_distance(left, right));
 
         expected_data_holders.truncate(MIN_GROUP_SIZE);
+        data_holders.truncate(MIN_GROUP_SIZE);
```
afck: We could just pick the single closest node to the data, and verify that the data holders are exactly its close group. Then we wouldn't need to truncate and ignore the other holders.

src/mock_crust_detail/test_client.rs
```diff
@@ -327,25 +327,28 @@ impl TestClient {
...
+                    }, .. }) => {
+                    assert_eq!(request_message_id, response_id);
+                    assert!(data.identifier() == data_id);
+                    let parsed_error = unwrap!(serialisation::deserialise(&response_error));
+                    result = Err(Some(parsed_error));
+                    trace!("returning with {:?}", result);
+                }
+                Ok(response) => println!("Unexpected Put response : {:?}", response),
+                // TODO: Once the network guarantees that every request gets a response, panic!
+                Err(_) => return result,
```
afck: Is this because there were several responses? If so, that's a bug that we should fix (probably in routing), not adapt the tests for it.

tests/maid_manager.rs
```diff
@@ -124,38 +126,35 @@ fn handle_put_with_account() {
...
-            "client account {} found on {} nodes",
-            count,
-            node_count);
+    // let count = nodes.iter()
+    //     .filter(|node| node.get_maid_manager_put_count(client.name()).is_some())
+    //     .count();
+    // assert!(MIN_GROUP_SIZE == count,
+    //         "client account {} found on {} nodes",
+    //         count,
+    //         node_count);
```
afck: Shouldn't we be able to assert `count >= MIN_GROUP_SIZE` now?

tests/maid_manager.rs
```diff
@@ -337,7 +329,7 @@ fn maid_manager_account_decrease_with_churn() {
 
     for i in 0..test_utils::iterations() as u64 {
         trace!("Churning on {} nodes, iteration {}", nodes.len(), i);
-        if nodes.len() <= MIN_GROUP_SIZE + 2 || rng.gen() {
+        if nodes.len() <= MIN_GROUP_SIZE + 4 || rng.gen() {
```
afck: Why do we need `+ 4` here, now?

src/mock_crust_detail/test_client.rs
maqi: Yes, the client will receive duplicated responses for the same msg (exactly two copies).

afck: @Fraser999: Do you think this is the same issue you're seeing? Then let's wait for your fix first. The test should then pass without this change here.

Fraser999: Yes, I think it is.  Working on that now.

PR #575, opened 2016-11-07
MAID-1690 refuse to create account if src and dst don't match

afck:


maidsafe-highfive:
r? @maqi

src/personas/maid_manager.rs
```diff
@@ -298,6 +298,12 @@ impl MaidManager {
         // exist.
         let client_name = utils::client_name(&src);
         if data.get_type_tag() == TYPE_TAG_SESSION_PACKET {
+            if dst != Authority::ClientManager(client_name) {
```
maqi: just curious, will `dst.name() == &client_name` will be enough?
as this MM won't receive this request if dst is not Authority::ClientManager() ?

PR #581, opened 2016-11-14
MAID-1708 chore/vault: update code for routing's new data_type API

maqi:


maidsafe-highfive:
r? @Fraser999

src/mock_crust_detail/test_client.rs
```diff
@@ -100,8 +101,11 @@ impl TestClient {
 
     /// create an account and store it
     pub fn create_account(&mut self, nodes: &mut [TestNode]) {
-        let account =
-            unwrap!(StructuredData::new(0, self.rng.gen(), 0, vec![], vec![], vec![], None));
+        let owner_pubkey = *self.full_id.public_id().signing_public_key();
+        let mut owner = BTreeSet::new();
```
afck: `iter::once(owner_pubkey).collect()` would be more concise. (Also in the other files.)

src/test_utils.rs
```diff
@@ -71,34 +74,43 @@ pub fn random_pub_appendable_data_with_size<R: Rng>(full_id: &FullId,
...
+                                        BTreeSet::new(),
+                                        Filter::black_list(None))
+        .expect("Cannot create public appendable data for test");
+
+    for _ in 0..size / 128 {
+        let pointer = DataIdentifier::Structured(rng.gen(), 12345);
+        let appended_data = unwrap!(AppendedData::new(pointer,
+                                                      full_id.public_id()
+                                                          .signing_public_key()
+                                                          .clone(),
```
afck: This could now be `owner_pubkey`.

afck:
We'll also need to update the build scripts, so that they don't time out while compiling: Both nightly builds did.

PR #583, opened 2016-11-15
MAID-1700 debug/data_manager: rejected request shall put into pending as well

maqi:


maidsafe-highfive:
r? @Fraser999

src/personas/data_manager.rs
```diff
@@ -461,7 +462,7 @@ impl DataManager {
                                               data_id,
                                               external_error_indicator,
                                               message_id);
-                        return Err(From::from(error));
+                        valid = false;
```
afck: The `send_put_failure` call could be deduplicated by replacing `valid` with an `opt_error: Option<MutationError>`.
You could set that to `Some(DataExists)` here, and below only check whether the chunk store is full if it is `None`, and then set it to `Some(NetworkFull)`.

src/personas/data_manager.rs
```diff
@@ -519,8 +523,8 @@ impl DataManager {
                           data_id,
                           message_id);
                     let post_error = serialisation::serialise(&MutationError::InvalidOperation)?;
-                    return Ok(self.routing_node
-                        .send_post_failure(dst, src, data_id, post_error, message_id)?);
+                    let _ = self.routing_node
```
afck: Also in line 557. (But not in 545, I guess?)

src/personas/data_manager.rs
```diff
@@ -568,17 +572,25 @@ impl DataManager {
 
         let error = match self.chunk_store.get(&data_id) {
             Ok(Data::Structured(mut data)) => {
-                if data.is_deleted() {
+                let error = if data.is_deleted() {
```
afck: Just like above, you could make this error optional and call `update_pending_writes(..., opt_error.is_none())` only once.
In line 598, you could just do `if let Err(error) = opt_error` …

src/personas/data_manager.rs
afck: That's what I mean: Only check whether the chunk store is full if the error is still `None`. Then it would effectively do the same thing as it does now.

src/personas/data_manager.rs
maqi: 557 cannot change to this as a `data` is required to be inserted into pending.
However, 557 means no available updated_data, and we have to use incoming new_data.
for SD, it is ok, as new_data will be equal to the updated_data.
but for AD, it is not equal, so push new_data into pending will be problematic, as it will be inconsistent with others.

btw. the returned result after line 526 shall be new_data directly.

src/personas/data_manager.rs
maqi: but if the error is not None, we still need to check whether chunk_store is full. As if it is full, the rejected request is no need to be put into the pending writes.

src/personas/data_manager.rs
```diff
@@ -503,58 +507,64 @@ impl DataManager {
...
-            (_, Err(error)) => {
-                trace!("DM sending post_failure for: {:?} with {:?} - {:?}",
-                       data_id,
-                       message_id,
-                       error);
-                let post_error = serialisation::serialise(&MutationError::NoSuchData)?;
-                return Ok(self.routing_node
-                    .send_post_failure(dst, src, data_id, post_error, message_id)?);
+            (Data::Structured(new_sd), Err(_)) => {
+                    warn!("Post operation for non-exists data. {:?} - {:?}",
```
afck: Typo: `nonexistent`.

src/personas/data_manager.rs
```diff
@@ -503,58 +507,64 @@ impl DataManager {
...
-                result.map(|()| Data::Structured(sd))
-            }
-            (Data::PubAppendable(new_ad), Ok(Data::PubAppendable(mut ad))) => {
-                let result = ad.update_with_other(new_ad);
-                result.map(|()| Data::PubAppendable(ad))
-            }
-            (Data::PrivAppendable(new_ad), Ok(Data::PrivAppendable(mut ad))) => {
-                let result = ad.update_with_other(new_ad);
-                result.map(|()| Data::PrivAppendable(ad))
+                Ok(Data::Structured(new_sd))
```
afck: We still need to call `replace_with_other` here to check the version and permissions. But I agree that we should return `new_sd` like this even in case of an error, and instead set `error_opt` if it fails. That way, we'll still put the new data into the `pending_writes`.

PR #584, opened 2016-11-16
MAID-1701 fix/data_manager: try to fetch data in case of missed pending_writes

maqi:


maidsafe-highfive:
r? @Fraser999

src/personas/data_manager.rs
```diff
@@ -886,12 +886,19 @@ impl DataManager {
...
-        for PendingWrite { mutate_type, src, dst, data, message_id, .. } in self.cache
-            .remove_expired_writes() {
-            let data_id = data.identifier();
-            let error = MutationError::NetworkOther("Request expired.".to_owned());
-            trace!("{:?} did not accumulate. Sending failure", data_id);
-            self.send_failure(mutate_type, src, dst, data_id, message_id, error)?;
+        for PendingWrite { data, .. } in self.cache.remove_expired_writes() {
+            if let Ok(Some(group)) = self.routing_node.close_group(*data.name()) {
+                let candidates: Vec<_> = group.iter().take(1).collect();
+                let data_idv = match self.to_id_and_version(data.identifier()) {
```
Fraser999: Shouldn't we construct the ID and version from `data` rather than calling `self.to_id_and_version()`, since that fails for mutable data which we don't hold in our chunkstore, and even if the chunkstore call succeeded we'd be getting an older version than the one in `data`?

src/personas/data_manager.rs
```diff
@@ -886,12 +886,19 @@ impl DataManager {
...
+                    Some(data_idv) => data_idv,
+                    None => continue,
+                };
+                self.cache.insert_into_ongoing_gets(candidates[0], &data_idv);
+                let src = Authority::ManagedNode(self.routing_node.name()?);
```
Fraser999: This would be best done outside the loop to minimise the calls to `routing_node` I think.

src/personas/data_manager.rs
```diff
@@ -886,12 +886,19 @@ impl DataManager {
                              message_id: MessageId,
                              refresh: bool)
                              -> Result<(), InternalError> {
-        for PendingWrite { mutate_type, src, dst, data, message_id, .. } in self.cache
-            .remove_expired_writes() {
-            let data_id = data.identifier();
-            let error = MutationError::NetworkOther("Request expired.".to_owned());
-            trace!("{:?} did not accumulate. Sending failure", data_id);
-            self.send_failure(mutate_type, src, dst, data_id, message_id, error)?;
+        for PendingWrite { data, .. } in self.cache.remove_expired_writes() {
```
Fraser999: This `data` hides the parameter passed into the function - maybe it could be renamed?

maqi: this `data` is member of `PendingWrites`, can't be renamed. :(

src/personas/data_manager.rs
maqi: I am not sure, each expired entry needs to be handled anyway.
so I don't think inside/outside loop make any difference?

src/personas/data_manager.rs
afck: I think `for PendingWrite { data: exp_data, .. }` would work, too.

src/personas/data_manager.rs
```diff
@@ -805,8 +805,20 @@ impl DataManager {
...
+        if pending_writes.is_empty() {
+            if let Ok(Some(group)) = self.routing_node.close_group(*data_id.name()) {
+                let candidates: Vec<_> = group.iter().take(1).collect();
+                self.cache.insert_into_ongoing_gets(candidates[0], &(data_id, version));
+                let src = Authority::ManagedNode(self.routing_node.name()?);
+                let dst = Authority::ManagedNode(*candidates[0]);
+                let _ = self.routing_node.send_get_request(src, dst, data_id, MessageId::new());
+            }
+            return Ok(());
+        }
```
afck: I think we shouldn't only do that if `pending_writes` is empty, but also if it doesn't contain a matching entry. So I'd rather check in the loop below whether `hash == refresh_hash` is ever true, and if it isn't, I'd put the _whole_ close group into the data holders for that data chunk. After that, we can just call `send_gets_for_needed_data` instead of sending the request explicitly.

maqi: if hash doesn't match, the `get` shall not be sent to individual, but to the group.
otherwise, will be difficult to tell which version to trust

afck: I think we'll need to check the hash when we receive it anyway, but we have a separate issue for that: https://maidsafe.atlassian.net/browse/MAID-1703

src/personas/data_manager.rs
```diff
@@ -848,6 +872,10 @@ impl DataManager {
                 trace!("{:?} did not accumulate. Sending failure", data_id);
                 let error = MutationError::NetworkOther("Concurrent modification.".to_owned());
                 self.send_failure(mutate_type, src, dst, data.identifier(), message_id, error)?;
+
+                let src = Authority::ManagedNode(self.routing_node.name()?);
+                let dst = Authority::NaeManager(*data_id.name());
+                let _ = self.routing_node.send_get_request(src, dst, data_id, MessageId::new());
```
afck: We don't need to send a request for the data for _every_ entry that doesn't match the hash.
We only need to send a single such request if there was _no_ entry with the right hash at all.
But I agree that it's a good idea to just send this request to the whole group.

src/personas/data_manager.rs
```diff
@@ -805,8 +817,20 @@ impl DataManager {
...
+        if pending_writes.is_empty() {
+            if let Ok(Some(group)) = self.routing_node.close_group(*data_id.name()) {
+                let data_idv = (data_id, version);
+                for node in &group {
+                    let _ = self.cache.register_data_with_holder(node, &data_idv);
+                }
+                self.send_gets_for_needed_data()?;
+            }
+            return Ok(());
+        }
```
afck: I don't think an empty `pending_writes` needs any special handling that is different from a non-empty one which is missing the right hash. It should be sufficient to _either_ request the data from the whole group _or_ put everyone else into the data holders.

src/personas/data_manager.rs
```diff
@@ -886,13 +914,16 @@ impl DataManager {
...
-            let error = MutationError::NetworkOther("Request expired.".to_owned());
-            trace!("{:?} did not accumulate. Sending failure", data_id);
-            self.send_failure(mutate_type, src, dst, data_id, message_id, error)?;
+        for PendingWrite { data: exp_data, .. } in self.cache.remove_expired_writes() {
+            if let Ok(Some(group)) = self.routing_node.close_group(*exp_data.name()) {
+                let data_idv = id_and_version_of(&exp_data);
+                for node in &group {
+                    let _ = self.cache.register_data_with_holder(node, &data_idv);
+                }
+            }
```
afck: Why would we register _expired_ writes that didn't accumulate to the data holders? Maybe I'm missing something but I don't think this method needs to be changed at all.

maqi: this is regarding the comment put in the jira task.
i.e. pending_writes `expired` due to receiving `accumulated refresh` first.

src/personas/data_manager.rs
maqi: `mis_matched hash` means there could be conflict happens in the group, which has to be resolved by fetching from group. (or a bit either risky or complicated way of fetching one by one)
`empty pending_writes` means totally missed something. A fetch from peer shall be enough to recover. Fetching from group will resolve as well, but will be a over-shot incurring too much msg exchange.

src/personas/data_manager.rs
afck: Your comment about the failing test? I'm pretty sure the failing test is still just because of a routing bug, not one in the data manager. I also don't see the connection with _expired_ pending writes.

src/personas/data_manager.rs
afck: Both cases simply mean that we didn't get the request that caused the pending write which the group agreed on. In both cases we need to request that data chunk from someone else. Either the group, or just another individual member - the trust issue should be fixed with https://maidsafe.atlassian.net/browse/MAID-1703.

PR #586, opened 2016-11-22
refactor/tests: re-enable some mock-crust tests

Fraser999:
Also renames any "test" mods to "tests".

Although the diff for [tests/network.rs](https://github.com/maidsafe/safe_vault/compare/dev...Fraser999:MAID-1679?expand=1#diff-fc594075c6df4aa13e74822d88e34a9e) is large, the only change there was to remove the `mod test` scope and run rustfmt.

### DIFF
--- a/tests/data_manager.rs
+++ b/tests/data_manager.rs
@@ -15,12 +15,8 @@
 // Please review the Licences for the specific language governing permissions and limitations
 // relating to use of the SAFE Network Software.
 
-// For explanation of lint checks, run `rustc -W help` or see
-// https://github.com/maidsafe/QA/blob/master/Documentation/Rust%20Lint%20Checks.md
-
 use rand::Rng;
 use rand::distributions::{IndependentSample, Range};
-
 use routing::{AppendWrapper, AppendedData, Authority, Data, DataIdentifier, Event, FullId,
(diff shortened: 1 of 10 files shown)


tests/data_manager.rs
```diff
@@ -1084,22 +1040,12 @@ fn caching_with_data_close_to_proxy_node() {
...
-
-fn gen_random_immutable_data_not_close_to<R: Rng>(_node: &TestNode, _rng: &mut R) -> Data {
-    unimplemented!();
-    // loop {
-    //     let data = Data::Immutable(test_utils::random_immutable_data(10, rng));
-    //     if !node.routing_table().is_close(&data.name(), MIN_GROUP_SIZE) {
-    //         return data;
-    //     }
-    // }
+fn generate_cache_data<R: Rng>(node: &TestNode, rng: &mut R, close_to_node: bool) -> Data {
```
maqi: the function name doesn't indicate the data needs to be generated with range expectation?

Fraser999: Yeah - wasn't sure how to indicate it could be close to _or_ not close to `node` without having a huge function name!

afck: Can't think of a better name either. And I think the parameter name makes that clear enough maybe.

PR #587, opened 2016-11-24
MAID-1790 MAID-1792 Refactor some mock-crust tests and helpers

Fraser999:


src/personas/data_manager.rs
```diff
@@ -897,16 +898,18 @@ impl DataManager {
...
             let error = MutationError::NetworkOther("Request expired.".to_owned());
             trace!("{:?} did not accumulate. Sending failure", data_id);
             self.send_failure(mutate_type, src, dst, data_id, message_id, error)?;
         }
         let data_name = *data.name();
-        if let Some(refresh_data) = self.cache
-            .insert_pending_write(data, mutate_type, src, dst, message_id) {
+        if let Some(refresh_data) =
+            self.cache
+                .insert_pending_write(data, mutate_type, src, dst, message_id) {
```
afck: Not important, but: would rustfmt allow these two to be on one line, since it seems to break after the `=` now?

src/mock_crust_detail/test_node.rs
```diff
@@ -157,6 +162,16 @@ pub fn create_nodes(network: &Network,
     nodes
 }
 
+/// Create a mock network of nodes handling cache data with two disjoint groups
+pub fn create_nodes_with_cache_till_split(network: &Network) -> Vec<TestNode> {
```
afck: (Should probably be renamed to `til` or `until` at some point.)

PR #588, opened 2016-11-25
MAID-1795 Fix failing network::tests::fill_network test

dhardy:
Tests pass. Rustfmt done. Clippy didn't want to install :(

### DIFF
--- a/tests/network.rs
+++ b/tests/network.rs
@@ -16,15 +16,13 @@
 // relating to use of the SAFE Network Software.
 
 use rand::Rng;
-use rand::distributions::{IndependentSample, Range};
-use routing::{Data, ImmutableData};
+use routing::{Data, ImmutableData, MIN_GROUP_SIZE};
 use routing::mock_crust::{self, Network};
 use safe_vault::Config;
-use safe_vault::mock_crust_detail::{poll, test_node};
 use safe_vault::mock_crust_detail::test_client::TestClient;
+use safe_vault::mock_crust_detail::test_node;
 use safe_vault::test_utils;
 
-#[ignore]
 #[test]
 fn fill_network() {
     let network = Network::new(None);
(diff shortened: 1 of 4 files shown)


maidsafe-highfive:

src/chunk_store/tests.rs
```diff
@@ -113,11 +113,10 @@ fn successful_put() {
...
 
-        for (index, &(ref data, ref size)) in
-            chunks.data_and_sizes
-                .iter()
-                .enumerate()
-                .rev() {
+        for (index, &(ref data, ref size)) in chunks.data_and_sizes
+            .iter()
+            .enumerate()
+            .rev() {
```
afck: This looks like it's undoing some earlier rustfmt changes - are you using version 0.6.3?

tests/network.rs
```diff
@@ -63,25 +62,23 @@ fn fill_network() {
...
-                trace!("Stored chunk {:?}", data_id);
-                return;
-            }
-            Err(opt_response) => {
-                trace!("Failed storing chunk {:?}, response: {:?}",
-                       data_id,
-                       opt_response);
-            }
+    
+    //TODO: this is a bit different than before — is it doing the right thing?
```
afck: Yes, that makes sense. Since there was only one group before, adding until it splits should definitely make space for more data.

tests/network.rs
```diff
@@ -63,25 +62,22 @@ fn fill_network() {
...
-                       opt_response);
-            }
+
+    // Let the network split
+    test_node::add_nodes_until_split(&network, &mut nodes, Some(config), true);
+    // Now try adding data:
+    let content = rng.gen_iter().take(100).collect();
+    let data = Data::Immutable(ImmutableData::new(content));
+    let data_id = data.identifier();
+    match client.put_and_verify(data, &mut nodes) {
```
Fraser999: This `match` could probably be reduced to something like
```rust
unwrap!(client.put_and_verify(data, &mut nodes), "Failed to put again after adding nodes");
```
(unless we end up reverting to the previous method of adding a node and trying to store inside a fixed-size loop)

tests/network.rs
```diff
@@ -63,25 +62,22 @@ fn fill_network() {
...
-                return;
-            }
-            Err(opt_response) => {
-                trace!("Failed storing chunk {:?}, response: {:?}",
-                       data_id,
-                       opt_response);
-            }
+
+    // Let the network split
+    test_node::add_nodes_until_split(&network, &mut nodes, Some(config), true);
```
Fraser999: I don't think this guarantees that the following `put_and_verify()` attempt will succeed.  If the network happens to be somewhat imbalanced prior to this call, we can end up with one of the new groups having a quorum of "full" nodes.  So if the random test data belongs in that group, it will still report `NetworkFull` I think.

PR #591, opened 2016-12-19
Maid-1872 port fixes

maqi:


maidsafe-highfive:
r? @Fraser999

src/personas/data_manager.rs
```diff
@@ -324,14 +332,25 @@ impl Cache {
...
+            let mut result = Some(RefreshData((data_id, version), hash));
+            match self.pending_writes.get(&data_id) {
+                None => {}
+                Some(entries) => {
+                    if entries.iter().any(|pending_write| !pending_write.rejected) {
+                        result = None;
+                    }
+                }
+            }
+            result
```
afck: That seems unnecessarily convoluted. Isn't the rest of this method equivalent to:
```rust
let writes = self.pending_writes.entry(data_id).or_insert_with(Vec::new);
let result = if !rejected && writes.iter().all(|pending_write| pending_write.rejected) {
    Some(RefreshData((data_id, version), hash))
} else {
    None
};
writes.insert(0, pending_write);
result
```

PR #592, opened 2017-01-04
MAID-1864: Merge routing and vault event loops.

michaelsproul:
Requires https://github.com/maidsafe/routing/pull/1305, which is now merged.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,7 +22,7 @@ log = "~0.3.6"
 maidsafe_utilities = "~0.10.0"
 quick-error = "~1.1.0"
 rand = "~0.3.14"
-routing = { git = "https://github.com/michaelsproul/routing.git", branch = "event-stream-refactor" }
+routing = { git = "https://github.com/maidsafe/routing.git", branch = "master" }
 rustc-serialize = "~0.3.19"
 rust_sodium = "~0.1.2"
 unwrap = "~1.1.0"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,7 +22,7 @@ log = "~0.3.6"
 maidsafe_utilities = "~0.10.0"
 quick-error = "~1.1.0"
 rand = "~0.3.14"
-routing = { git = "https://github.com/maidsafe/routing.git", branch = "master" }
+routing = { git = "https://github.com/michaelsproul/routing.git", branch = "event-stream-refactor" }
 rustc-serialize = "~0.3.19"
 rust_sodium = "~0.1.2"
 unwrap = "~1.1.0"
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -198,12 +198,12 @@
 // For explanation of lint checks, run `rustc -W help` or see
 // https://github.com/maidsafe/QA/blob/master/Documentation/Rust%20Lint%20Checks.md
 #![forbid(bad_style, exceeding_bitshifts, mutable_transmutes, no_mangle_const_items,
-          unknown_crate_types, warnings)]
+          unknown_crate_types)]
 #![deny(deprecated, improper_ctypes, missing_docs,
         non_shorthand_field_patterns, overflowing_literals, plugin_as_library,
         private_no_mangle_fns, private_no_mangle_statics, stable_features, unconditional_recursion,
         unknown_lints, unsafe_code, unused, unused_allocation, unused_attributes,
-        unused_comparisons, unused_features, unused_parens, while_true)]
+        unused_comparisons, unused_features, unused_parens, warnings, while_true)]
 #![warn(trivial_casts, trivial_numeric_casts, unused_extern_crates, unused_import_braces,
         unused_qualifications, unused_results)]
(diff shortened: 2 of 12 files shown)


src/lib.rs
```diff
@@ -198,12 +198,12 @@
...
 // https://github.com/maidsafe/QA/blob/master/Documentation/Rust%20Lint%20Checks.md
 #![forbid(bad_style, exceeding_bitshifts, mutable_transmutes, no_mangle_const_items,
-          unknown_crate_types, warnings)]
+          unknown_crate_types)]
 #![deny(deprecated, improper_ctypes, missing_docs,
         non_shorthand_field_patterns, overflowing_literals, plugin_as_library,
         private_no_mangle_fns, private_no_mangle_statics, stable_features, unconditional_recursion,
         unknown_lints, unsafe_code, unused, unused_allocation, unused_attributes,
-        unused_comparisons, unused_features, unused_parens, while_true)]
+        unused_comparisons, unused_features, unused_parens, warnings, while_true)]
```
michaelsproul: I had to change it from a `forbid` (can't be turned off) to a `deny` so that I could do `allow(too_many_arguments)` on some of the functions that now take a `&mut RoutingNode`.

src/lib.rs
```diff
@@ -225,7 +225,6 @@ extern crate maidsafe_utilities;
 extern crate config_file_handler;
 // Needed because the crate is only used for macros
 #[cfg_attr(feature="clippy", allow(useless_attribute))]
-#[allow(unused_extern_crates)]
```
michaelsproul: Yep, good catch!

src/personas/maid_manager.rs
```diff
@@ -110,38 +109,41 @@ impl MaidManager {
...
             Some((src, dst)) => {
                 // Send success response back to client
                 let client_name = utils::client_name(&src);
-                self.send_refresh(&client_name,
+                self.send_refresh(routing_node,
+                                  &client_name,
                                   self.accounts.get(&client_name).expect("Account not found."),
                                   MessageId::zero());
-                let _ = self.routing_node.send_put_success(dst, src, data_id, msg_id);
+                routing_node.send_put_success(dst, src, data_id, msg_id)?;
```
michaelsproul: Another good catch! I'm not really sure why I changed this, I'll put it back how it was.

src/lib.rs
Fraser999: I don't think that requires "unforbidding" `warnings` though.  That would only be if you wanted to `allow(warnings)`.

src/personas/maid_manager.rs
Fraser999: I do agree that there are way too many places where we ignore returns - I itch to change these too :)  Just need to keep these changes until a planned task unfortunately!

michaelsproul:
Thanks for reviewing @Fraser999! Regarding the removal of `Rc<RoutingNode>` from `DataManager` and `MaidManager`, that was required because most of the methods they call on `RoutingNode` now require mutable access to the routing node (see: [send_action][] in routing). The reason for the change in routing was the removal of the `event_sender` channel, which essentially hid mutation and side-effects behind a regular `&self` reference (as `mpsc::Sender::send` can be used with a shared/immutable reference).

[send_action]: https://github.com/maidsafe/routing/blob/4b611109e13ecf9f42706718b0bce28aafb73b5d/src/node.rs#L400-L425

src/lib.rs
```diff
@@ -235,7 +234,6 @@ extern crate rustc_serialize;
 extern crate rust_sodium;
 #[cfg(test)]
 extern crate tempdir;
-#[cfg(any(test, feature = "use-mock-crust"))]
```
michaelsproul: Not required, I'm putting this line back

src/mock_crust_detail/mod.rs
```diff
@@ -49,9 +49,9 @@ pub fn check_deleted_data(deleted_data: &[Data], nodes: &[TestNode]) {
 }
 
 /// Checks that the given `nodes` store the expected number of copies of the given data.
-pub fn check_data(all_data: Vec<Data>, nodes: &[TestNode]) {
+pub fn check_data(all_data: Vec<Data>, nodes: &mut [TestNode]) {
```
michaelsproul: I changed this back. My change was a relic from when I had `RoutingNode::name` requiring a mutable reference

Cargo.toml
```diff
@@ -22,7 +22,7 @@ log = "~0.3.6"
 maidsafe_utilities = "~0.10.0"
 quick-error = "~1.1.0"
 rand = "~0.3.14"
-routing = { git = "https://github.com/maidsafe/routing.git", branch = "master" }
+routing = { git = "https://github.com/michaelsproul/routing.git", branch = "event-stream-refactor" }
```
michaelsproul: Done! (points to master now)

src/lib.rs
michaelsproul: I dunno what was going on, but I've reverted this change and clippy has stopped yelling

PR #605, opened 2017-02-23
Mutable data handling

madadam:


maidsafe-highfive:
r? @maqi

src/personas/maid_manager/account.rs
```diff
@@ -0,0 +1,111 @@
...
+// Licenses can be found in the root directory of this project at LICENSE, COPYING and CONTRIBUTOR.
+//
+// Unless required by applicable law or agreed to in writing, the SAFE Network Software distributed
+// under the GPL Licence is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.
+//
+// Please review the Licences for the specific language governing permissions and limitations
+// relating to use of the SAFE Network Software.
+
+use routing::AccountInfo;
```
ustulation: Why is account-info held with routing?

src/personas/maid_manager/mod.rs
```diff
@@ -0,0 +1,803 @@
...
+    pub fn handle_put_idata_success(&mut self,
+                                    routing_node: &mut RoutingNode,
+                                    msg_id: MessageId)
+                                    -> Result<(), InternalError> {
+        let (src, dst) = self.remove_cached_request(msg_id)?;
+
+        // Send success response back to client
+        let client_name = utils::client_name(&src);
+        // TODO (adam): are we sure this can't panic?
+        let account = self.accounts.get(&client_name).expect("Account not found.").clone();
```
ustulation: Would this not panic if it's a new node that's just joined the group, has not yet got the user-account-info but has got idata-put-success instead ? Or maybe there could be other such cases. I think usage of any panic should be converted to `error!` so that it shows up something went wrong with our logic but does not immediately cause the whole program to crash ?

src/personas/maid_manager/mod.rs
```diff
@@ -0,0 +1,803 @@
...
+                routing_node.send_put_mdata_response(dst, src, Err(err.clone()), msg_id)?;
+                return Err(From::from(err));
+            }
+
+            // Create the account.
+            let _ = self.accounts.insert(client_name, Account::default());
+            info!("Managing {} client accounts.", self.accounts.len());
+        }
+
+        if let Err(err) = self.increment_mutation_counter(&client_name) {
```
ustulation: Should we not remove the account in this case ? Letting it get created but sending an error seems strange although it logically we should not hit a low balance for something just created (unless we wrongly coded the cap at 1 mutation or something). Still, the clean up codes should do proper clean up, such as rollback what was done (since it's all at one place here, so easier) ? Not sure on this though.

src/personas/maid_manager/mod.rs
```diff
@@ -0,0 +1,803 @@
...
+    pub fn handle_put_mdata_success(&mut self,
+                                    routing_node: &mut RoutingNode,
+                                    msg_id: MessageId)
+                                    -> Result<(), InternalError> {
+        let (src, dst) = self.remove_cached_request(msg_id)?;
+
+        // Send success response back to client
+        let client_name = utils::client_name(&src);
+        // TODO (adam): are we sure this can't panic?
+        let account = self.accounts.get(&client_name).expect("Account not found.").clone();
```
ustulation: See previous comment on this topic of panic

src/personas/maid_manager/mod.rs
```diff
@@ -0,0 +1,803 @@
...
+                                    error: ClientError,
+                                    msg_id: MessageId)
+                                    -> Result<(), InternalError> {
+        let (src, dst) = self.remove_cached_request(msg_id)?;
+
+        if !self.handle_put_failure(routing_node, &src) {
+            return Ok(());
+        }
+
+        // TODO (adam): originally, we had special handling for session packets
```
ustulation: MD still has type tags - so not sure what this comment means. We can distinguish login packet MD which has a type tag of `TYPE_TAG_SESSION_PACKET`. [That is what safe-core does too.](https://github.com/maidsafe/safe_client_libs/blob/28978cd10aad433332950284a49e96d881242ef9/safe_core/src/client/mod.rs#L197)

src/personas/maid_manager/mod.rs
```diff
@@ -0,0 +1,803 @@
...
+                              src: &Authority<XorName>,
+                              dst: &Authority<XorName>)
+                              -> Result<(), ClientError> {
+        let client_manager_name = utils::client_manager_name(dst);
+
+        if let Some(account) = self.accounts.get_mut(&client_manager_name) {
+            let client_key = utils::client_key(src);
+            let client_name = utils::client_name_from_key(&client_key);
+
+            if client_name == client_manager_name || account.auth_keys.contains(&client_key) {
```
ustulation: This check will allow any listed auth-key to transfer MD ownership too. This is not what we want. Only owners should be able to do this. So calling this function [from here](https://github.com/maidsafe/safe_vault/pull/605/files#diff-0ecf02157a05b27233d32af1eee457adR313) would be wrong for instance.

src/personas/maid_manager/mod.rs
```diff
@@ -0,0 +1,803 @@
...
+        }
+    }
+
+    pub fn handle_change_mdata_owner(&mut self,
+                                     routing_node: &mut RoutingNode,
+                                     src: Authority<XorName>,
+                                     dst: Authority<XorName>,
+                                     msg_id: MessageId)
+                                     -> Result<(), InternalError> {
+        if let Err(err) = self.prepare_mdata_mutation(&src, &dst) {
```
ustulation: This is not a sufficient condition for this request. Check https://github.com/maidsafe/safe_vault/pull/605/files#diff-0ecf02157a05b27233d32af1eee457adR313

ustulation:
Where is the verification of the `requester` field ? We are supposed to verify in the MMs that `requester == sender-client` else DMs would not be able to verify permissions for incoming request to them. I could have missed it but could you pls point to where it is here in this change-set ?

src/personas/maid_manager/account.rs
madadam: It's part of `Response`. I reused it in `Account`, because I needed the same fields.

madadam:
> Where is the verification of the requester field ?
Currently nowhere. Wasn't sure how to handle it, then forgot to bring it up :grimacing: Anyway, where would you put it - in `MaidManager` or `DataManager`?

ustulation:
@madadam 

> Currently nowhere. Wasn't sure how to handle it, then forgot to bring it up :grimacing: Anyway, where would you put it - in `MaidManager` or `DataManager`?

Ah right, nw. It should be in MM, DM's wont be able to verify anyway. By the the info reaches to them the orig sender info is lost - only latest sender which here woul be the MM group auth. That is why MMs use the cache to remap src and dst fields once response comes from DMs to send it further to orig sender - the client auth.

src/personas/maid_manager/mod.rs
ustulation: hmm .. we can store it in the cache before forwarding to dm's? because there might be many special tags and thus handling in the future, so worth storing some more info in request-cache ? But ya would wait for others to comment on this too for the best way - pls feel free to suggest if u have something too :)

madadam: Yeah, storing it in the request cache would probably be quickest to implement. Alternatively, we can put it into `Response`, where it used to be before. Something like:

```rust
enum Response {
    // ...

    PutMData {
        res: Result<(), (ClientError, DataId)>,
        msg_id: MessageId,
    }
}
```

or

```rust
PutMData {
    data_id: DataId,
    res: Result<(), ClientError>,
    msg_id: MessageId,
}
```

That would probably be more invasive change.

src/personas/maid_manager/mod.rs
madadam: Right. That check must happen in `DataManager` though, because `MaidManager` doesn't have access to the data itself, so it doesn't know who the owner is.

madadam: Actually, just found out this check's been already there, but it happens in `DataManager`. I added an explicit test for it though.

src/personas/data_manager/mod.rs
```diff
@@ -0,0 +1,1248 @@
...
+
+const MAX_FULL_PERCENT: u64 = 50;
+/// The quorum for accumulating refresh messages.
+const ACCUMULATOR_QUORUM: usize = GROUP_SIZE / 2 + 1;
+/// The timeout for accumulating refresh messages.
+const ACCUMULATOR_TIMEOUT_SECS: u64 = 180;
+/// The interval for print status log.
+const STATUS_LOG_INTERVAL: u64 = 120;
+
+macro_rules! log_status {
```
michaelsproul: This definitely needs to be factored out, but couldn't it be a function rather than a macro?

michaelsproul: (sorry, just a nitpick, I would like to help review this more broadly though)

madadam: The reason it's a macro is so the file/line number in the log points to its invocation rather than definition.

src/personas/maid_manager/mod.rs
```diff
@@ -0,0 +1,695 @@
...
+                            data: ImmutableData,
+                            msg_id: MessageId)
+                            -> Result<(), InternalError> {
+        if !data.validate_size() {
+            routing_node
+                .send_put_idata_response(dst, src, Err(ClientError::DataTooLarge), msg_id)?;
+            return Ok(());
+        }
+
+        if let Err(err) = self.increment_mutation_counter(&utils::client_name(&src)) {
```
ustulation: How is this going to work for Apps ? The clients they use (`src` in above) are not the same as the client the owner would use and that is the one with the client account)

src/personas/maid_manager/mod.rs
```diff
@@ -0,0 +1,695 @@
...
+                    .send_put_mdata_response(dst, src, Err(err.clone()), msg_id)?;
+                return Ok(());
+            }
+
+            // Create the account.
+            let _ = self.accounts.insert(src_name, Account::default());
+            info!("Managing {} client accounts.", self.accounts.len());
+        }
+
+        if let Err(err) = self.prepare_mdata_mutation(&src, &dst, None, true) {
```
ustulation: If the reqester field is always passed as none, how will an App create MD if it has the permission to ?

src/personas/maid_manager/mod.rs
```diff
@@ -0,0 +1,695 @@
...
+                let _ = self.accounts.remove(&client_name);
+                let refresh = Refresh::Delete(client_name);
+                if let Ok(serialised_refresh) = serialisation::serialise(&refresh) {
+                    trace!("MM sending delete refresh for account {}", src.name());
+                    let _ = routing_node.send_refresh_request(dst, dst, serialised_refresh, msg_id);
+                }
+
+                Err(ClientError::AccountExists)
+            }
+            (_, Err(err)) => Err(err),
```
ustulation: Don't we have to decrement the mutatiion counter in case of a failure ? (Incerment was done in `prepare_mdata_mutation`)

src/personas/maid_manager/mod.rs
```diff
@@ -0,0 +1,695 @@
...
+                                       routing_node: &mut RoutingNode,
+                                       src: Authority<XorName>,
+                                       dst: Authority<XorName>,
+                                       name: XorName,
+                                       tag: u64,
+                                       actions: BTreeMap<Vec<u8>, EntryAction>,
+                                       msg_id: MessageId,
+                                       requester: sign::PublicKey)
+                                       -> Result<(), InternalError> {
+        if let Err(err) = self.prepare_mdata_mutation(&src, &dst, Some(requester), true) {
```
ustulation: This function increments the mutation counter by 1. We should be incrementing it based on the number of `actions` perhaps ?

src/personas/maid_manager/mod.rs
```diff
@@ -0,0 +1,695 @@
...
+            routing_node.send_mutate_mdata_entries_request(src,
+                                                           dst,
+                                                           name,
+                                                           tag,
+                                                           actions,
+                                                           msg_id,
+                                                           requester)?;
+        }
+
+        self.insert_cached_request(msg_id, src, dst, Some(tag));
```
ustulation: This ties into the previous comment in this function. If bulk mutations are all or none, we should also cache the number of them, so that we can undo the mutation counter by the same amount it was incremented.

src/personas/maid_manager/mod.rs
```diff
@@ -0,0 +1,695 @@
...
+                                     routing_node: &mut RoutingNode,
+                                     src: Authority<XorName>,
+                                     dst: Authority<XorName>,
+                                     name: XorName,
+                                     tag: u64,
+                                     new_owners: BTreeSet<sign::PublicKey>,
+                                     version: u64,
+                                     msg_id: MessageId)
+                                     -> Result<(), InternalError> {
+        if let Err(err) = self.prepare_mdata_mutation(&src, &dst, None, false) {
```
ustulation: This `true/false` parameter is very easy to miss in review an difficult to maintian in general. Perhaps it's better to put accept it as a more concrete user-define type to make it much more obvious to the coder as well as the reviewer:
```
enum MutationAllowedBy {
   OwnerOnly,
   AnyoneAuthorised,
}
```

src/personas/maid_manager/mod.rs
```diff
@@ -0,0 +1,695 @@
...
+        if client_name != client_manager_name {
+            // TODO (adam): is this the right error to return here?
+            return Err(ClientError::AccessDenied);
+        }
+
+        if let Some(account) = self.accounts.get_mut(&client_manager_name) {
+            if version == account.version + 1 {
+                f(account)?;
+                account.version = version;
+                Ok(())
```
ustulation: i suspect we can just return an Ok here and finish it off. There needs to be proper refresh and group consensus on the change, just like DM's doing it via pending writes etc.

(Also, is this not required during updating of mutation counters ?)

src/personas/data_manager/mod.rs
```diff
@@ -0,0 +1,1295 @@
...
+
+    pub fn handle_refresh(&mut self,
+                          routing_node: &mut RoutingNode,
+                          src: XorName,
+                          serialised_refresh: &[u8])
+                          -> Result<(), InternalError> {
+        let fragments: Vec<FragmentInfo> = serialisation::deserialise(serialised_refresh)?;
+        for fragment in fragments {
+            if self.cache
+                   .register_needed_fragment_with_holder(fragment.clone(), src) {
```
ustulation: This is a confusing name for what this function is doing - will `additionally_register_needed_fragment_with_another_holder` be less confusing ?

src/personas/data_manager/mod.rs
```diff
@@ -0,0 +1,1295 @@
...
+                                              Duration::from_secs(ACCUMULATOR_TIMEOUT_SECS)),
+               cache: Default::default(),
+               immutable_data_count: 0,
+               mutable_data_count: 0,
+               client_get_requests: 0,
+               logging_time: Instant::now(),
+           })
+    }
+
+    pub fn handle_refresh(&mut self,
```
ustulation: Maybe write a brief description on top of this function to distinguish it from a group-refresh (so in this case it's the vault's responsibility to accumulate and is sent from single-auth to single-auth, unlike `group-refresh` which is sent from `group-auth to group-auth` - but pls check if this is correct explanation).

src/personas/data_manager/cache.rs
```diff
@@ -0,0 +1,688 @@
...
+        }
+
+        // Skip the fragments that already have request ongoing.
+        for fragments in self.needed_fragments.values() {
+            for (fragment, request) in fragments {
+                if request.is_ongoing() {
+                    let _ = result.remove(fragment);
+                }
+            }
+        }
```
ustulation: Can we not combine this entire nested for loop with the set above ? So if we want to skip the ones for which the request is ongoing we could do something like:
```
 +        let mut result = HashMap::default();
 +        for (holder, fragments) in &self.needed_fragments {
              for (fragment, request) in &fragments {  // <<<< Modified
                   if request.is_ongoing() { continue }  // <<<< Modified
 +                result
 +                    .entry(fragment.clone())
 +                    .or_insert_with(Vec::new)
 +                    .push(*holder);
 +            }
 +        }
```

src/personas/data_manager/mod.rs
```diff
@@ -0,0 +1,1295 @@
...
+                mutation,
+                src,
+                dst,
+                message_id,
+                hash,
+                rejected,
+                ..
+            } in self.cache.take_pending_writes(&data_id) {
+            if hash != refresh_hash {
+                if !rejected {
```
ustulation: What is the purpose of this (i.e. why is something that is already known to be rejected still be in our pending writes) ?

src/personas/data_manager/mod.rs
```diff
@@ -0,0 +1,1295 @@
...
+                                                message_id)?;
+                }
+
+                continue;
+            }
+
+            let mutation_type = mutation.mutation_type();
+            let fragments = mutation.fragment_infos();
+
+            if self.handle_pending_mutation(routing_node, src, dst, mutation, message_id)? {
```
ustulation: This is also sending the response, so better to call it `commit_pending_mutation_and_send_resp` ? Had to look it up more than once to remember it's also sending the response back, so would just add clarity if renamed i think.

src/personas/data_manager/mod.rs
```diff
@@ -0,0 +1,1295 @@
...
+                            _requester: sign::PublicKey)
+                            -> Result<(), InternalError> {
+
+        let data_id = DataId::Mutable(*data.name(), data.tag());
+        let rejected = if self.chunk_store.has(&data_id) {
+            trace!("DM sending PutMData failure for data {:?}, it already exists.",
+                   data_id);
+            routing_node
+                .send_put_mdata_response(dst, src, Err(ClientError::DataExists), msg_id)?;
+            true
```
ustulation: If something already exists then why not send the failure response and return immediately (i.e. why still write to pending writes with an extra `rejected: bool` status) ?

src/personas/data_manager/mod.rs
```diff
@@ -0,0 +1,1295 @@
...
+                                            dst,
+                                            mutation_type,
+                                            DataId::Mutable(name, tag),
+                                            Err(error),
+                                            msg_id)
+            }
+        }
+    }
+
+    fn update_pending_writes(&mut self,
```
ustulation: (Optional) rename to `update_pending_writes_and_do_grp_refresh` for clarity - makes it easier to follow code.

src/personas/data_manager/mod.rs
```diff
@@ -0,0 +1,1295 @@
...
+            Ok(Data::Mutable(ref old_data)) => {
+                // The shell is more recent than the data in the chunk store.
+                // Repace the data with the shell, but keep the entries.
+                for (key, value) in old_data.entries() {
+                    shell.mutate_entry_without_validation(key.clone(), value.clone());
+                }
+
+                false
+            }
+            Ok(_) => unreachable!(),
```
ustulation: log `error!` and return instead of unreachable

src/personas/data_manager/mod.rs
```diff
@@ -0,0 +1,1295 @@
...
+        let data_id = DataId::Mutable(name, tag);
+
+        match self.chunk_store.get(&data_id) {
+            Ok(Data::Mutable(mut data)) => {
+                if data.mutate_entry_without_validation(key, value) {
+                    self.clean_chunk_store();
+                    self.chunk_store.put(&data_id, &Data::Mutable(data))?;
+                }
+            }
+            Ok(_) => unreachable!(),
```
ustulation: Same as above (Scan the rest of the file too pls)

src/personas/data_manager/mod.rs
```diff
@@ -0,0 +1,1295 @@
...
+                    self.cache.remove_needed_fragment(&fragment);
+                    Some((name, tag, key.clone()))
+                }
+                _ => None,
+            }
+        } else {
+            None
+        };
+
+        self.request_needed_fragments(routing_node)?;
```
ustulation: Why do we care about requesting in a handle-success function ? Should it not be already reqesting ?
For handle-failure goup of functions it makes sense (as we need to try yet another holder if available for the fragment), but i can't think of a reason why we need do this in handle-success group of functions as well.

src/personas/maid_manager/mod.rs
```diff
@@ -0,0 +1,695 @@
...
+                    Entry::Occupied(mut entry) => {
+                        if entry.get().version < account.version {
+                            trace!("Client account {:?}: {:?}", maid_name, account);
+                            let _ = entry.insert(account);
+                        }
+                    }
+                }
+            }
+            Refresh::Delete(maid_name) => {
+                let _ = self.accounts.remove(&maid_name);
```
madadam: This sounds like a valid concern to me. Possible solution would be to send the account version in `Refresh::Delete` and compare it to the current account version and ignore/fail if they don't match. Would that suffice?

Also, I think this same issue is present in master too.

src/personas/maid_manager/mod.rs
madadam: I think this is a bug, good catch! The logic should be similar to how ti works in `put_mdata`.

src/personas/maid_manager/mod.rs
madadam: requester doesn't matter for MaidManager, it is only there to be forwarded to DataManager. That said, MaidManager should verify it matches the `src` client key, which isn't done here. So it's still a bug.

src/personas/maid_manager/mod.rs
madadam: I though `MutateMDataEntries` is counted as one mutation. Has there been a discussion about this?

src/personas/maid_manager/mod.rs
madadam: That's done in `handle_mutation_response`.

src/personas/maid_manager/mod.rs
madadam: Yeah, as a minimum, we need to send refresh here. Not sure about the group consensus and pending writes thing. We don't seem to do that in master currently.

src/personas/data_manager/mod.rs
madadam: :+1:

src/personas/data_manager/mod.rs
madadam: I think the `additionally_` is redundant. What about just `register_needed_fragment_with_another_holder` ?

src/personas/data_manager/mod.rs
```diff
@@ -0,0 +1,1295 @@
...
+                    } => {
+                        if let Ok(Data::Mutable(data)) =
+                            self.chunk_store.get(&DataId::Mutable(name, tag)) {
+                            match data.get(key) {
+                                None => true,
+                                Some(value) if value.entry_version < version => true,
+                                Some(_) => false,
+                            }
+                        } else {
+                            true
```
madadam: Fixed by statically enforcing that the `ChunkStore::get` returns the right type of data.

src/personas/data_manager/mod.rs
madadam: Don't know, but it was in the original code (in master), so I just copied it here.

src/personas/data_manager/mod.rs
madadam: I have the same question. It's like this in master, so I copied the logic here too.

src/personas/data_manager/cache.rs
madadam: No, because a fragment might be registered with more than one holder and if it has ongoing request on at least one of them, we want to exclude it.

src/personas/data_manager/mod.rs
madadam: Good question. Again, I just copied the logic from master here: https://github.com/maidsafe/safe_vault/blob/master/src/personas/data_manager.rs#L727

src/personas/data_manager/mod.rs
```diff
@@ -0,0 +1,1295 @@
...
+                    FragmentInfo::MutableDataEntry {
+                        name,
+                        tag,
+                        ref key,
+                        version,
+                        ..
+                    } => {
+                        if let Ok(Data::Mutable(data)) =
+                            self.chunk_store.get(&DataId::Mutable(name, tag)) {
+                            match data.get(key) {
```
madadam: `map_or` is not defined for `Result` (only for `Option`). It can probably be rewritten to a one-liner by some other means but not sure it's worth it.

src/personas/data_manager/mod.rs
```diff
@@ -0,0 +1,1329 @@
...
+        let res = self.mutate_mdata(name, tag, f);
+
+        match res {
+            Ok(data) => {
+                self.update_pending_writes(routing_node,
+                                           create_pending_mutation_for_mdata(mutation_type, data),
+                                           src,
+                                           dst,
+                                           msg_id,
+                                           false)
```
afck: Shouldn't this `rejected` be `true` if we believe that the change is not permitted?

src/personas/data_manager/mod.rs
```diff
@@ -0,0 +1,1329 @@
...
+                                                                  key.clone(),
+                                                                  msg_id)?;
+                            }
+                        }
+
+                        break;
+                    }
+                }
+            }
+        }
```
afck: Won't this now request all needed fragments at once?
That doesn't work for a vault joining a group which holds gigabytes of data. In master, we currently have only one ongoing request per peer at a time.

src/personas/data_manager/mod.rs
```diff
@@ -0,0 +1,1329 @@
...
+        result
+    }
+
+    // Get all fragments we are responsible for (irrespective of whether we already
+    // have them or not).
+    fn our_fragments(&self) -> HashSet<FragmentInfo> {
+        self.cache
+            .needed_fragments()
+            .union(&self.chunk_store_fragments())
+            .cloned()
```
afck: Can those two lines be switched? I imagine that would avoid cloning all the chunk store fragments.

src/vault.rs
```diff
@@ -203,70 +151,395 @@ impl Vault {
...
+            // ========== ListMDataValues ==========
+            (Authority::Client { .. },
+             Authority::NaeManager(_),
+             Request::ListMDataValues { name, tag, msg_id }) |
+            (Authority::ManagedNode(_),
+             Authority::ManagedNode(_),
+             Request::ListMDataValues { name, tag, msg_id }) => {
+                self.data_manager
+                    .handle_list_mdata_values(&mut self.routing_node, src, dst, name, tag, msg_id)
+            }
```
afck: I think we need to visibly document somewhere that for a busy chunk of mutable data (with several entry mutations going on in parallel at any time), the data manager could be unable to accumulate a response to this request.

It would be even better to help it accumulate one, of course, but I don't know how. Or maybe it's not actually an issue in practice, but we'll need to test that…

src/personas/maid_manager/mod.rs
afck: I think that's mainly because experiments have shown that the simple versioning scheme works fine for maid manager in practice, and refresh messages fix any missed updates. We might need to reconsider this as accounts get more complex.

tests/data_manager.rs
```diff
@@ -642,337 +650,174 @@ fn appendable_data_parallel_post() {
...
+            } else {
+                // Mutate existing data.
+                let j = rng.gen_range(0, all_data.len());
+                let data = &mut all_data[j];
+
+                if !mutated_data.insert((*data.name(), data.tag())) {
+                    trace!("Skipping data with name {:?}, tag {:?}.",
+                           data.name(),
+                           data.tag());
+                    continue;
```
afck: Once parallel mutations of different entries in the same chunk are fully handled, `mutated_data` should probably contain name, tag _and_ key.

tests/maid_manager.rs
```diff
@@ -18,354 +18,401 @@
...
             data_list.clear();
             for data in (0..chunks_per_iter).map(|_| {
-                Data::Structured(test_utils::random_structured_data(100000, &full_id, &mut rng))
-            }) {
-                client.put(data.clone());
+                                                     test_utils::gen_mutable_data(TEST_TAG,
+                                                                                  10,
+                                                                                  client_key,
+                                                                                  &mut rng)
+                                                 }) {
```
afck: That would probably be formatted more nicely if it were just a simple loop:
```rust
for _ in 0..chunks_per_iter {
    let data = test_utils::gen_mutable_data(…);
```

src/utils.rs
```diff
@@ -15,14 +15,79 @@
...
+pub type HashMap<K, V> = collections::HashMap<K, V, BuildHasherDefault<DefaultHasher>>;
+#[cfg(not(feature = "use-mock-crust"))]
+pub type HashMap<K, V> = collections::HashMap<K, V>;
+
+#[cfg(feature = "use-mock-crust")]
+pub type HashSet<T> = collections::HashSet<T, BuildHasherDefault<DefaultHasher>>;
+#[cfg(not(feature = "use-mock-crust"))]
+pub type HashSet<T> = collections::HashSet<T>;
+
+/// Extract client key (a public singing key) from the authority.
```
ustulation: :D

src/personas/data_manager/mod.rs
madadam: Does it mean that any mutation must be always written to the pending writes (with `rejected = true`), even if it fails (for example validation failure, version mismatch, etc...)?

afck: Yes, I think the general rules should be:
* If it's valid relative to your local state and you haven't voted for anything conflicting yet, vote for it, set `rejected = false` and put it in the pending writes.
* Otherwise if it isn't inherently invalid, i.e. if there is a chance that the group outvotes you and accepts the change, set `rejected = true`, put it in the pending writes and send an error response. (If the group outvotes you, the error response won't accumulate anyway.)
* If there is no chance that you will ever write it to your chunk store, e.g. because it is inherently invalid or its version number is older than your local one, drop it and send an error response.

src/personas/data_manager/cache.rs
```diff
@@ -0,0 +1,679 @@
...
+        // Insert the same fragment but with different holder. It should be present
+        // in the collection only one, but with both holders.
+        assert!(cache.insert_needed_fragment(fragment0.clone(), holder1));
+        let result = cache.unrequested_needed_fragments();
+        assert_eq!(result.len(), 1);
+        let item = first(result);
+        assert_eq!(item.0, fragment0);
+        assert_eq!(item.1.len(), 2);
+        assert!(item.1.contains(&holder0));
+        assert!(item.1.contains(&holder1));
```
madadam: No, because the order of holders is not guaranteed.

src/personas/data_manager/tests.rs
```diff
@@ -0,0 +1,723 @@
...
+    let data0 = test_utils::gen_immutable_data(10, &mut rng);
+    let data1 = test_utils::gen_mutable_data(TEST_TAG, 2, client_key, &mut rng);
+
+    dm.put_into_chunk_store(data0.clone());
+    dm.put_into_chunk_store(data1.clone());
+
+    // Node receives NodeAdded event. It should send all the fragments that it has
+    // to the new node in a Refresh message.
+    let rt = node.routing_table().clone();
+    dm.handle_node_added(&mut node, &new_node_name, &rt);
```
madadam: Sounds good, but should we also remove `RoutingTable` from `Event::NodeAdded` / `Event::NodeLost`?

src/personas/data_manager/cache.rs
afck: Would it make sense to make that a set instead of a vector then?
(Doesn't have to be in this PR, of course, if it's a big change.)

src/personas/data_manager/tests.rs
afck: Yes, definitely. Feel free to add that change to the mutable-data branch in routing. (Or later, if you prefer to do it as a separate change.)

src/personas/maid_manager/account.rs
```diff
@@ -0,0 +1,126 @@
...
+use routing::ClientError;
+use rust_sodium::crypto::sign;
+use std::collections::BTreeSet;
+
+/// Default available number of mutations per account.
+#[cfg(not(feature = "use-mock-crust"))]
+pub const DEFAULT_ACCOUNT_SIZE: u64 = 500;
+/// Default available number of mutations per account.
```
madadam: Just tested it. The conditional compilation seems to apply to the doc comments too, so it's good the way it is.

src/personas/maid_manager/tests.rs
```diff
@@ -0,0 +1,740 @@
...
+                                         data_name,
+                                         TEST_TAG,
+                                         new_owners.clone(),
+                                         5,
+                                         msg_id));
+    let message = unwrap!(node.sent_responses.remove(&msg_id));
+    assert_match!(message.response,
+                  Response::ChangeMDataOwner { res: Err(ClientError::AccessDenied), .. });
+
+    // Attempt to change owner even by authorised app fails.
```
madadam: Yes, I'll modify the comments to make it clearer.

src/utils.rs
```diff
@@ -15,14 +15,79 @@
...
 
-pub fn client_name(authority: &Authority<XorName>) -> XorName {
+#[derive(Clone, Copy, Debug, Default, Eq, Ord, PartialEq, PartialOrd, Hash, Serialize, Deserialize)]
+pub struct SecureHash(sha3::Digest256);
+
+/// Compute secure hash of the given value.
+pub fn secure_hash<T: Serialize>(value: &T) -> SecureHash {
+    serialisation::serialise(value)
+        .map(|data| SecureHash(tiny_keccak::sha3_256(&data)))
+        .unwrap_or_else(|_| Default::default())
```
madadam: Yep. Even better would be to return `Result<SecureHash, _>` but not sure it's worth it. Under what circumstances can serialisation actually fail?

src/utils.rs
```diff
@@ -15,14 +15,79 @@
...
+}
+
+// Note: for testing with mock crust, use collections with deterministic hashing.
+
+#[cfg(feature = "use-mock-crust")]
+pub type HashMap<K, V> = collections::HashMap<K, V, BuildHasherDefault<DefaultHasher>>;
```
madadam: As far as I understand, it should produce the same hashes across multiple executions, which is what we care about the most, because we want to have reproducible test runs. Not sure about different architectures though. Would we need that?

afck: Not sure. At least if it differed between 32 and 64 bit Windows, that would make it cumbersome to reproduce a failure found on AppVeyor locally.

src/mock_crust_detail/test_client.rs
```diff
@@ -92,7 +92,7 @@ impl TestClient {
         self.routing_client.try_next_ev()
     }
 
-    /// empty this client event loop
+    /// Empties this clients event loop
```
afck: `client's` (also below)

src/utils.rs
afck: But we still have several other sources of randomness in the safe_vault tests anyway, so I'm happy to postpone the deterministic tests issue to a later PR!

src/personas/data_manager/cache.rs
madadam: I guess there is no strong reason why it should be anything else than `Vec`, since the only thing we do with it in the production code is iteration. It would make this test a bit nicer, but not sure it's worth it.

src/personas/data_manager/cache.rs
```diff
@@ -52,200 +54,176 @@ pub struct Cache {
...
-        // Reset expired requests
-        for request in self.needed_fragments
-                .iter_mut()
-                .flat_map(|(_, fragments)| fragments.values_mut()) {
-            request.stop_if_expired();
-        }
+        self.remove_expired_requests();
 
-        let unneeded: HashSet<_> = self.unneeded_chunks
+        let unneeded_names: HashSet<_> = self.unneeded_chunks
```
afck: I think we should just never insert unneeded fragments into `self.needed_fragments`, then we wouldn't need to filter here (and below).
If we don't insert anything in `self.needed_fragments` that is in our chunk store, that should already be the case.

src/personas/data_manager/cache.rs
```diff
@@ -52,200 +54,176 @@ pub struct Cache {
...
-                        continue;
-                    }
-                }
+        let mut result = Vec::new();
 
-                result
-                    .entry(fragment.clone())
-                    .or_insert_with(Vec::new)
-                    .push(*holder);
+        for (fragment, state) in &self.needed_fragments {
```
afck: I think the needed fragments should still (like in master) be sorted by holder. That way we could avoid iterating through all of them here, and just for every holder pick the next fragment in line.
(If a group is storing 10000 mutable data chunks, this could be a million entries!)

src/personas/data_manager/mod.rs
```diff
@@ -1163,15 +1151,23 @@ impl DataManager {
         let src = Authority::ManagedNode(routing_node.name()?);
         let candidates = self.cache.unrequested_needed_fragments();
 
+        // Set of holders we already sent a request to. Used to prevent sending
+        // multiple request to a single holder.
```
afck: `requests`

tests/data_manager.rs
```diff
@@ -90,15 +89,20 @@ fn immutable_data_operations_with_churn_without_cache() {
 }
 
 fn immutable_data_operations_with_churn(use_cache: bool) {
-    let network = Network::new(GROUP_SIZE, None);
-    let mut rng = network.new_rng();
+    extern crate maidsafe_utilities;
+    unwrap!(maidsafe_utilities::log::init(false));
```
afck: That's probably left over from debugging.

src/personas/data_manager/mod.rs
```diff
@@ -0,0 +1,1310 @@
...
+        self.chunk_store.used_space() > (self.chunk_store.max_space() / 100) * MAX_FULL_PERCENT
+    }
+
+    /// Removes data chunks we are no longer responsible for until the chunk store is not full
+    /// anymore.
+    fn clean_chunk_store(&mut self) {
+        while self.chunk_store_full() {
+            if let Some(data_id) = self.cache.pop_unneeded_chunk() {
+                if let Err(error) = self.chunk_store.delete(&data_id) {
+                    error!("DM failed to delete unneeded chunk {:?}: {:?}",
```
maqi: shall this just be considered as a `warning`?

src/personas/data_manager/mod.rs
```diff
@@ -0,0 +1,1310 @@
...
+        for data_id in self.chunk_store.keys() {
+            match data_id {
+                DataId::Immutable(data_id) => {
+                    let _ = result.insert(FragmentInfo::ImmutableData(*data_id.name()));
+                }
+                DataId::Mutable(data_id) => {
+                    let data = if let Ok(data) = self.chunk_store.get(&data_id) {
+                        data
+                    } else {
+                        error!("Failed to get {:?} from chunk store.", data_id);
```
maqi: we just fetched the key but failed to get the chunk, shall we do something here?
say try a re-fetch or clean-up the key?

also, I think this shall be considered as a failure during mock_crust tests (not sure whether it will happen)

src/personas/data_manager/mod.rs
```diff
@@ -0,0 +1,1310 @@
...
+
+        if self.cache.prune_needed_fragments(routing_table) {
+            self.request_needed_fragments(routing_node)?;
+        }
+
+        let mut refreshes = HashMap::default();
+        for fragment in self.our_fragments() {
+            match routing_table.other_closest_names(fragment.name(), GROUP_SIZE) {
+                None => {
+                    error!("Moved out of close group of {:?} in a NodeLost event.",
```
maqi: shall we panic here during the mock_crust test as this indicates a failure in the RT or event fired from routing?

src/personas/data_manager/mod.rs
```diff
@@ -0,0 +1,1310 @@
...
+            }
+            Err(error) => {
+                self.send_mutation_response(routing_node,
+                                            src,
+                                            dst,
+                                            mutation_type,
+                                            DataId::mutable(name, tag),
+                                            Err(error),
+                                            msg_id)
+            }
```
maqi: just wondering even if the mutation is rejected from our perspective, do we still need to put it into the pending list,
so when group decides to mutate later on, we can carry out writes anyway

src/personas/data_manager/mod.rs
```diff
@@ -0,0 +1,1310 @@
...
+                }
+            }
+        }
+
+        self.request_needed_fragments(routing_node)
+    }
+
+    // When a node in a group receives request to mutate some data it holds, it first sends
+    // "group refresh" message to all the other members of the group. Only when the message
+    // accumulates, the node applies the mutation to the data and updates it in the chunk
```
maqi: loos like the entry also MUST exists in the pending_writes as well?

src/personas/data_manager/mod.rs
```diff
@@ -0,0 +1,1310 @@
...
+    fn our_fragments(&self) -> HashSet<FragmentInfo> {
+        let mut result = self.cache.needed_fragments();
+        result.extend(self.chunk_store_fragments());
+        result
+    }
+}
+
+#[cfg(feature = "use-mock-crust")]
+impl DataManager {
+    pub fn get_stored_ids(&self) -> Vec<(DataId, u64)> {
```
maqi: i think it is actually idv being returned?

src/personas/data_manager/mod.rs
```diff
@@ -0,0 +1,1310 @@
...
+        self.chunk_store
+            .keys()
+            .into_iter()
+            .filter(|data_id| match *data_id {
+                        DataId::Immutable(ref id) => !self.cache.is_in_unneeded(id),
+                        DataId::Mutable(_) => true,
+                    })
+            .filter_map(|data_id| {
+                            self.get_version(&data_id)
```
maqi: `get_version` could return None for mutable data (when failed to fetch it from chunk_store) and here it will then be filtered out.
will this be expected?

src/personas/maid_manager/mod.rs
```diff
@@ -0,0 +1,717 @@
...
+        routing_node
+            .send_mutate_mdata_entries_response(dst, src, res, msg_id)?;
+        Ok(())
+    }
+
+    #[cfg_attr(feature = "cargo-clippy", allow(too_many_arguments))]
+    pub fn handle_set_mdata_user_permissions(&mut self,
+                                             routing_node: &mut RoutingNode,
+                                             src: Authority<XorName>,
+                                             dst: Authority<XorName>,
```
maqi: can we have a struct to hold most of the parameters, and use such struct in functions like: handle_set_mdata_user_permissions, handle_del_mdata_user_permissions, handle_change_mdata_owner ?

src/personas/maid_manager/mod.rs
```diff
@@ -0,0 +1,717 @@
...
+        Ok(())
+    }
+
+    pub fn handle_list_auth_keys_and_version(&mut self,
+                                             routing_node: &mut RoutingNode,
+                                             src: Authority<XorName>,
+                                             dst: Authority<XorName>,
+                                             msg_id: MessageId)
+                                             -> Result<(), InternalError> {
+        let res = self.get_account(&dst)
```
maqi: do we need to check the requester?

src/personas/maid_manager/mod.rs
```diff
@@ -0,0 +1,717 @@
...
+    pub fn handle_ins_auth_key(&mut self,
+                               routing_node: &mut RoutingNode,
+                               src: Authority<XorName>,
+                               dst: Authority<XorName>,
+                               key: sign::PublicKey,
+                               version: u64,
+                               msg_id: MessageId)
+                               -> Result<(), InternalError> {
+        let res = self.mutate_account(routing_node, &src, &dst, version, |account| {
+            let _ = account.auth_keys.insert(key);
```
maqi: a duplication of key will be counted as a success, right?

src/personas/maid_manager/mod.rs
```diff
@@ -0,0 +1,717 @@
...
+        // Remove all requests from the cache that we are no longer responsible for.
+        let msg_ids_to_delete = self.request_cache
+            .iter()
+            .filter(|&(_, entry)| accounts_to_delete.contains(&entry.src.name()))
+            .map(|(msg_id, _)| *msg_id)
+            .collect_vec();
+        for msg_id in msg_ids_to_delete {
+            let _ = self.request_cache.remove(&msg_id);
+        }
+        if !accounts_to_delete.is_empty() {
```
maqi: why not put this log after the removal?

src/mock_crust_detail/test_client.rs
```diff
@@ -95,309 +123,388 @@ impl TestClient {
...
+        let request_msg_id = self.put_idata(data);
+        let _ = poll::poll_and_resend_unacknowledged(nodes, self);
+
+        match self.try_recv() {
+            Ok(Event::Response { response: Response::PutIData { res, msg_id }, .. }) => {
+                assert_eq!(request_msg_id, msg_id);
+                res
+            }
+            Ok(response) => panic!("Unexpected response: {:?}", response),
+            Err(error) => Err(ClientError::from(format!("{:?}", error))),
```
maqi: shall we panic here?

src/mock_crust_detail/test_client.rs
```diff
@@ -17,54 +17,82 @@
...
+            Ok(event) => panic!("Unexpected event: {:?}", event),
+            Err(error) => panic!("Unexpected error: {:?}", error),
+        }
+    }
+}
+
 /// Client for use in tests only
 pub struct TestClient {
     _handle: ServiceHandle,
     routing_client: routing::Client,
```
maqi: can we use routing::Client directly now? 
so here only need to be `routing_client: Client`

src/mock_crust_detail/test_client.rs
```diff
@@ -95,309 +123,388 @@ impl TestClient {
...
-                    return Err(Some(parsed_error));
-                }
-                Ok(response) => panic!("Unexpected Post response : {:?}", response),
-                Err(err) => panic!("Unexpected error : {:?}", err),
-            }
-        }
+                    .get_mdata_shell(dst, name, tag, msg_id));
+        let _ = poll::nodes_and_client(nodes, self);
+
+        assert_recv_response!(self, GetMDataShell, msg_id)
```
maqi: looks like the previous tests allows a failure response to be returned and pass back such failure to the testing code to verify (for the testing purpose of `failing scenarios`)
however, the new `assertion` seems won't catering for the `testing of failing scenarios` any more.
is this expected?

src/personas/data_manager/mod.rs
```diff
@@ -0,0 +1,1310 @@
...
+                             message_id: MessageId,
+                             rejected: bool)
+                             -> Result<(), InternalError> {
+        for PendingWrite {
+                mutation,
+                src,
+                dst,
+                message_id,
+                ..
+            } in self.cache.remove_expired_writes() {
```
maqi: only remove_expired when there is a new pending_write update, which I think might cause issue in the following scenario:
assuming MM(client) is in group A and DM(mut_data) is in group B, and group A and B are separated (no node in both)
client then sends out two concurrent mutations, which cause its account in MMs having mutation count increased by 2.
However, if the two mutation requests arrive group B at the same time, and gets a 50-50 votes, it means there is no accumulation achieved, and the records get stalled in the pending writes.
Meanwhile, there is nodes keeps got added into group A. the new MMs to client only got account through refresh, but its cache_request has no knowledge of the mutation requests.
When group B got a new pending writes update, they sends two mutation_failure_responses.
However, those new MMs will ignore the response, but the old MMs will update account and refresh.
This leaves a chance of breaking the accumulation if the quorum cannot be reached.

src/personas/data_manager/mod.rs
madadam: It does. Not sure it's necessary to mention it in the comment. The comment is just a high-level description of what the purpose of the method is.

src/personas/data_manager/mod.rs
madadam: Would it be better to change the method to return `Result` and propagate the error to `get_stored_ids` too?

src/personas/maid_manager/mod.rs
```diff
@@ -0,0 +1,717 @@
...
+        Ok(())
+    }
+
+    pub fn handle_get_account_info(&mut self,
+                                   routing_node: &mut RoutingNode,
+                                   src: Authority<XorName>,
+                                   dst: Authority<XorName>,
+                                   msg_id: MessageId)
+                                   -> Result<(), InternalError> {
+        let res = self.get_account(&dst).map(|account| account.info);
```
madadam: Good question. Not sure, @afck, @ustulation - what do you guys think?

src/personas/maid_manager/mod.rs
```diff
@@ -0,0 +1,717 @@
...
+            routing_node
+                .send_put_mdata_response(dst, src, Err(ClientError::InvalidOwners), msg_id)?;
+            return Ok(());
+        }
+
+        // If the type_tag is `TYPE_TAG_SESSION_PACKET`, the account must not exist, else it must
+        // exist.
+        if data.tag() == TYPE_TAG_SESSION_PACKET {
+            if dst_name != src_name {
+                trace!("Cannot create account for {:?} as {:?}.", src, dst);
```
madadam: In what sense is it misleading? What alternative would you suggest?

src/personas/maid_manager/mod.rs
afck: I don't think it would be a good idea to let anyone query an account. The account balance is kind of private?

src/personas/maid_manager/mod.rs
madadam: We can, but not sure what the point of doing that would be?

src/personas/maid_manager/mod.rs
madadam: See the discussion here: https://github.com/maidsafe/safe_vault/pull/605#discussion_r115425043

src/personas/maid_manager/mod.rs
maqi: first clippy complains too many parameters, second it can de-duplicate some code.

src/personas/maid_manager/mod.rs
madadam: The keys are stored in `HashSet`, so inserting a key that already exists has no effect. I don't think it's necessary to respond with error here, but I don't have strong opinion either way.

src/personas/maid_manager/mod.rs
maqi: trace!("Requester {:?} is not a valid owner for account  {:?}.", src, dst);    ?

src/personas/data_manager/mod.rs
maqi: yes, probably we should

src/mock_crust_detail/test_client.rs
madadam: possibly, but in master, there is this comment: https://github.com/maidsafe/safe_vault/blob/master/src/mock_crust_detail/test_client.rs#L388 so not sure. I'll try to panic and see what happens. If it doesn't work, I'll put the comment back.

src/mock_crust_detail/test_client.rs
madadam: I think it's fine. `assert_recv_response!` returns both successes and failures. It only panics if the response is not of the expected type, or if there is no response, which is fine in tests, because those cases are always unexpected.

src/mock_crust_detail/test_node.rs
```diff
@@ -141,22 +143,27 @@ pub fn create_nodes(network: &Network,
     let crust_config = mock_crust::Config::with_contacts(&[nodes[0].endpoint()]);
 
     // Create other nodes using the seed node endpoint as bootstrap contact.
-    for _ in 1..size {
+    for _ in 1..size - 1 {
```
maqi: can we have
for _ in 1..size {
...
}
drop(config);
?

src/personas/data_manager/mod.rs
```diff
@@ -0,0 +1,1440 @@
...
+                             message_id: MessageId,
+                             rejected: bool)
+                             -> Result<(), InternalError> {
+        for PendingWrite {
+                mutation,
+                src,
+                dst,
+                message_id,
+                ..
+            } in self.cache.remove_expired_pending_writes() {
```
maqi: only remove_expired when there is a new pending_write update, which I think might cause issue in the following scenario:
assuming MM(client) is in group A and DM(mut_data) is in group B, and group A and B are separated (no node in both)
client then sends out two concurrent mutations, which cause its account in MMs having mutation count increased by 2.
However, if the two mutation requests arrive group B at the same time, and gets a 50-50 votes, it means there is no accumulation achieved, and the records get stalled in the pending writes.
Meanwhile, there is nodes keeps got added into group A. the new MMs to client only got account through refresh, but its cache_request has no knowledge of the mutation requests.
When group B got a new pending writes update, they sends two mutation_failure_responses.
However, those new MMs will ignore the response, but the old MMs will update account and refresh.
This leaves a chance of breaking the accumulation if the quorum cannot be reached.

src/personas/data_manager/mod.rs
```diff
@@ -0,0 +1,1440 @@
...
+            return Ok(());
+        };
+
+        // If we're no longer in the close group, return.
+        if !close_to_address(routing_node, &name) {
+            return Ok(());
+        }
+
+        let data_id = MutableDataId(name, tag);
+        match self.chunk_store.get(&data_id) {
```
maqi: there is separate function of `fetch_mdata`,  then what's the purpose of having it (or not to use it here)?
and same to the other places calling self.chunk_store.get for mutable data.

src/personas/data_manager/mod.rs
```diff
@@ -0,0 +1,1440 @@
...
+
+        if self.cache.prune_needed_fragments(routing_table) {
+            self.request_needed_fragments(routing_node)?;
+        }
+
+        let mut refreshes = HashMap::default();
+        for data_id in self.our_chunks() {
+            match routing_table.other_closest_names(data_id.name(), GROUP_SIZE) {
+                None => {
+                    error!("Moved out of close group of {:?} in a NodeLost event.",
```
maqi: shall we panic here during the mock_crust test as this indicates a failure in the RT? 
or it won't happen anyway?

tests/data_manager.rs
```diff
@@ -642,337 +650,174 @@ fn appendable_data_parallel_post() {
...
             }
-            trace!("No response received for client {:?} in iteration {:?}.",
+            panic!("Client {:?} received no response for data with name {:?}, tag {}.",
                    client.name(),
-                   i + 1);
+                   data.name(),
+                   data.tag());
         }
+
+        assert!(successes > 0, "No MutateMDataEntry attempt succeeded.");
```
maqi: the new concurrent tests verifies the conflict situation and adjusts `success` expectation on that

src/personas/data_manager/cache.rs
```diff
@@ -0,0 +1,679 @@
...
+        tag: u64,
+        version: u64,
+        hash: SecureHash,
+    },
+    MutableDataEntry {
+        name: XorName,
+        tag: u64,
+        key: Vec<u8>,
+        version: u64,
+        hash: SecureHash,
```
madadam: We don't send `FragmentInfo`s in refreshes send to newly joined nodes anymore, so this should not be an issue. `FragmentInfo`s are only send on refreshes following mutations: https://github.com/madadam/safe_vault/blob/mutable-data/src/personas/data_manager/mod.rs#L208 and then there is an upper limit on their size (because there is a limit on the number of entry action in a single mutation request).

tests/data_manager.rs
maqi: when there is concurrent mutations to the same key, there response from network will be undecided.
Hence making this test more complicated.
I will suggest leave this test as is, as the concurrent test covers the mutations to keys

src/personas/data_manager/mod.rs
maqi: required here as need to update (key,value).
however, there is another place better use fetch_mdata instead of calling chunk_store.get, and has been updated.

src/personas/data_manager/tests.rs
```diff
@@ -0,0 +1,723 @@
...
+    unwrap!(dm.handle_put_idata(&mut node, client_manager, nae_manager, data.clone(), msg_id));
+
+    let message = unwrap!(node.sent_requests.remove(&msg_id));
+    let refresh = assert_match!(message.request, Request::Refresh(payload, _) => payload);
+    assert_eq!(message.src, nae_manager);
+    assert_eq!(message.dst, nae_manager);
+
+    // Simulate receiving the refresh. This should result in the data being
+    // put into the chunk store.
+    unwrap!(dm.handle_group_refresh(&mut node, &refresh));
```
maqi: now asserted

tests/network.rs
```diff
@@ -26,6 +26,12 @@ use safe_vault::{Config, GROUP_SIZE, test_utils};
 use safe_vault::mock_crust_detail::{poll, test_node};
 use safe_vault::mock_crust_detail::test_client::TestClient;
 
+// Keeps storing data till network is full. Then keeps adding nodes till network can store a new
+// chunk again.
+// Among the GROUP_SIZE vauls of a chunk, the response to the client can be:
+// 1, Put succeed when majority of vaults are able to store the data.
+// 2, Put failed (NetworkFull) when majority of vaults don't have space to store the data.
+// 3, No response, when part of vaults have space but part of vaults don't, and none accumulates.
```
afck: Why can point 3 happen now? Why doesn't the strategy described in the removed comment below work anymore?

tests/data_manager.rs
```diff
@@ -689,6 +690,266 @@ fn mutable_data_parallel_mutations() {
...
+
+                if !sent_actions.is_empty() &&
+                   sent_actions
+                       .iter()
+                       .any(|(_, prev_actions)| {
+                                prev_actions
+                                    .iter()
+                                    .any(|(key, _)| actions.contains_key(key))
+                            }) {
+                    expected_successes = 1;
```
afck: As discussed, we probably can't expect one of two conflicting changes to succeed, since the vaults may be split 50-50. (The comment to this function also needs to be updated.)

tests/data_manager.rs
```diff
@@ -689,6 +690,266 @@ fn mutable_data_parallel_mutations() {
...
+
+        let mut successes: usize = 0;
+
+        while let Ok(event) = client.try_recv() {
+            if let Event::Response {
+                       response: Response::MutateMDataEntries { res, msg_id }, ..
+                   } = event {
+                match res {
+                    Ok(()) => {
+                        trace!("Client {:?} received successful response.",
```
afck: `success response` (also below)

tests/network.rs
maqi: that removed comment is actually an old comment about the case that `if say we have 20 nodes, keeps putting till full will have situation of 50-50 vote, and using 8 nodes can prevent this during the full-up procedure`
However, during the add node then poll procedure, the 50-50 vote situation may still happen.
But the old put_and_verify method has been updated and no longer hangs on no-response (will return with `Err(None)`).

i.e. point 3 will happen in master as well.

src/mock_crust_detail/test_node.rs
```diff
@@ -142,21 +145,23 @@ pub fn create_nodes(network: &Network,
 
     // Create other nodes using the seed node endpoint as bootstrap contact.
     for _ in 1..size {
+        // (2nd to N-1th node clone the config objects.)
```
afck: This should now say `Nth` again.

src/personas/maid_manager/mod.rs
madadam: JIRA tak: https://maidsafe.atlassian.net/browse/MAID-2068

src/personas/data_manager/mod.rs
maqi: This issue could happen in master as well.
raised https://maidsafe.atlassian.net/browse/MAID-2069 and this shall be covered separately later on

src/vault.rs
afck: As discussed, this will be added to the documentation as part of https://github.com/maidsafe/routing/pull/1450#discussion_r117943721.

src/utils.rs
```diff
@@ -15,14 +15,79 @@
...
+/// Extract client name (a `XorName`) from the authority.
+///
+/// # Panics
+///
+/// Panics when the authority is not `Client` or `ClientManager`.
+pub fn client_name(authority: &Authority<XorName>) -> XorName {
+    match *authority {
+        Authority::Client { ref client_key, .. } => client_name_from_key(client_key),
+        Authority::ClientManager(name) => name,
+        _ => unreachable!("Logic error"),
```
madadam: It would be best if the methods that accept `Authority` accepted a type that represents a particular variant of it (e.g. `struct ClientAuthority { ... }`), then that `unreachable!` branch would be actually statically unreachable. I'd postpone such change for the future, as it doesn't seem critical.

ustulation:
@madadam Do not use MsgId-zero, maybe just use the msg-id from the msg itself?

PR #612, opened 2017-03-29
style/rustfmt: adhere to rustfmt 0.8.1

afck:


maidsafe-highfive:
r? @maqi

PR #619, opened 2017-06-06
feature/maid-manager: implement invites

madadam:


maidsafe-highfive:
r? @Fraser999

src/personas/maid_manager/mod.rs
```diff
@@ -573,19 +590,126 @@ impl MaidManager {
...
+        }
+
+        match data.tag() {
+            TYPE_TAG_SESSION_PACKET => {
+                if dst_name != src_name {
+                    trace!("MM Cannot create account for {:?} as {:?}.", src, dst);
+                    return Err(ClientError::InvalidOperation);
+                }
+
+                if is_admin || self.invite_key.is_none() {
```
maqi: looks like the None case of self.invite_key is only for the use_mock_routing tests.
if so, can we use conditional compiler instead, and make self.invite_key not an Option?

Also, in the MM persona unit test (use-mock-routing), shall we include test case of  involving that invite_key ?

tests/maid_manager.rs
```diff
@@ -255,6 +257,70 @@ fn create_account_twice() {
 }
 
 #[test]
```
maqi: do we also need a test for the scenario of two clients currently claiming the same invite_key?

src/personas/maid_manager/mod.rs
```diff
@@ -293,8 +280,38 @@ impl MaidManager {
...
-        let CachedRequest { src, dst, .. } =
-            self.handle_data_mutation_response(routing_node, msg_id, res.is_ok())?;
+        let (src, dst, res) = if let Some(CachedAccountCreation { src, dst, data }) =
+            self.account_creation_cache.remove(&msg_id) {
+
+            let res = match res {
+                Ok(()) => {
+                    let (client_name, client_key) = utils::client_name_and_key(&src);
+                    let _ = self.accounts.insert(*client_name, Account::default());
+                    return self.forward_put_mdata(routing_node,
```
maqi: if this put_mdata fails from DM, the failure response will make MM remove that client's account.
however the invite_key has been used.

tests/maid_manager.rs
madadam: Did you mean **con**currently?

maqi: yes, sry, updated

src/mock_crust_detail/test_client.rs
```diff
@@ -131,6 +138,44 @@ impl TestClient {
...
+                                   entry_version: 0,
+                               });
+
+        let data = unwrap!(MutableData::new(self.rng.gen(),
+                                            TYPE_TAG_SESSION_PACKET,
+                                            Default::default(),
+                                            entries,
+                                            owners));
+
+        let request_msg_id = self.put_mdata(data);
```
madadam: No, because there are two valid response types: `PutMData` and `MutateMDataEntries`. `put_mdata_response` expects only `PutMData` and would panic on other response types.

src/personas/maid_manager/mod.rs
madadam: Valid point. I guess what we can do is un-claim the invite in that case? Problem is that is currently tricky. The claiming is currently implemented as inserting a new entry (with specific key and value) into the invite mdata. un-claiming would then have to be implemented as deleting the entry, but entry deletion is not implemented yet. Alternative would be to use update instead of insert and have specific values representing not-claimed / claimed invites, but that is also tricky because of entry versions. Is it worth doing?

src/personas/maid_manager/mod.rs
madadam: I'm not sure about it. Vault master also has this as `Option` and it does not have mock routing. It needs to be confirmed.

Yes, unit-test for the invite flow would be useful, but not sure it's necessary.

src/mock_crust_detail/test_client.rs
maqi: you the client could receive any of two response types regarding the same request?

bit strange. but maybe in a separate PR the response shall be made into one type.

src/personas/maid_manager/mod.rs
maqi: I think not to carry out un-claiming is acceptable, as long as we are aware of this limitation and chance of such failure case happen is tiny.

src/personas/maid_manager/mod.rs
```diff
@@ -573,19 +590,126 @@ impl MaidManager {
...
+                           src_name,
+                           invite_name);
+
+                    let item = CachedAccountCreation {
+                        src: src,
+                        dst: dst,
+                        data: data,
+                    };
+                    if let Some(old) = self.account_creation_cache.insert(msg_id, item) {
+                        debug!("Received two account creation requests with message ID {:?}. {:?} \
```
madadam: As long as clients don't send two or more messages with the same `msg_id`, I don't think it can happen. That said, clients sending multiple messages with the same msg_id is something we should look into to check whether it can cause any problems.

tests/maid_manager.rs
```diff
@@ -255,6 +257,70 @@ fn create_account_twice() {
...
+fn invite() {
```
madadam: Will do.

src/personas/maid_manager/mod.rs
madadam: According to @nbaksalyar , it's useful to allow `None` even in production. So I'm keeping it.

src/personas/maid_manager/mod.rs
madadam: Actually, thinking about it, this is a perfect use case for the transaction feature we have planned. What about postponing this until we have that?

src/mock_crust_detail/test_client.rs
```diff
@@ -131,6 +139,35 @@ impl TestClient {
...
+                                          -> Result<(), ClientError> {
+        let owner = *self.signing_public_key();
+        let owners = iter::once(owner).collect();
+
+        let content = AccountPacket::WithInvitation {
+            invitation_string: invitation_code.to_owned(),
+            acc_pkt: Vec::new(),
+        };
+        let content = unwrap!(serialisation::serialise(&content));
+        let mut entries = BTreeMap::new();
```
afck: This could be immutable if it were initialised like `owners` above.

src/personas/maid_manager/mod.rs
```diff
@@ -573,19 +590,126 @@ impl MaidManager {
...
+                    trace!("MM Cannot create account for {:?} as {:?}.", src, dst);
+                    return Err(ClientError::InvalidOperation);
+                }
+
+                if is_admin || self.invite_key.is_none() {
+                    let len = self.accounts.len();
+                    match self.accounts.entry(*src_name) {
+                        Entry::Vacant(entry) => {
+                            let _ = entry.insert(Account::default());
+                            info!("Managing {} client accounts.", len + 1);
```
afck: This does end up returning `PutMDataAction::Forward(data)`, does it? It's a bit hard to tell with all the nesting and indenting. It would be more readable if some of this was extracted into anothere method.

tests/maid_manager.rs
```diff
@@ -254,6 +256,75 @@ fn create_account_twice() {
...
+    let name = XorName(tiny_keccak::sha3_256(b"fake invite"));
+    let mut owners = BTreeSet::new();
+    let _ = owners.insert(*client1.signing_public_key());
+    let data = unwrap!(MutableData::new(name,
+                                        TYPE_TAG_INVITE,
+                                        permissions,
+                                        Default::default(),
+                                        owners));
+    assert_eq!(Err(ClientError::InvalidOperation),
+               client1.put_mdata_response(data, &mut nodes));
```
afck: Wouldn't that fail anyway because the client doesn't even have an account yet?
Maybe it should try (only, or again?) putting the fake invite after succesfully creating its own account.

madadam: I changed the order - account creation is now before the attempt to put the fake invite.

PR #620, opened 2017-06-07
fix/tests: increase iterations in fill_network test

madadam:


maidsafe-highfive:
r? @maqi

tests/network.rs
```diff
@@ -75,12 +78,18 @@ fn fill_network() {
...
 
-    for _ in 0..10 {
+    for i in 0..max_iterations {
         let index = Range::new(1, nodes.len()).ind_sample(&mut rng);
         trace!("Adding node with bootstrap node {}.", index);
         test_node::add_node(&network, &mut nodes, index, true);
         let _ = poll::nodes_and_client_with_resend(&mut nodes, &mut client);
 
+        // Because the original 8 nodes are all full, we need at least `QUORUM`
+        // new nodes before we have a chance of the network accepting the data.
```
maqi: could that happen, say 1..8 holds {A, B, C, D, E} being full, then 9 joins, and 1 shift A to 9, 2 shift B to 9, 3 shift C to 9, ...
results in 1..5 have one empty space to store a new data ?

anyway, I feel it's not that critical to `save` this couple of iterations.
if do want to, why not just after fill the network, join 10 more nodes in a bunch then starts iteration.
this will give a better chance of a success in just few steps.

src/mock_crust_detail/test_client.rs
```diff
@@ -163,6 +163,7 @@ impl TestClient {
                                   nodes: &mut [TestNode])
                                   -> Result<(), ClientError> {
         let request_msg_id = self.put_idata(data.clone());
+        // TODO: why do we use `*_with_resend` here, but not in `put_idata_response`?
```
maqi: yeah, seems `with_send` is not required.
you can just remove it and see what happens.

tests/network.rs
afck: It can, e.g. with node `00000000` joining nodes `10000000`, `01000000`, `00100000`, …, `00000001` who are storing data chunks `01111111`, `10111111`, …, `11111110`, every node would drop one chunk. (But if it's unlikely enough and this `continue` makes the test run faster on average, I'm fine with it, of course.)

madadam: Valid points, but it does not affect validity of the test and it should make it run a bit faster on average. I'm going to keep it in, unless someone strongly disagrees.

src/mock_crust_detail/test_client.rs
madadam: Tried to use the one without resend and got weird test failure - received response whose `msg_id` didn't match the request. Probably a message which didn't accumulate in the previous iteration, but accumulated in this? Not sure if it is a bug or expected behaviour.

PR #623, opened 2017-06-12
Improve test coverage

maqi:


maidsafe-highfive:
r? @Fraser999

src/mock_crust_detail/test_client.rs
```diff
@@ -139,11 +139,24 @@ impl TestClient {
...
     pub fn create_account_with_invitation(&mut self,
                                           invitation_code: &str,
                                           nodes: &mut [TestNode])
                                           -> Result<(), ClientError> {
+        let data = unwrap!(self.compose_account_data(invitation_code));
+        self.put_mdata_response(data, nodes)
+    }
+
+    /// Creates an account using the given invitation code, doesn't expect response.
+    pub fn create_account_with_invitation_no_response(&mut self,
```
madadam: I'd prefer to rename these methods like this:
- `create_account_with_invitation` - don't expect response
- `create_account_with_invitation_response` - expects response
to be consistent with how all the other operations in this module are named.

tests/data_manager.rs
```diff
@@ -101,7 +101,7 @@ fn immutable_data_operations_with_churn(use_cache: bool) {
     let mut rng = network.new_rng();
 
     let mut nodes = test_node::create_nodes(&network, node_count, None, use_cache);
-    let config = BootstrapConfig::with_contacts(&[nodes[0].endpoint()]);
+    let config = BootstrapConfig::with_contacts(&[nodes[1].endpoint()]);
```
madadam: What about picking the bootstrap node randomly? Would that make sense?

tests/data_manager.rs
```diff
@@ -120,14 +120,14 @@ fn immutable_data_operations_with_churn(use_cache: bool) {
         }
 
         if nodes.len() <= GROUP_SIZE + 2 || !rng.gen_weighted_bool(4) {
-            let index = rng.gen_range(1, nodes.len());
+            let index = rng.gen_range(2, nodes.len());
```
madadam: Same here - pick the bootstrap node index randomly. (also in all the other places where bootstrap node is mentioned)

tests/data_manager.rs
maqi: then the `churn` process needs to care for the case not to drop out the proxy node.
I will prefer leave it as current.

tests/data_manager.rs
maqi: same as the above comment.
prefer leave it as current.

tests/maid_manager.rs
```diff
@@ -322,7 +325,98 @@ fn invite() {
...
+
+    // Attempt to reuse already claimed invite fails.
+    let mut client3 = TestClient::new(&network, Some(config));
+    client3.ensure_connected(&mut nodes);
+
+    // May receive `Err(InvitationAlreadyClaimed)` or
+    // `Err(NetworkOther("Error claiming invitation: Conflicting concurrent mutation"))`
+    match client3.create_account_with_invitation_response(invite_code, &mut nodes) {
+        Ok(()) => panic!("re-claiming a used invitation shall not succeed."),
+        Err(err) => {
```
ustulation: It maybe a good idea to also assert the correct error is returned. So something like `Err(Error::Abc) => { ... }, Ok(()) => panic!(..), Err(e) => panic!("Expected Error::Abc got {:?}", e),`

PR #625, opened 2017-06-13
Prevent msg_id reuse, add use-mock-crypto feature, improve tests

madadam:


maidsafe-highfive:
r? @maqi

tests/maid_manager.rs
```diff
@@ -667,3 +667,73 @@ fn account_concurrent_insert_key_put_data() {
...
+
+    let mut successes = 0;
+    while let Ok(event) = client.try_recv() {
+        if let Event::Response { response: Response::PutIData { res: Ok(()), .. }, .. } = event {
+            successes += 1;
+        }
+    }
+
+    // Only one request does succeed.
+    assert_eq!(successes, 1);
```
maqi: could there be chance that receives no response from network? 
i.e. half passed data0 req to DM and half passed data1, hence no accumulation?

src/personas/maid_manager/mod.rs
```diff
@@ -137,19 +137,27 @@ impl MaidManager {
...
         }
 
         // Forwarding the request to NAE Manager.
-        let fwd_src = dst;
-        let fwd_dst = Authority::NaeManager(*data.name());
-        trace!("MM forwarding PutIData request to {:?}", fwd_dst);
-        routing_node
-            .send_put_idata_request(fwd_src, fwd_dst, data, msg_id)?;
-        self.insert_into_request_cache(msg_id, src, dst, None);
+        if self.insert_into_request_cache(msg_id, src, dst, None) {
```
maqi: function of insert_into_request_cache will replace the old one anyway.
so actually here we are not only invalidates the current request but also mismatch the response of old request from DM with the current request
I think here we shall carry out a `has` check then carry on the operations?

and same to the other places

madadam: Fixed. `insert_into_request_cache` won't overwrite previous entries anymore.

src/personas/maid_manager/mod.rs
```diff
@@ -1163,6 +1172,14 @@ enum PutMDataAction {
     Forward(MutableData),
 }
 
+struct RequestCacheInsert<'a>(VacantEntry<'a, MessageId, CachedRequest>, CachedRequest);
+
+impl<'a> RequestCacheInsert<'a> {
+    fn commit(self) {
```
ustulation: ah nicely done :+1:

PR #626, opened 2017-06-15
Mutable data

ustulation:
err polporene, this still doesn't mean test network today. Just getting the branches aligned so further patches can land in a single branch


maidsafe-highfive:
r? @maqi

PR #627, opened 2017-06-16
(WIP) Limit concurrent mutations

madadam:
**DO NOT MERGE** yet. Requires https://github.com/maidsafe/routing/pull/1469.

This PR implements limit on the number of concurrent mutations against the same data chunk. In particular, it allow a mutation only if it, together with all the other pending mutations, increments the the number of entries and/or size by at most half of the respective remaining allowance.

### DIFF
--- a/src/personas/data_manager/cache.rs
+++ b/src/personas/data_manager/cache.rs
@@ -17,7 +17,7 @@
 
 use super::STATUS_LOG_INTERVAL;
 use super::data::{DataId, ImmutableDataId, MutableDataId};
-use super::mutation::Mutation;
+use super::mutation::{self, Mutation};
 use GROUP_SIZE;
 use QUORUM;
 use maidsafe_utilities::serialisation::serialised_size;
@@ -25,6 +25,7 @@ use routing::{Authority, MAX_MUTABLE_DATA_ENTRIES, MAX_MUTABLE_DATA_SIZE_IN_BYTE
               MutableData, RoutingTable, Value, XorName};
 use std::collections::VecDeque;
 use std::collections::hash_map::Entry;
+use std::iter;
 use std::time::Duration;
 use utils::{self, HashMap, HashSet, Instant, SecureHash};
 
@@ -482,8 +483,8 @@ impl Cache {
             }
         }
 
-        let mut data = if let Some(data) = existing_data {
-            data.clone()
+        let data = if let Some(data) = existing_data {
+            data
         } else {
             return true;
         };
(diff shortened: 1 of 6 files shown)


maidsafe-highfive:
r? @Fraser999

src/personas/data_manager/cache.rs
```diff
@@ -469,30 +463,62 @@ impl Cache {
...
-                unindex_needed_fragment(&mut self.fragment_index, &fragment, holder_name);
+        let size_before = serialised_size(&data);
+        let size_remaining = MAX_MUTABLE_DATA_SIZE_IN_BYTES.saturating_sub(size_before);
+
+        let count_before = data.entries().len() as u64;
+        let count_remaining = MAX_MUTABLE_DATA_ENTRIES.saturating_sub(count_before);
+
+        for write in writes {
+            if !write.rejected {
+                write.mutation.apply(&mut data);
```
afck: I think we need to do this part twice, once for the entry count and once for the size:
In each case, we should _only_ apply the writes that _increase_ the count/size, because we need to compute the worst-case scenario, and the ones that decrease it might not go through.

src/personas/data_manager/cache.rs
```diff
@@ -469,30 +463,62 @@ impl Cache {
...
+        let size_diff = size_after.saturating_sub(size_before);
+        if size_diff > size_remaining / 2 {
+            return false;
         }
+
+        let count_after = data.entries().len() as u64;
+        let count_diff = count_after.saturating_sub(count_before);
+        if count_diff > count_remaining / 2 {
+            return false;
+        }
```
afck: Note that if we're e.g. at 99 out of 100 entries, we can never add another one, because that's going to be more than `count_remaining / 2`. I think the condition here should actually be:
```rust
count_diff * 2 > count_remaining + 1
```
Since the quorum is "strictly more than 50%", I think the amount we end up adding to the data can be anything "strictly less than twice" of what we allow here?

src/personas/data_manager/tests.rs
```diff
@@ -693,6 +693,123 @@ fn mdata_conflicting_parallel_mutations() {
     assert_match!(message.response, Response::MutateMDataEntries { res: Err(_), .. });
 }
 
+#[test]
+fn mdata_parallel_mutations_limits() {
+    let mut rng = rand::thread_rng();
```
afck: Should this use a `SeededRng` instead?

src/personas/data_manager/tests.rs
```diff
@@ -693,6 +693,123 @@ fn mdata_conflicting_parallel_mutations() {
...
+    // Refresh both mutations.
+    let message = unwrap!(node.sent_requests.remove(&msg_id_0));
+    let payload = assert_match!(message.request, Request::Refresh(payload, _) => payload);
+    unwrap!(dm.handle_group_refresh(&mut node, &payload));
+
+    let message = unwrap!(node.sent_requests.remove(&msg_id_1));
+    let payload = assert_match!(message.request, Request::Refresh(payload, _) => payload);
+    unwrap!(dm.handle_group_refresh(&mut node, &payload));
+
+    // Both request should succeed.
```
afck: `requests`

tests/data_manager.rs
```diff
@@ -189,10 +189,7 @@ fn mutable_data_normal_flow() {
     }
 
     // Mutate and verify the data.
-    let actions = test_utils::gen_mutable_data_entry_actions(&data,
-                                                             MAX_MUTABLE_DATA_ENTRY_ACTIONS as
-                                                             usize,
-                                                             &mut rng);
+    let actions = test_utils::gen_mutable_data_entry_actions(&data, 10, &mut rng);
```
maqi: why we no longer use `MAX_MUTABLE_DATA_ENTRY_ACTIONS` anymore and starts using 10 ?
if do need to, can we make it a const instead of using 10 directly in couple of places?

PR #628, opened 2017-06-20
Concurrent mutations limits

madadam:
~~**DO NOT MERGE** until https://github.com/maidsafe/routing/pull/1469 is merged.~~

### DIFF
--- a/src/personas/data_manager/cache.rs
+++ b/src/personas/data_manager/cache.rs
@@ -17,12 +17,15 @@
 
 use super::STATUS_LOG_INTERVAL;
 use super::data::{DataId, ImmutableDataId, MutableDataId};
-use super::mutation::Mutation;
+use super::mutation::{self, Mutation};
 use GROUP_SIZE;
 use QUORUM;
-use routing::{Authority, MessageId, MutableData, RoutingTable, Value, XorName};
+use maidsafe_utilities::serialisation::serialised_size;
+use routing::{Authority, MAX_MUTABLE_DATA_ENTRIES, MAX_MUTABLE_DATA_SIZE_IN_BYTES, MessageId,
+              MutableData, RoutingTable, Value, XorName};
 use std::collections::VecDeque;
 use std::collections::hash_map::Entry;
+use std::iter;
 use std::time::Duration;
 use utils::{self, HashMap, HashSet, Instant, SecureHash};
 
@@ -371,26 +374,18 @@ impl Cache {
         let hash = utils::secure_hash(&mutation);
         let data_id = mutation.data_id();
 
-        let mut writes = self.pending_writes
+        self.pending_writes
             .entry(data_id)
-            .or_insert_with(Vec::new);
-
-        if !rejected &&
-           writes
-               .iter()
-               .any(|other| !other.rejected && other.mutation.conflicts_with(&mutation)) {
-            return None;
-        } else {
-            writes.push(PendingWrite {
-                            hash: hash,
-                            mutation: mutation,
-                            timestamp: Instant::now(),
-                            src: src,
-                            dst: dst,
-                            message_id: msg_id,
-                            rejected: rejected,
-                        });
-        }
+            .or_insert_with(Vec::new)
+            .push(PendingWrite {
+                      hash: hash,
+                      mutation: mutation,
+                      timestamp: Instant::now(),
+                      src: src,
+                      dst: dst,
+                      message_id: msg_id,
+                      rejected: rejected,
+                  });
 
         if rejected {
             None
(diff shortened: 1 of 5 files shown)


maidsafe-highfive:
r? @Fraser999

src/personas/data_manager/mutation.rs
```diff
@@ -136,6 +169,59 @@ pub enum MutationType {
     ChangeMDataOwner,
 }
 
+/// Compute the size of the data after applying only those mutations that
+/// increase the size.
+pub fn compute_size_increase<'a, T>(data: &MutableData, mutations: T) -> u64
```
ustulation: The name of this function is misleading - it seems to suggest delta but actually returns an absolute

src/personas/data_manager/mutation.rs
```diff
@@ -136,6 +169,59 @@ pub enum MutationType {
     ChangeMDataOwner,
 }
 
+/// Compute the size of the data after applying only those mutations that
+/// increase the size.
```
ustulation: I cannot recall what was the reason for counting only the ones that increase the overall size as opposed to all mutations in a mutation request. So if there are 2 mutations in a request, one which increases the size by 10 and one that decreases by 5 so that once applied the overall increase would be 5, then why do we actually care about only that increases by 10 ? Should we not just apply all and calculate the overall or was there a specific reason to not do so ?

src/personas/data_manager/cache.rs
```diff
@@ -469,30 +464,59 @@ impl Cache {
...
+    /// all the pending mutations for that data.
+    pub fn validate_concurrent_mutations(&self,
+                                         existing_data: Option<&MutableData>,
+                                         new_mutation: &Mutation)
+                                         -> bool {
+        let writes = if let Some(writes) = self.pending_writes.get(&new_mutation.data_id()) {
+            writes
+        } else {
+            // Always accept if there are no other pending mutations.
+            return true;
```
ustulation: Why do we not want to check the case where this mutation alone can make the entry count > MAX or the size > MAX allowed ?

src/personas/data_manager/cache.rs
```diff
@@ -469,30 +464,59 @@ impl Cache {
...
+            }
         }
-    }
 
-    fn stop_expired_fragment_requests(&mut self) {
-        let mut empty_holders = Vec::new();
+        let data = if let Some(data) = existing_data {
+            data
+        } else {
+            return true;
```
ustulation: Why do we not want to check the case where this mutation alone can make the entry count > MAX or the size > MAX allowed? So if it's a (1st time) PUT of MData for instance, will this lead to allowing more than MAX size MData to be put into the cache ?

src/personas/data_manager/mutation.rs
```diff
@@ -136,6 +169,59 @@ pub enum MutationType {
...
+            data = new_data;
+        }
+    }
+
+    size
+}
+
+/// Compute the number of entries after applying only those mutations that
+/// increase the number of entries.
+pub fn compute_entry_count_increase<'a, T>(data: &MutableData, mutations: T) -> u64
```
ustulation: The name of this function is misleading - it seems to suggest delta but actually returns an absolute

src/personas/data_manager/mod.rs
```diff
@@ -682,7 +679,11 @@ impl DataManager {
             actions: actions.clone(),
         };
         let res = self.fetch_mdata(name, tag)
-            .and_then(|mut data| data.mutate_entries(actions.clone(), requester));
+            .and_then(|data| {
+                          data.clone().mutate_entries(actions.clone(), requester)?;
```
ustulation: Do we need these extra clones specially since the [blocker (for this task) PR](https://github.com/maidsafe/routing/pull/1469/files#diff-ba9c9e31131f3e63a4cfbea6a1fd129cR332) in routing does an internal clone and will not mutate data in case of error anyway ?

src/personas/data_manager/mutation.rs
madadam: Because we have no guarantee that the decreasing mutation will accumulate. We need to consider the worst case scenario.

src/personas/data_manager/cache.rs
madadam: Because this function only checks the validity of **multiple** concurrent mutations. It assumes that individually, each single mutation is already valid.

src/personas/data_manager/cache.rs
madadam: See the above comment.

src/personas/data_manager/mod.rs
madadam: Yeah, because we don't want to mutate the data even in case of success, as the function on the following line - `validate_concurrent_mutations` - needs the data to be in the same state as what is stored in the chunk store.

This is unfortunate. A proper solution would be to add new method to `MutableData` which only does the validation, but does not actually mutate the data.

PR #629, opened 2017-06-26
feat/all: implement hard delete of mutable data entries

madadam:


maidsafe-highfive:
r? @maqi

ustulation:
Not currently decided - but pls keep this work around for future @madadam

PR #630, opened 2017-06-30
Miscelaneous fixes and improvements

madadam:


maidsafe-highfive:
r? @Fraser999

tests/data_manager.rs
```diff
@@ -834,6 +834,97 @@ fn mutable_data_concurrent_mutations() {
     verify_data_is_stored(&mut nodes, &mut client, &all_data);
 }
 
+// Concurrently put and mutate a mutable data.
+// Verify the data is consistently stored (it may or might not be mutated, depending
+// on the order in which the request are processed by the nodes).
+#[test]
+fn mutable_data_concurrent_put_and_mutate() {
```
maqi: I reverted the change in data_manager/mutation.rs and executed this test for some iterations,
but haven't seen the failure yet.

have you seen the failure with this test and the old code?

madadam: Haven't tried it. Do you know what was the exact setup that triggered the bug before? I though it was put + some other mutation going on in parallel...

src/lib.rs
```diff
@@ -247,6 +248,7 @@ pub mod test_utils;
 #[cfg(all(feature = "use-mock-crust", not(feature = "use-mock-routing")))]
 pub mod mock_crust_detail;
 
+mod authority;
```
madadam: But it does sort of replace routing's `Authority`. It adds first-class types for some of it's variants, for better type safety. Not sure how else to call it.

src/mock_crust_detail/test_client.rs
```diff
@@ -29,12 +30,19 @@ use std::collections::{BTreeMap, BTreeSet};
 use std::iter;
 use std::sync::mpsc::TryRecvError;
 
+const RETRY_DELAY_MS: u64 = 800;
```
madadam: It's the same value used in safe_core. Don't know why it was chosen. If we want to be really pedantic, we might define and expose this constant in routing.

tests/data_manager.rs
maqi: yes, it was a concurrent PutMData and SetPermission mutation reaches pending_writes cache at the same time.
However, it may just need to execute `more` iterations to have the problem hit by luck, and I don't know how many iterations will take (or even never ?)

PR #633, opened 2017-07-11
MAID-2099: add dev config options

nbaksalyar:
Add dev configuration option to allow running multiple nodes on the same local network.

Also, change handling of chunk store root: instead of using a single hard-coded path create a temp directory with a random name (with a prefix) for each vault run. If you want a stable name, just provide it in the vault config file as the `chunk_store_root` config value.


maidsafe-highfive:
r? @maqi

maqi:
replaced by PR https://github.com/maidsafe/safe_vault/pull/637

Fraser999:
Closed in favour of #637.

PR #635, opened 2017-07-14
(WIP) feat/data-manager: add support for idempotent operations

madadam:


maidsafe-highfive:
r? @Fraser999

src/personas/data_manager/mod.rs
```diff
@@ -392,8 +392,16 @@ impl DataManager {
...
-        let res = if self.chunk_store.has(&data_id) {
-            Err(ClientError::DataExists)
+        let res = if let Ok(old_data) = self.chunk_store.get(&data_id) {
+            if old_data == data {
+                // Putting the same data more than once is allowed to support idempotence.
+                routing_node
+                    .send_put_mdata_response(dst, src, Ok(()), msg_id)?;
+                return Ok(());
+            } else {
+                // But putting different data with the same id is not.
```
maqi: what if the client sends seq of request with: put(v_0), update(v_1), put(v_0)  ?

tests/data_manager.rs
```diff
@@ -1156,6 +1157,87 @@ fn mutable_data_operations_with_churn() {
...
+
+    client.ensure_connected(&mut nodes);
+    client.create_account(&mut nodes);
+
+    // Sequentially putting the same mdata more than once.
+    let data = test_utils::gen_mutable_data(10000, 2, *client.signing_public_key(), &mut rng);
+    unwrap!(client.put_mdata_response(data.clone(), &mut nodes));
+    unwrap!(client.put_mdata_response(data, &mut nodes));
+
+    // Concurrently putting the same mdata more than once.
```
maqi: better using two clients to carry out this concurrently put ?

tests/data_manager.rs
```diff
@@ -1156,6 +1157,87 @@ fn mutable_data_operations_with_churn() {
...
+
+    // Sequentially mutating the data with the same actions more than once.
+    let actions = test_utils::gen_mutable_data_entry_actions(&data, 2, &mut rng);
+    unwrap!(client.mutate_mdata_entries_response(*data.name(),
+                                                 data.tag(),
+                                                 actions.clone(),
+                                                 &mut nodes));
+    unwrap!(client.mutate_mdata_entries_response(*data.name(), data.tag(), actions, &mut nodes));
+
+    // Concurrently mutating the data with the same actions more than once.
```
maqi: using two clients to carry out this concurrently mutate can be more convincing?

tests/data_manager.rs
```diff
@@ -1156,6 +1157,87 @@ fn mutable_data_operations_with_churn() {
...
+                }
+                if msg_id == msg_id_1 {
+                    responded_1 = true
+                }
+            }
+        }
+    }
+
+    assert!(responded_0);
+    assert!(responded_1);
```
maqi: why not just put msg_ids in BTreeSet, then only need to insert received msg_id assert tree_set are identical?
same comment to the above.

ustulation:
We are going for a different paradigm that does not affect routing and vaults: MAID-2146 in JIRA.

PR #636, opened 2017-07-17
test/data_manager: simulate delayed group refresh messages on DataManager

maqi:


maidsafe-highfive:
r? @Fraser999

src/personas/data_manager/mod.rs
```diff
@@ -1392,6 +1407,50 @@ impl DataManager {
         Ok(result)
     }
 
+    pub fn delay_group_refreshes(&mut self, routing_node: &mut RoutingNode, delayed: bool) {
+        if let Some(cache) = self._delayed_group_refresh_cache.clone() {
```
Fraser999: We could use `take()` here to avoid `clone()` or even `mem::replace`:
```rust
if let Some(cache) = mem::replace(&mut self._delayed_group_refresh_cache,
                                  if delayed { Some(BTreeSet::new()) } else { None }) {
    for delayed_refresh in cache {
        let _ = self.handle_group_refresh(routing_node, delayed_refresh);
    }
}
```

tests/data_manager.rs
```diff
@@ -834,13 +834,13 @@ fn mutable_data_concurrent_mutations() {
     verify_data_is_stored(&mut nodes, &mut client, &all_data);
 }
 
-// Concurrently put and mutate a mutable data.
-// Verify the data is consistently stored (it may or might not be mutated, depending
-// on the order in which the request are processed by the nodes).
+// Put a mutable data then immediately mutate it, meanwhile simulates out-of-order group_refresh.
+// This test was used to expose an issue that a mutate request conflicts with a `PutMData` entry in
+// the pending_writes of DM's ache.
```
Fraser999: Typo `ache`.

src/personas/data_manager/mod.rs
maqi: `handle_group_refresh` will call `handle_delayed_group_refresh`, which requires the entry `exists` in the self._delayed_group_refresh_cache to be properly popped out and to be handled. 
so I don't think we can use `take` or `mem::replace` here?

Fraser999: Yeah - good call.

PR #643, opened 2017-08-17
MAID-2259 refactor/rate_limiter: removes proxy rate exceed event.

maqi:
Remove custom mock crust handling for rate exceeded events which is handled by routing.


maidsafe-highfive:
r? @Fraser999

PR #644, opened 2017-08-22
Version changed to 0.17.0;

maqi:


maidsafe-highfive:
r? @Fraser999

Cargo.toml
```diff
@@ -32,7 +32,7 @@ tiny-keccak = "~1.3.1"
 unwrap = "~1.1.0"
 
 [build-dependencies]
-hyper = {version = "~0.9.10", optional = true}
+hyper = {version = "~0.11.2", optional = true}
```
Fraser999: This update causes the build to break (in build.rs) with `--features=generate-diagrams`.

PR #645, opened 2017-08-25
MAID-2297: change test to use 'MAX_MUTABLE_DATA_ENTRIES' rather than magic numbers

Fraser999:


maidsafe-highfive:
r? @maqi

PR #648, opened 2017-09-29
changes to make it easier to build and deploy docker containers

mattmccarty:
- Translate environment variables (the usual way to configure the Docker containers) into the config files by replacing template strings with `sed`
- Add the ability to compile a fresh `safe_vault` binary inside of the container if it's missing (i.e. there's no need to pre-compile it now)
- Simplify the task of running several vaults in parallel (if the min. section size setting is > 1)
- Add support for `docker-compose`
- Improve documentation


maidsafe-highfive:

mattmccarty:
Hopefully this will make it easier to develop, build and deploy using Docker. I have added the ability to build using docker-compose as well. The changes will allow a developer (or a CI service like Jenkins) to build from source and set configuration values by using environment variables.

nbaksalyar:
@mattmccarty, thanks for the PR!
This looks good, but it might take some time for us to review and approve it.
You should hear from us in a day or two.
Thank you!

mattmccarty:
No worries!

Dockerfile
```diff
@@ -0,0 +1,115 @@
...
+ENV     EDITOR /usr/bin/vim
+WORKDIR /home/maidsafe/src/safe_vault
+
+#
+# Copy files and folders to the current working directory
+#
+COPY src/         ./src/
+COPY tests/       ./tests/
+COPY installer/   ./installer/
+COPY target/      ./target/
```
nbaksalyar: I would probably optimise this as
```
COPY target/release/safe_vault ./target/release/safe_vault
```
We don't need the rest of build targets (and debug builds can get quite large, up to several GBs) or any dependencies because if we set the target to `x86_64-linux-unknown-musl` the compiler will build a static binary.

I'd also add a `.dockerignore` file with the following contents:
```
target/
!target/release/safe_vault
```
This way we won't need to send the entire `target` dir to the Docker server and it will make the build faster.

mattmccarty: Hello @nbaksalyar. Thanks for the feed back. What you are saying makes sense to me. I will implement the changes this weekend. I'm just getting back in town from travelling, so it may take me until Sunday.

mattmccarty: Just as an update, work has had me slammed. I'll get to these changes this week.

mattmccarty:
Changes implemented @nbaksalyar :)

PR #656, opened 2018-02-26
chore/ci: update ci scripts to latest rust stable

m-cat:


maidsafe-highfive:

src/chunk_store/mod.rs
```diff
@@ -21,7 +21,7 @@
 #[cfg(test)]
 mod tests;
 
-use hex::{FromHex, ToHex};
+use hex::{FromHex, encode};
```
Fraser999: We normally don't bring functions into scope like this, only structs and enums.

m-cat: Ah, first time I hear about this rule. Does this apply for the rest of our Rust codebase or just safe_vault? We import functions liberally in client libs.

PR #659, opened 2018-05-09
[MAID-2652] chore/license: update license to include linking exception

m-cat:


maidsafe-highfive:
r? @Fraser999

PR #660, opened 2018-06-01
[MAID-2670] chore/ci: update ci scripts to latest rust and rustfmt

m-cat:


.travis.yml
```diff
@@ -1,6 +1,6 @@
-sudo: false
-
-language: rust
+vars:
+  - &stable "1.26.1"
```
nbaksalyar: Great idea on using anchors for this purpose! 👍 Would be nice to change other CI scripts in the same way too.

PR #670, opened 2019-01-04
Upgrade to rust 2018 edition

pierrechevalier83:
* Upgrade compiler
* Switch to 2018 edition
* Fix compilation, fmt and clippy warnings
* Remove nightly from CI
* Remove crud from CI configuration


maidsafe-highfive:
r? @maqi

PR #675, opened 2019-01-22
Fix builds with mocked crust and crypto

povilasb:
There was a special variation when Crust and crypto was mocked:
```
cargo test --features="use-mock-crust,use-mock-crypto"
```

In such case builds would be failing because some of the modules always used real `rust_sodium` disregard to `use-mock-crypto` feature. This PR fixes that and makes sure mocked `rust_sodium` from Routing will be used when `use-mock-crypto` feature is on.


maidsafe-highfive:

tests/data_manager.rs
```diff
@@ -11,15 +11,19 @@
...
 #![cfg(feature = "use-mock-crust")]
 #![cfg(not(feature = "use-mock-routing"))]
+use self::rust_sodium::crypto::sign;
 use fake_clock::FakeClock;
-use log::{log, trace};
+use log::trace;
 use rand::Rng;
 use routing::mock_crust::Network;
+#[cfg(feature = "use-mock-crypto")]
+use routing::mock_crypto::rust_sodium;
```
ustulation: Do we still need to mention "rust-sodium" when safe-crypto is in use @povilasb ? Wern't those details supposed to be abstracted away ? (also elsewhere in the file)

Issue #711, opened 2019-06-13
Add Quic-P2P for client connections


Fraser999:

Lindsey17:
Reopened for @octol to write the handler code for receiving a new connection event from quic-p2p

PR #713, opened 2019-06-17
Handle new connections

octol:


src/source_elder.rs
```diff
@@ -6,45 +6,115 @@
...
+        (quic_p2p, event_receiver)
+    }
+
+    pub fn is_client(&self, peer_addr: &SocketAddr) -> bool {
+        self.clients.contains_key(peer_addr)
+    }
+
+    pub fn handle_new_connection(&mut self, peer: Peer) -> Option<Action> {
+        // If we already know the peer, drop the connection attempt
+        if self.client_candidates.contains_key(&peer.peer_addr()) {
```
Fraser999: Think we should be checking `self.clients` too.

src/utils.rs
```diff
@@ -5,3 +5,13 @@
 // under the GPL Licence is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
 // KIND, either express or implied. Please review the Licences for the specific language governing
 // permissions and limitations relating to use of the SAFE Network Software.
+
+use rand::RngCore;
+
+#[allow(unsafe_code)]
```
Fraser999: I think there should be a few ways to do this without using `unsafe`.  Maybe something like `sample_iter().take(size)` would be safer?

src/source_elder.rs
```diff
@@ -6,45 +6,115 @@
...
+            return None;
+        }
+        // Peer here is assumed to be `Peer::Client` during phase 1, and has been checked at the
+        // Vault layer.
+        let challenge = random_vec(8);
+        let _ = self
```
Fraser999: Probably better to only insert this if the subsequent `serialize` call succeeds.

src/source_elder.rs
```diff
@@ -6,45 +6,115 @@
...
+        }
+        None
+    }
+
+    pub fn handle_terminated_connection(&mut self, peer_addr: SocketAddr) -> Option<Action> {
+        let _ = self.clients.remove(&peer_addr);
+        None
+    }
+
+    pub fn handle_establised_connection(
```
Fraser999: Typo: "established"

src/source_elder.rs
```diff
@@ -6,45 +6,115 @@
...
+                Ok(()) => {
+                    if let Some(peer) = self.client_candidates.remove(&peer_addr) {
+                        let _ = self.clients.insert(peer_addr, public_id);
+                    } else {
+                        info!("Failed to promote client from candidate to established");
+                    }
+                    None
+                }
+                Err(err) => {
+                    info!("Failed to established connection: {}", err);
```
Fraser999: Typo: "establish"

Lindsey17:
This PR resolves #711

src/source_elder.rs
```diff
@@ -6,45 +6,115 @@
...
+        None
+    }
+
+    pub fn handle_established_connection(
+        &mut self,
+        peer_addr: SocketAddr,
+        public_id: ClientPublicId,
+        signature: Signature,
+    ) -> Option<Action> {
+        if let Some(challenge) = self.client_candidates.get(&peer_addr) {
```
Fraser999: We could call `self.client_candidates.remove()` here rather than `get` and that would save us calling it below.

src/source_elder.rs
```diff
@@ -6,45 +6,115 @@
...
+            return match public_id.public_key().verify(&signature, challenge) {
```
Fraser999: No need for `return` here.

src/vault.rs
```diff
@@ -50,13 +47,120 @@ impl Vault {
...
+                            self.handle_message(peer_addr, msg)
+                        }
+                        Err(err) => {
+                            info!("Unable to deserialise message: {}", err);
+                            None
+                        }
+                    }
+                } else {
+                    match bincode::deserialize(&msg) {
+                        Ok(challenge) => {
```
Fraser999: We could maybe match on `Challenge::Response` here, and pass `public_id` and `signature` to `handle_challenge()` instead?  It'd probably be worth logging the case where the client sent a `Challenge::Request` too.

src/source_elder.rs
octol: This changes the logic slightly (when verify fails), but maybe that's preferable anyway?

Fraser999: Yes, I think it might be worth only giving the client one chance to validate itself.

src/vault.rs
octol: Ok moved that part too into `handle_quic_p2p_event`. It's quite a deeply nested method now though...

Fraser999: Ah - I wasn't clear enough!  I meant something more like:
```rust
match bincode::deserialize(&msg) {
    Ok(Challenge::Response(public_id, signature)) => {
        trace!("Received challenge response from {}", peer_addr);
        self.handle_challenge(peer_addr, public_id, signature)
    }
    Ok(Challenge::Request(_)) => {
        info!("Received unexpected challenge request");
        None
    }
    Err(err) => {
        info!("Unable to deserialise challenge: {}", err);
        None
    }
}
```

octol: ah yes that's much nicer :)

PR #714, opened 2019-06-19
chore/all: update config handling

Fraser999:


src/source_elder.rs
```diff
@@ -53,65 +62,140 @@ impl SourceElder {
...
+                    info!("{} invalidly sent {:?}", client_id, response);
+                }
+                Err(err) => {
+                    info!("Unable to deserialise message from {}: {}", client_id, err);
+                }
+            }
+        } else {
+            match bincode::deserialize(&bytes) {
+                Ok(Challenge::Response(public_id, signature)) => {
+                    // TODO - Fix before committing.
```
octol: What does this comment refer to?

octol:
The appveyor failure looks like it's a real one?

src/config_handler.rs
```diff
@@ -8,45 +8,153 @@
...
 };
+use structopt::StructOpt;
+use unwrap::unwrap;
 
 const CONFIG_DIR_QUALIFIER: &str = "net";
 const CONFIG_DIR_ORGANIZATION: &str = "MaidSafe";
 const CONFIG_DIR_APPLICATION: &str = "vault";
 const CONFIG_FILE: &str = "vault.config";
-const LOG_CONFIG_FILE: &str = "vault.log.config";
+const DEFAULT_ROOT_DIR_NAME: &str = "safe_vault";
```
octol: I notice `CONFIG_DIR_APPLICATION` is `vault`. Should we change that to also be `safe_vault`?

src/config_handler.rs
```diff
@@ -8,45 +8,153 @@
...
 pub struct Config {
-    /// Used to store the address where SafeCoin will be sent.
-    pub wallet_address: Option<XorName>,
-    /// Upper limit for allowed network storage on this vault.
-    pub max_capacity: Option<u64>, // measured by Bytes
-    /// Root directory for chunk_store directories.
-    pub chunk_store_root: Option<String>,
+    /// The address to be credited when this vault farms SafeCoin.
+    #[structopt(short, long, parse(try_from_str))]
+    // TODO - Fix before committing - Use ClientPublicId rather than String
```
octol: Does the comment here indicate you wanted to change this before raising the PR?

src/config_handler.rs
```diff
@@ -8,45 +8,153 @@
...
+            self.quic_p2p_config.our_complete_cert = Some(unwrap!(value.parse()));
+        } else if arg == ARGS[10] {
+            self.quic_p2p_config.our_type = unwrap!(value.parse());
+        } else {
+            println!("ERROR");
+        }
+    }
+
+    /// Reads the default vault config file.
+    fn read_from_file() -> Result<Config> {
```
octol: Put a trace in there about which file is read?

src/destination_elder.rs
```diff
@@ -6,8 +6,20 @@
...
+    vault::Init,
+    Result,
+};
 use pickledb::PickleDb;
 use safe_nd::NodePublicId;
+use std::{cell::RefCell, path::Path, rc::Rc};
+
+const IMMUTABLE_META_DB_NAME: &str = "i_data.db";
+const MUTABLE_META_DB_NAME: &str = "m_data.db";
+const APPEND_ONLY_META_DB_NAME: &str = "a_data.db";
```
octol: Hey you claimed you didn't like abbreviated names ;)

src/source_elder.rs
octol: Should `public_id` in `Challenge::Response` be changed to `PublicId`?

nbaksalyar:
>The appveyor failure looks like it's a real one?

Yes, but it shouldn't matter: it's on a 32-bit system and it fails because `mem::size_of<Config>` expectedly returns a different value.

We should disable CI for x86 Windows.

src/chunk_store.rs
```diff
@@ -42,7 +43,6 @@ use std::{
 use used_space::UsedSpace;
 
 const CHUNK_STORE_DIR: &str = "chunks";
-const DEFAULT_TOTAL_CAPACITY: u64 = 2 * 1024 * 1024 * 1024;
 
 /// The max name length for a chunk file.
 const MAX_CHUNK_FILE_NAME_LENGTH: usize = 104;
```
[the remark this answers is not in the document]
Fraser999: I'm not sure, but that sounds plausible.

src/config_handler.rs
Fraser999: Yes, it does :D  I have a pre-commit hook which searches for this term, but I've disabled that for now since we have a similar comment about removing `#![allow(unused)]` in lib.rs.

I'll update `wallet_address` in a fixup commit.

src/destination_elder.rs
Fraser999: Yeah, I really don't when it's not a commonplace abbreviation :D 

I guess I don't apply that as strictly to DB filenames though.  I don't think it's as important to have them as clearly named as variables, since there's less chance of that causing a bug in the code.  Also, there's always the filesystem limit on absolute path length which niggles at the back of my mind whenever I create a file on disk at a location where the user can set the root dir.  Probably both weak arguments here, so I'll rename these :)

src/source_elder.rs
Fraser999: Yes, exactly.  I should have gone into more details in the comment, but https://github.com/maidsafe/safe-nd/pull/34 would break this build, so it won't get forgotten anyway.  I'll delete the comment.

Fraser999: And update the code too now since that PR is merged.

PR #716, opened 2019-06-21
 chore/all: stub handlers for client->source elders RPCs 

Fraser999:


src/coins_handler.rs
```diff
@@ -8,18 +8,29 @@
...
+use safe_nd::NodePublicId;
+use std::{
+    fmt::{self, Display, Formatter},
+    path::Path,
+};
 
 const COINS_DB_NAME: &str = "coins.db";
 
 pub(crate) struct CoinsHandler {
+    _id: NodePublicId,
```
octol: Nitpick: you're using this in `fmt` so you can remove the underscore

src/source_elder.rs
```diff
@@ -147,11 +175,145 @@ impl SourceElder {
...
+
+    fn is_valid_client_signature(
+        &self,
+        client_id: &PublicId,
+        request: &Request,
+        message_id: &MessageId,
+        signature: &Signature,
+    ) -> bool {
+        let pub_key = match client_id {
+            PublicId::Node(_) => return false, // unreachable - no need to log
```
octol: Should we put in a `warn!` here to detect if this (accidentally) becomes reachable?

src/vault.rs
```diff
@@ -116,14 +132,48 @@ impl Vault {
...
+                info!("{}: Unexpected event: {}", self, event);
             }
         }
         None
     }
 
-    fn handle_action(&mut self, _action: Action) -> Option<Action> {
-        None
+    fn handle_action(&mut self, action: Action) -> Option<Action> {
+        use Action::*;
```
octol: I must say I like this habit of importing only in local scope, something I'll do more now :)

src/source_elder.rs
Fraser999: Yeah - fair point.  I'll make it `error!` though, since it'd represent a true logic error.

src/vault.rs
Fraser999: Well, afaik, we're only really meant to do this in fairly limited cases.  Normally we're supposed to do all the importing at the start of the module.  We agreed to make an exception for this sort of case where repeating the enum's name for every variant is a bit tedious and doesn't improve clarity.

octol: Ok fair enough if it's style rule

(As a general overarching rule though, isn't it preferable to only keep things in the smallest scope necessary?)

Fraser999: Yeah, I can certainly see that point too, although what we're doing seems to be fairly standard.  One for the next coding standards discussion maybe?

PR #717, opened 2019-06-25
feat/all: add stubs for remaining RPCs

### LINKED ISSUE #683: Define and stub out the RPC handlers
- Get the RPC handlers stubbed out (unimpl so others can code).

Fraser999:
Closes #683

### DIFF
--- a/src/destination_elder.rs
+++ b/src/destination_elder.rs
@@ -13,11 +13,14 @@ use crate::{
     vault::Init,
     Result,
 };
+use log::{error, trace, warn};
 use pickledb::PickleDb;
-use safe_nd::{MessageId, NodePublicId, Request, Signature, XorName};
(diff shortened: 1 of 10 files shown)


src/destination_elder.rs
```diff
@@ -91,13 +95,273 @@ impl DestinationElder {
...
+    }
+
+    fn handle_put_idata(
+        &mut self,
+        src: XorName,
+        kind: IDataKind,
+        message_id: MessageId,
     ) -> Option<Action> {
+        if &src == kind.name() {
+            // We're a dst elder chosen to store the chunk
```
octol: Help me understand this part

If the client `XorName` that sent the message is the same as the `XorName` of the data we store it...shouldn't we check against our own `XorName` instead?

Issue #691, opened 2019-06-13
Handle GetPubIData and GetUnpubIData


Fraser999:
Tasks:
- [x] Source to destination elder flow #718 
- [x] Destination elder to adult flow #725 
- [x] ~Check published/unpublished matches between req/resp #733~ 
- [x] Check owner permission #732 
- [x] Tests (partially #736)

PR #718, opened 2019-06-26
Get published and unpublished immutable data

octol:
Get immutable data

- Mark clients as registered/unregistered when connecting
- Handle GetIData RFC for both published and unpublished immutable data

Remains to be implemented

- Check permissions for unpublished immutable data
- ~Return to source elders~ and then client
- ~Return failure response in source elder to client~

Part of #691

### DIFF
--- a/src/source_elder.rs
+++ b/src/source_elder.rs
@@ -426,7 +426,7 @@ impl SourceElder {
         }
     }
 
-    pub fn _handle_node_response(
+    pub fn handle_node_response(
         &mut self,
         dst_elders: XorName,
         src_elders: XorName,
@@ -449,7 +449,11 @@ impl SourceElder {
             // ===== Immutable Data =====
             //
             PutIData(result) => unimplemented!(),
-            GetIData(result) => unimplemented!(),
+            GetIData(ref result) => {
+                let _msg = utils::serialise(&response);
+                // TODO - Send this msg back to the client
+                None
+            }
             DeleteUnpubIData(result) => unimplemented!(),
             //
             // ===== Mutable Data =====
--- a/src/source_elder.rs
+++ b/src/source_elder.rs
@@ -14,8 +14,8 @@ use log::{error, info, trace, warn};
 use pickledb::PickleDb;
 use quic_p2p::{Config as QuicP2pConfig, Event, Peer, QuicP2p};
 use safe_nd::{
-    AppPermissions, Challenge, ClientPublicId, Coins, Message, MessageId, NodePublicId, PublicId,
-    PublicKey, Request, Response, Signature, XorName,
+    AppPermissions, Challenge, ClientPublicId, Coins, Error as NdError, Message, MessageId,
+    NodePublicId, PublicId, PublicKey, Request, Response, Signature, XorName,
 };
 use serde::{Deserialize, Serialize};
 use std::{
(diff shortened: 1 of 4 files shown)


src/source_elder.rs
```diff
@@ -37,10 +37,16 @@ struct ClientAccount {
     balance: Coins,
 }
 
+#[derive(Clone, PartialEq, Eq)]
+enum RegisteredClient {
```
Fraser999: I'd maybe just name this `ClientState` or `ClientStatus` to avoid "registered" so much.  Not a biggie though.

src/source_elder.rs
```diff
@@ -208,7 +221,21 @@ impl SourceElder {
                     message_id,
                 })
             }
-            GetIData(ref address) => unimplemented!(),
+            GetIData(ref address) => {
+                let owner = utils::owner(client_id)?;
+                if address.published()
+                    || (registered_client == RegisteredClient::Registered
+                        && address.name() == owner.name())
```
Fraser999: `address.name() == owner.name()` shouldn't be done - address.name() is the chunk's name and there's no requirement for that to equal the client's name.  We can't check any more than `registered_client == RegisteredClient::Registered` here.

At the dst elders we would check that the `UnpubImmutData::owners().name() == src`.

src/source_elder.rs
```diff
@@ -208,7 +221,21 @@ impl SourceElder {
...
+                {
+                    return Some(Action::ForwardClientRequest {
+                        client_name: *client_id.name(),
+                        request,
+                        message_id,
+                    });
+                } else {
+                    return None;
```
Fraser999: We'll be returning a failure response back to the client here.

src/source_elder.rs
```diff
@@ -340,8 +367,13 @@ impl SourceElder {
         if let Some(challenge) = self.client_candidates.remove(&peer_addr) {
             match public_key.verify(&signature, challenge) {
                 Ok(()) => {
+                    let registered = if self.client_accounts.exists(&public_id.to_string()) {
+                        RegisteredClient::Registered
+                    } else {
+                        RegisteredClient::Unregistered
+                    };
                     info!("{}: Accepted {} on {}", self, public_id, peer_addr);
```
Fraser999: Probably worth adding that info (registered or not) to the log line.

src/source_elder.rs
```diff
@@ -340,8 +367,13 @@ impl SourceElder {
         if let Some(challenge) = self.client_candidates.remove(&peer_addr) {
             match public_key.verify(&signature, challenge) {
                 Ok(()) => {
+                    let registered = if self.client_accounts.exists(&public_id.to_string()) {
```
Fraser999: A couple of points here:
1. We need to destructure the pub ID.  If it's an app, we get the owner's account from `self.client_accounts` and check it's an authorised app in there.  If it's a client, then we just need to check the account exists (as you're doing here)
2. That `public_id.to_string()` will need to be replaced with `client_id.to_db_key()`.  This will be the `ClientPublicId` type, not the `PublicId` enum (the `PublicId` enum doesn't implement `ToDbKey`, so that should help avoid errors like this).  For the case of the pub ID representing an app, an `AppPublicId` has an `owners()` getter for the owner's `ClientPublicId`, which is what the account will be stored under in `self.client_accounts`

PR #722, opened 2019-07-01
chore/source_elder: extract determine client state

octol:
Minor follow up tidying, extracting out a method in `handle_challenge`

### DIFF
--- a/src/source_elder.rs
+++ b/src/source_elder.rs
@@ -387,27 +387,7 @@ impl SourceElder {
         if let Some(challenge) = self.client_candidates.remove(&peer_addr) {
             match public_key.verify(&signature, challenge) {
                 Ok(()) => {
-                    let registered = ClientState::from_bool(match public_id {
-                        PublicId::Client(ref pub_id) => {
-                            self.client_accounts.exists(&pub_id.to_db_key())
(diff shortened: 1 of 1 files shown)


src/source_elder.rs
```diff
@@ -426,6 +406,25 @@ impl SourceElder {
...
+        ClientState::from_bool(match public_id {
+            PublicId::Client(ref pub_id) => self.client_accounts.exists(&pub_id.to_db_key()),
+            PublicId::App(ref app_pub_id) => {
+                let owner = app_pub_id.owner();
+                let app_pub_key = app_pub_id.public_key();
+                self.client_accounts
+                    .get(&owner.to_db_key())
+                    .and_then(|account: ClientAccount| account.apps.get(app_pub_key).cloned())
+                    .is_some()
+                // TODO: actually check permissions
```
Fraser999: This todo can get removed.  Permissions will only be checked on the basis of future received requests from this client, and mostly they'll be checked at the destination elders (only checked at source for being allowed to transfer Coins currently, although that will likely extend to any action which requires debiting the owner's balance I expect).

octol:
Part of #691

PR #725, opened 2019-07-02
GetIData destination elder to adult flow

octol:
The destination elder to adult flow for immutable get.

Part of #691. 
Final thing left to implement is verifying owner when the data is unpublished.

### DIFF
--- a/src/destination_elder.rs
+++ b/src/destination_elder.rs
@@ -15,22 +15,23 @@ use crate::{
     vault::Init,
     Result, ToDbKey,
 };
-use idata_op::IDataOp;
+use idata_op::{IDataOp, RpcState};
 use log::{error, trace, warn};
 use pickledb::PickleDb;
 use safe_nd::{
-    IDataAddress, IDataKind, MessageId, NodePublicId, Request, Response, Result as NdResult,
-    XorName,
+    Error as NdError, IDataAddress, IDataKind, MessageId, NodePublicId, Request, Response,
+    Result as NdResult, XorName,
 };
 use serde::{Deserialize, Serialize};
 use std::{
     cell::RefCell,
-    collections::BTreeMap,
+    collections::{btree_map::Entry, BTreeMap},
     fmt::{self, Display, Formatter},
     iter,
     path::Path,
     rc::Rc,
 };
+use unwrap::unwrap;
 
 const IMMUTABLE_META_DB_NAME: &str = "immutable_data.db";
 const MUTABLE_META_DB_NAME: &str = "mutable_data.db";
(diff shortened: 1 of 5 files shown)


src/destination_elder.rs
```diff
@@ -474,6 +469,49 @@ impl DestinationElder {
...
+            );
+            return None;
+        };
+
+        if idata_ops
+            .rpc_states()
+            .values()
+            .any(|rpc_state| rpc_state == &RpcState::Actioned)
+        {
+            return None;
```
octol: Need to toggle the state before returning so we can eventually remove the IDataOp from the cache

src/destination_elder.rs
```diff
@@ -474,6 +469,49 @@ impl DestinationElder {
...
+            rpc_state
+        } else {
+            warn!(
+                "{}: Received response from sender {} that we didn't expect.",
+                self, sender
+            );
+            return None;
+        };
+
+        *rpc_state = RpcState::Actioned;
```
octol: Maybe we should consider removing the IDataOp on the first returned response

src/destination_elder.rs
```diff
@@ -428,41 +429,36 @@ impl DestinationElder {
...
+            let idata_op = unwrap!(IDataOp::new(
+                src,
+                Request::GetIData(address),
+                metadata.holders.clone()
+            ));
+            match self.idata_ops.entry(message_id) {
+                Entry::Occupied(_) => {
+                    // TODO - Consider return another Error
+                    Some(Action::RespondToClient {
+                        sender: *self.id.name(),
```
Fraser999: I think here we need to be using the name of the chunk as `sender`, since we're replying as part of the section rather than an individual vault.

src/destination_elder.rs
```diff
@@ -428,41 +429,36 @@ impl DestinationElder {
...
-                );
-                return None;
-            };
-
-            if metadata_holder == self.id.name() {
-                self.get_idata(address, message_id)
-            } else {
-                // TODO - Send Get request to adult (or elder occasionally) with src = data.
-                None
+            // Can't fail
```
Fraser999: I can't apply the comment at the right place, but at line 429 I think we should be replying with a `safe_nd::Error::NoSuchData` to the client here (so returning `Some(Action::RespondToClient{...})`).

src/destination_elder/idata_op.rs
```diff
@@ -61,11 +61,48 @@ impl IDataOp {
         &self.request
     }
 
-    pub(super) fn rpc_states(&self) -> &BTreeMap<XorName, RpcState> {
-        &self.rpc_states
+    pub(super) fn is_actioned(&self) -> bool {
```
Fraser999: Probably better to rename this to `is_any_actioned()` to be explicit.  (Just that for the Put case, we probably wouldn't want to react if _any_ were actioned, but rather once enough have succeeded.

src/destination_elder/idata_op.rs
```diff
@@ -61,11 +61,48 @@ impl IDataOp {
...
+    pub(super) fn handle_response(
+        &mut self,
+        sender: XorName,
+        response: Response,
+        own_id: String,
+        message_id: MessageId,
+    ) -> Option<Action> {
+        let is_already_actioned = self.is_actioned();
+        match response {
+            Response::GetIData(ref result) => self
```
Fraser999: We should probably destructure `self.request` here for a couple of reasons.  First, we can check the variant matches (i.e. the request is a Get too).  Secondly, we can use the contained data address as the `sender` field of `Action::RespondToClient` and avoid passing it into this method.

src/source_elder.rs
```diff
@@ -241,7 +241,9 @@ impl SourceElder {
                 })
             }
             GetIData(ref address) => {
-                let owner = utils::owner(client_id)?;
+                if !address.published() {
+                    self.has_signature(client_id, &request, &message_id, &signature)?;
```
Fraser999: Nice catch :)

src/destination_elder/idata_op.rs
octol: We need `sender` for the `get_mut(&sender)` call though?

PR #726, opened 2019-07-02
feat/destination_elder: add IDataOp and SendToPeers

octol:
Common scaffolding used for put and get immutable data


octol:
Part of #691

PR #727, opened 2019-07-02
Test improvements

madadam:
This PR implements some modifications to improve the test experience:

1. `Config::default` no longer calls `Config::new`, instead it just populates the fields with default values. This is useful in tests to avoid the command line parsing which was otherwise interfering with the test runner flags.

2. There is new feature flag to disable dumping the db to the filesystem. Again, this is useful in tests, to avoid multiple test threads to interfere with each others databases and also to speed the tests up.


madadam:
Closed in favor of https://github.com/maidsafe/safe_vault/pull/728.

PR #731, opened 2019-07-04
chore/mock-quic-p2p: update Event to match latest quic-p2p

octol:


src/source_elder.rs
```diff
@@ -134,7 +134,8 @@ impl SourceElder {
         info!("{}: Connected to new client on {}", self, peer_addr);
     }
 
-    pub fn handle_connection_failure(&mut self, peer_addr: SocketAddr) {
+    pub fn handle_connection_failure(&mut self, peer_addr: SocketAddr, error: Error) {
+        info!("{}: {}", self, error);
```
madadam: Let's be a bit more descriptive here:
```rust
info!("{}: connection failure: {}", self, error);
```

octol: Doesn't the error already contain that string?

madadam: Hm, seems that it does. Not in mock though, where the error is just `struct Error;`. But that's probably OK.

PR #732, opened 2019-07-04
fix/destination_elder: check owner for unpublished idata

octol:
Part of #691

### DIFF
--- a/src/destination_elder.rs
+++ b/src/destination_elder.rs
@@ -23,8 +23,8 @@ use idata_op::IDataOp;
 use log::{error, trace, warn};
 use pickledb::PickleDb;
 use safe_nd::{
-    AccountData, Error as NdError, IDataAddress, IDataKind, MessageId, NodePublicId, Request,
-    Response, Result as NdResult, XorName,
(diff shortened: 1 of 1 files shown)


src/destination_elder.rs
```diff
@@ -474,16 +474,34 @@ impl DestinationElder {
...
         let result = self
             .immutable_chunks
             .get(&address)
             .map_err(|error| error.to_string().into())
-            .map(|kind| {
-                if !kind.published() {
-                    // TODO - Verify ownership
+            .and_then(|kind| match kind {
+                IDataKind::Unpub(ref data) => {
+                    if &XorName::from(PublicKey::from(*data.owners())) != client {
```
Fraser999: Isn't `data.owners()` already a `PublicKey`?

octol: It's a `threshold_crypto::PublicKey`

nbaksalyar: Here's an issue to fix this in safe-nd: https://github.com/maidsafe/safe_client_libs/issues/882
We can fix this in the vault code once that's changed in safe-nd.

PR #733, opened 2019-07-04
fix/destination_elder: verify retreived data is the expected kind

octol:
Part of #691


octol:
With the change of making published and unpublished storable on the same address this is now unnecessary

PR #734, opened 2019-07-04
Implement UpdateAccount RPC

nbaksalyar:
Also add more tests for accounts (checking the permissions).


src/destination_elder.rs
```diff
@@ -319,6 +320,42 @@ impl DestinationElder {
...
+            .and_then(|existing_account| {
+                if XorName::from(*existing_account.authorised_getter()) != src {
+                    // Request does not come from the owner
+                    Err(NdError::AccessDenied)
+                } else {
+                    self.account_chunks
+                        .put(&AccountData::new(
+                            *existing_account.destination(),
+                            *existing_account.authorised_getter(),
+                            updated_account.data().to_vec(),
```
Fraser999: I was hoping we could just do `self.account_chunks.put(updated_account)` which avoids cloning all the fields.  We already know the destination fields match, and if the authorised_getter has changed shouldn't we honour that by writing the updated version, or if we don't want to support changing it, by returning an error rather than partially updating the info?

src/destination_elder.rs
```diff
@@ -329,14 +366,17 @@ impl DestinationElder {
...
         } else {
-            self.account_chunks
-                .put(&Account {
-                    address: *new_account.destination(),
-                    data: new_account.data().to_vec(), // FIXME: we probably don't need cloning here, try mem::replace
-                    authorised_getter: *new_account.authorised_getter(),
-                    signature: new_account.signature().clone(),
-                })
-                .map_err(|error| error.to_string().into())
+            AccountData::new(
```
Fraser999: Here too, can't we just do `self.account_chunks.put(new_account)`?

tests/integration.rs
```diff
@@ -135,6 +136,108 @@ fn accounts() {
...
+
+    // Try to get the account data and signature.
+    let message_id = client.send_request(conn_info.clone(), Request::GetAccount(account_locator));
+    env.poll(&mut vault);
+
+    match client.expect_response(message_id) {
+        Response::GetAccount(Ok((data, sig))) => {
+            assert_eq!(data, new_account_data);
+
+            match client.public_id().public_key().verify(&sig, &data) {
```
Fraser999: This could just be `unwrap!(client.public_id().public_key().verify(&sig, &data));`

PR #740, opened 2019-07-05
Handle CreateBalance and TransferCoins

### LINKED ISSUE #708: Handle TransferCoins

### LINKED ISSUE #729: Handle CreateCoinBalance

madadam:
Closes #729 
Closes #708

### DIFF
--- a/src/source_elder.rs
+++ b/src/source_elder.rs
@@ -6,6 +6,9 @@
 // KIND, either express or implied. Please review the Licences for the specific language governing
 // permissions and limitations relating to use of the SAFE Network Software.
 
+mod balance;
+
+use self::balance::BalanceDb;
 use crate::{
     action::Action,
     quic_p2p::{self, Config as QuicP2pConfig, Event, NodeInfo, Peer, QuicP2p},
@@ -19,8 +22,8 @@ use lazy_static::lazy_static;
 use log::{error, info, trace, warn};
 use pickledb::PickleDb;
 use safe_nd::{
-    AppPermissions, Challenge, ClientPublicId, Coins, Error as NdError, IDataKind, Message,
-    MessageId, NodePublicId, PublicId, PublicKey, Request, Response, Signature, XorName,
+    AppPermissions, Challenge, Coins, Error as NdError, IDataKind, Message, MessageId,
+    NodePublicId, PublicId, PublicKey, Request, Response, Signature, XorName,
 };
 use serde::{Deserialize, Serialize};
 use std::{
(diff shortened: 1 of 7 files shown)


src/source_elder.rs
```diff
@@ -553,6 +586,95 @@ impl SourceElder {
...
+        let result = self
+            .withdraw_coins_for_transfer(client_id, amount)
+            .and_then(|cost| {
+                self.deposit(&destination, amount).map_err(|error| {
+                    self.refund(client_id, cost);
+                    error
+                })
+            });
+
+        self.send_response_to_client(public_id, message_id, Response::Mutation(result));
```
nbaksalyar: I think the `TransferCoins` should not return any response to a client at all?
The client is supposed to check the status of the transaction by polling `GetTransaction`.
But I guess this can be handled in a separate PR.

src/source_elder.rs
```diff
@@ -553,6 +586,95 @@ impl SourceElder {
...
+        let result = self
+            .withdraw_coins_for_transfer(client_id, amount)
+            .and_then(|cost| {
+                self.create_balance(owner_key, amount).map_err(|error| {
+                    self.refund(client_id, cost);
+                    error
+                })
+            });
+
+        self.send_response_to_client(public_id, message_id, Response::Mutation(result));
```
nbaksalyar: Same for `CreateBalance`.

src/source_elder.rs
madadam: Is there a reason why not? I think removing it would make the API more cumbersome to use.

tests/integration.rs
```diff
@@ -236,16 +214,80 @@ fn update_account() {
...
+            amount: unwrap!(Coins::from_nano(10)),
+            transaction_id: 0,
+        },
+    );
+    env.poll(&mut vault);
+
+    let balance = common::get_balance(&mut env, &mut client_a, &mut vault);
+    assert_eq!(balance, unwrap!(Coins::from_nano(10)));
+
+    // Create B'a balance
```
Fraser999: Typo: "B's"

tests/integration.rs
```diff
@@ -236,16 +214,80 @@ fn update_account() {
     }
 }
 
+#[test]
+fn coin_operations() {
```
Fraser999: Maybe worth also checking that client A can't retrieve client B's balance?

madadam: `GetBalance` doesn't specify which balance to get, it's always the sender's one. So not sure how would I even construct such request?

Fraser999: Aha - cunning of us to already make that impossible! :D

Issue #692, opened 2019-06-13
Handle DeleteUnpubIData


Fraser999:

Lindsey17:
Resolved by PR #741

PR #741, opened 2019-07-05
Add DeleteUnpubIData

octol:
Adds `DeleteUnpubIData`.

Left to do:

1. `handle_mutation_resp` needs to be also updated
2. Duplication with the other immutable data handlers should be dealt with

Part of #692

### DIFF
--- a/src/source_elder.rs
+++ b/src/source_elder.rs
@@ -259,7 +259,31 @@ impl SourceElder {
                     None
                 }
             }
-            DeleteUnpubIData(ref address) => unimplemented!(),
+            DeleteUnpubIData(ref address) => {
+                if address.published() {
+                    self.send_response_to_client(
+                        client_id,
+                        message_id,
+                        // TODO: consider changing this error
+                        Response::GetIData(Err(NdError::InvalidOperation)),
+                    );
+                    return None;
+                }
+                if registered_client == ClientState::Registered {
+                    Some(Action::ForwardClientRequest {
+                        client_name: *client_id.name(),
+                        request,
+                        message_id,
+                    })
+                } else {
+                    self.send_response_to_client(
+                        client_id,
+                        message_id,
+                        Response::Mutation(Err(NdError::AccessDenied)),
+                    );
+                    None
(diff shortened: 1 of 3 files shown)


src/source_elder.rs
```diff
@@ -241,6 +241,7 @@ impl SourceElder {
                 })
             }
             GetIData(ref address) => {
+                dbg!(&registered_client);
```
Fraser999: I guess this was left in by mistake?

src/source_elder.rs
```diff
@@ -259,7 +260,31 @@ impl SourceElder {
...
+                    Some(Action::ForwardClientRequest {
+                        client_name: *client_id.name(),
+                        request,
+                        message_id,
+                    })
+                } else {
+                    self.send_response_to_client(
+                        client_id,
+                        message_id,
+                        Response::GetIData(Err(NdError::AccessDenied)),
```
Fraser999: This should be the `Mutation` variant.

src/source_elder.rs
```diff
@@ -426,6 +451,7 @@ impl SourceElder {
         if let Some(challenge) = self.client_candidates.remove(&peer_addr) {
             match public_key.verify(&signature, challenge) {
                 Ok(()) => {
+                    dbg!(&public_id);
```
Fraser999: This should be removed too I think.

tests/integration.rs
```diff
@@ -172,3 +172,40 @@ fn get_immutable_data() {
 
     // TODO - Get immutable data that exist once we have PutIData working
 }
+
+#[test]
+fn delete_unpublished_immutable_data() {
+    env_logger::init();
```
Fraser999: This should get removed too please.  @nbaksalyar will soon be adding a commit to enable logging inside `Environment::new()`, and this instance will panic if another test has already initialised the logger.

src/source_elder.rs
octol: Oopsie, good thing we have code review ;)

PR #742, opened 2019-07-05
Vault-level RPCs

nbaksalyar:
This change is to have full information about a requester (if it's an app) at the destination elders.

### DIFF
--- a/src/destination_elder.rs
+++ b/src/destination_elder.rs
@@ -14,6 +14,7 @@ use crate::{
         error::Error as ChunkStoreError, AppendOnlyChunkStore, ImmutableChunkStore,
         LoginPacketChunkStore, MutableChunkStore,
     },
+    rpc::Rpc,
     utils,
     vault::Init,
     Result, ToDbKey,
@@ -22,8 +23,8 @@ use idata_op::{IDataOp, OpType};
 use log::{error, trace, warn};
 use pickledb::PickleDb;
 use safe_nd::{
-    Error as NdError, IData, IDataAddress, LoginPacket, MessageId, NodePublicId, Request, Response,
-    Result as NdResult, XorName,
+    Error as NdError, IData, IDataAddress, LoginPacket, MessageId, NodePublicId, PublicId, Request,
+    Response, Result as NdResult, XorName,
 };
 use serde::{Deserialize, Serialize};
 use std::{
@@ -118,19 +119,36 @@ impl DestinationElder {
         })
     }
 
-    pub fn handle_request(
+    pub fn handle_vault_message(&mut self, src: XorName, message: Rpc) -> Option<Action> {
+        match message {
+            Rpc::Request {
+                request,
+                requester,
+                message_id,
+            } => self.handle_request(src, requester, request, message_id),
+            Rpc::Response {
+                response,
+                message_id,
+                ..
+            } => self.handle_response(src, response, message_id),
+        }
+    }
+
+    fn handle_request(
         &mut self,
         src: XorName,
+        requester: PublicId,
         request: Request,
         message_id: MessageId,
     ) -> Option<Action> {
         use Request::*;
         trace!(
-            "{}: Received ({:?} {:?}) from {}",
+            "{}: Received ({:?} {:?}) from src {} (client {:?})",
             self,
             request,
             message_id,
-            src
+            src,
+            requester
         );
         // TODO - remove this
         #[allow(unused)]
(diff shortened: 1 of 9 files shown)


src/destination_elder.rs
```diff
@@ -319,10 +332,19 @@ impl DestinationElder {
...
     ) -> Option<Action> {
+        let requester_pk = match requester {
+            PublicId::Client(ref client) => client.public_key(),
+            PublicId::App(ref app) => app.public_key(),
+            PublicId::Node(..) => {
+                // TODO: drop the connection, possible malice
+                warn!("Unexpected PublicId::Node requester: {:?}", requester);
+                return None;
+            }
+        };
```
madadam: Might be worth extracting this into a function. The same snippet is used in https://github.com/maidsafe/safe_vault/blob/315c3a88c49447ac98e51326d53fba24259d1dec/src/source_elder.rs#L378-L385, https://github.com/maidsafe/safe_vault/blob/315c3a88c49447ac98e51326d53fba24259d1dec/src/source_elder.rs#L472-L483 and possibly elsewhere.

src/destination_elder.rs
```diff
@@ -369,18 +391,27 @@ impl DestinationElder {
...
```
madadam: Ditto.

src/destination_elder.rs
```diff
@@ -628,14 +670,23 @@ impl DestinationElder {
...
+        let client = self.client_id(&message_id)?;
+        let client_pk = match client {
+            PublicId::Client(ref client) => client.public_key(),
+            PublicId::App(ref app) => app.public_key(),
+            PublicId::Node(..) => {
+                // TODO: drop the connection, possible malice
+                warn!("Unexpected PublicId::Node requester: {:?}", client);
+                return None;
+            }
+        };
```
madadam: Ditto

src/destination_elder.rs
```diff
@@ -652,16 +703,26 @@ impl DestinationElder {
...
-        if &src == address.name() {
+        let requester_pk = match requester {
+            PublicId::Client(ref client) => client.public_key(),
+            PublicId::App(ref app) => app.public_key(),
+            PublicId::Node(..) => {
+                // TODO: drop the connection, possible malice
+                warn!("Unexpected PublicId::Node requester: {:?}", requester);
+                return None;
+            }
+        };
```
madadam: Ditto

src/destination_elder.rs
```diff
@@ -698,15 +762,24 @@ impl DestinationElder {
...
+        let client = self.client_id(&message_id)?;
+        let client_pk = match client {
+            PublicId::Client(ref client) => client.public_key(),
+            PublicId::App(ref app) => app.public_key(),
+            PublicId::Node(..) => {
+                // TODO: drop the connection, possible malice
+                warn!("Unexpected PublicId::Node requester: {:?}", client);
+                return None;
+            }
+        };
```
madadam: Ditto

src/destination_elder.rs
nbaksalyar: Yeah, I thought that it probably should be a part of `safe-nd`, but it's not very clear how to handle `PublicId::App` - there are no universal rules.

So maybe it should be 2 separate functions, like `PublicId::owner_public_key` and `PublicId::own_public_key`? @Fraser999 what's your opinion on this?

madadam: :+1: for the two functions.

madadam: Don't think it needs to go to safe-nd necessarily. Unless the client libs have a use for it too?

Fraser999: Yeah, I'd keep it out of safe-nd if possible.  But +1 from me on the two helpers.  The main downside I see is that the `Node` variant should return `None` for our use cases, but the name of the helpers doesn't really suggest that.

src/destination_elder.rs
```diff
@@ -118,9 +119,19 @@ impl DestinationElder {
...
+        match message {
+            VaultMessage::ClientRequest {
+                request,
+                requester,
+                message_id,
+            } => self.handle_request(requester, request, message_id),
+        }
+    }
+
     pub fn handle_request(
```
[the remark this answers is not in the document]
Fraser999: It does seem like we should choose between handling safe_nd-level messages or vault-level ones in the persona APIs.

src/destination_elder.rs
```diff
@@ -787,6 +852,6 @@ impl DestinationElder {
 
 impl Display for DestinationElder {
     fn fmt(&self, formatter: &mut Formatter) -> fmt::Result {
-        write!(formatter, "{}", self.id.name())
+        write!(formatter, "{}", self.id)
```
Fraser999: Did you mean to revert this recent change?

PR #744, opened 2019-07-07
Implement delete unpublished immutable data response

octol:
Part of #692

The first commit implements the delete response, and then the other commits are purely grouping methods together. So it's probably easier to review by looking at each commit separately.

### DIFF
--- a/src/destination_elder.rs
+++ b/src/destination_elder.rs
@@ -20,7 +20,7 @@ use crate::{
     Result, ToDbKey,
 };
 use idata_op::{IDataOp, OpType};
-use log::{error, trace, warn};
+use log::{error, info, trace, warn};
 use pickledb::PickleDb;
 use safe_nd::{
     Error as NdError, IData, IDataAddress, LoginPacket, MessageId, NodePublicId, PublicId, Request,
@@ -747,6 +747,11 @@ impl DestinationElder {
         message_id: MessageId,
     ) -> Option<Action> {
         let result = if self.immutable_chunks.has(kind.address()) {
+            info!(
+                "{}: Immutable chunk already exists, not storing: {:?}",
+                self,
+                kind.address()
+            );
             Ok(())
         } else {
             self.immutable_chunks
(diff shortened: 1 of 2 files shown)


src/destination_elder.rs
```diff
@@ -498,6 +498,61 @@ impl DestinationElder {
...
+                    sender: *address.name(),
+                    message: Rpc::Response {
+                        requester: client_id,
+                        response: Response::Mutation(result),
+                        message_id,
+                    },
+                })
+            };
+
+            // We're acting as dst elder, received request from src elders
```
Fraser999: This comment has been duplicated - suggest removing this one.

src/destination_elder.rs
```diff
@@ -568,38 +623,54 @@ impl DestinationElder {
...
+                    self, idata_address
+                );
+                None
+            });
 
-    // Returns an iterator over all of our section's elders' names, sorted by closest to `target`.
-    fn elders_sorted(&self, _target: &XorName) -> impl Iterator<Item = &XorName> {
-        iter::once(self.id.name())
+        if let Some(mut metadata) = metadata {
+            if metadata.holders.take(&sender).is_none() {
```
Fraser999: I guess I'd have used `if !metadata.holders.remove(&sender) {` rather than `take()` since we don't need that returned entry.  Just a nitpick though :)

src/destination_elder.rs
```diff
@@ -654,6 +727,54 @@ impl DestinationElder {
...
+    }
+
+    fn store_idata(
+        &mut self,
+        kind: IData,
+        requester: PublicId,
+        message_id: MessageId,
+    ) -> Option<Action> {
+        let result = if self.immutable_chunks.has(kind.address()) {
+            Ok(())
```
Fraser999: Might be worth putting an info-level log line here, since we shouldn't normally be receiving such a request.  The elders should be responding with success to requests to store duplicate chunks by checking their metadata, and not actually forwarding the Put request to individual adults/elders.  There might be cases where they do actually send a chunk out to an adult/elder which already has it stored, so might be useful to see a log line in that case.  It could represent potentially large messages which didn't need to have been sent.

PR #745, opened 2019-07-08
Allow overlapping ChunkStore XorNames

### LINKED ISSUE #739: Change ChunkStore handling of chunks with same XorNames
It's been decided to change the chunkstore to allow multiple chunks of different types but with identical XorNames to co-exist.

Hopefully this should be a simple matter of changing the `file_path()` method to not use `id.raw_name()`, but rather just `id`.  I expect we can get rid of the `raw_name()` method from the `ChunkId` trait, and replace it with a `Serialize` bounds on the trait.

octol:
Resolves #739

### DIFF
--- a/src/chunk_store/chunk.rs
+++ b/src/chunk_store/chunk.rs
@@ -15,13 +15,6 @@ pub(crate) trait Chunk: Serialize + DeserializeOwned {
     fn id(&self) -> &Self::Id;
 }
 
-pub(crate) trait ChunkId: ToDbKey + PartialEq + Eq {
-    /// The XorName represented by this identifier.
-    fn raw_name(&self) -> &XorName;
-}
+pub(crate) trait ChunkId: ToDbKey + PartialEq + Eq {}
 
-impl ChunkId for XorName {
-    fn raw_name(&self) -> &XorName {
-        self
-    }
-}
+impl ChunkId for XorName {}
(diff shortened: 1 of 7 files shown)


src/chunk_store/chunk.rs
```diff
@@ -15,13 +15,6 @@ pub(crate) trait Chunk: Serialize + DeserializeOwned {
     fn id(&self) -> &Self::Id;
 }
 
-pub(crate) trait ChunkId: ToDbKey + PartialEq + Eq {
-    /// The XorName represented by this identifier.
-    fn raw_name(&self) -> &XorName;
-}
+pub(crate) trait ChunkId: ToDbKey + PartialEq + Eq + Serialize {}
```
madadam: `ToDbKey` already implies `Serialize`, so I don't think it needs to be repeated here.

tests/integration.rs
```diff
@@ -272,6 +275,89 @@ fn get_immutable_data() {
...
+        Request::GetIData(IDataAddress::Pub(address)),
+    );
+    env.poll(&mut vault);
+    match client.expect_response(message_id) {
+        Response::GetIData(Ok(IDataKind::Pub(idata))) => assert_eq!(idata.value(), &value),
+        x => unexpected!(x),
+    }
+
+    let message_id = client.send_request(
+        conn_info.clone(),
```
madadam: Don't need to `clone` - `conn_info` is not used after this.

PR #747, opened 2019-07-08
Adapt Vaults to latest changes in safe-nd API

nbaksalyar:


src/utils.rs
```diff
@@ -119,12 +118,12 @@ pub(crate) fn dst_elders_address(request: &Request) -> Option<&XorName> {
 
 /// Temporary helpers to work around changes required in safe-nd.
 pub(crate) mod work_arounds {
-    use safe_nd::{IDataAddress, IDataKind};
+    use safe_nd::{IData, IDataAddress};
```
Fraser999: Don't these changes in safe-nd allow us to remove this workaround altogether now?

PR #750, opened 2019-07-09
Send response on CreateBalance and TransferCoins

madadam:


src/source_elder.rs
```diff
@@ -630,36 +633,50 @@ impl SourceElder {
...
-        let cost = self
+        let result = self
             .withdraw_coins_for_transfer(public_id.name(), amount)
-            .ok()?;
-
-        if self.create_balance(owner_key, amount).is_err() {
-            self.refund(public_id.name(), cost)
-        }
+            .and_then(|cost| {
+                self.create_balance(owner_key, amount).map_err(|error| {
```
Fraser999: We should really be breaking the flow by returning an Action here to get Vaults to handle sending this to the elders at `owner_key`.  When they do that, they'll also check if this vault is a member of the target section and pass the message back in to this persona.

Fraser999: Agreed to try and get this resolved before phase 1 is complete.

PR #752, opened 2019-07-11
Implement basic MData RPCs

### LINKED ISSUE #693: Handle PutMData

### LINKED ISSUE #694: Handle GetUnseqMData, GetUnseqMDataShell, GetUnseqMDataValue, ListUnseqMDataEntries and ListUnseqMDataValues

### LINKED ISSUE #697: Handle GetSeqMData, GetSeqMDataShell, GetSeqMDataValue, ListSeqMDataEntries and ListSeqMDataValues

### LINKED ISSUE #699: Handle GetMDataVersion, ListMDataKeys, ListMDataPermissions, ListMDataUserPermissions

nbaksalyar:
Closes #693, closes #699, closes #694, closes #697

### DIFF
--- a/src/destination_elder.rs
+++ b/src/destination_elder.rs
@@ -23,9 +23,9 @@ use idata_op::{IDataOp, OpType};
 use log::{error, info, trace, warn};
 use pickledb::PickleDb;
 use safe_nd::{
-    Error as NdError, IData, IDataAddress, LoginPacket, MData, MDataAddress, MDataPermissionSet,
-    MDataSeqEntryActions, MDataUnseqEntryActions, MessageId, NodePublicId, PublicId, PublicKey,
-    Request, Response, Result as NdResult, XorName,
+    Error as NdError, IData, IDataAddress, LoginPacket, MData, MDataAction, MDataAddress,
+    MDataPermissionSet, MDataSeqEntryActions, MDataUnseqEntryActions, MessageId, NodePublicId,
+    PublicId, PublicKey, Request, Response, Result as NdResult, XorName,
 };
 use serde::{Deserialize, Serialize};
 use std::{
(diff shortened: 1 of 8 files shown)


src/destination_elder.rs
```diff
@@ -331,6 +375,428 @@ impl DestinationElder {
...
+        address: MDataAddress,
+        actions: MDataSeqEntryActions,
+        message_id: MessageId,
+    ) -> Option<Action> {
+        let requester_pk = *utils::own_key(&requester)?;
+
+        self.mutate_mdata_chunk(&address, requester, message_id, move |mut data| {
+            match data {
+                MData::Seq(ref mut mdata) => mdata.mutate_entries(actions, requester_pk)?,
+                MData::Unseq(..) => return Err(NdError::NoSuchData), // FIXME
```
Fraser999: It's not clear how this should be fixed.  This represents a logic error again I think, so I think the flow should stop there, or if we need to send a response, it should probably be the `NetworkOther` since it's a problem with vault logic.

Same comment for the other "FIXME"s in this PR.

tests/integration.rs
```diff
@@ -509,3 +512,567 @@ fn delete_unpublished_immutable_data() {
...
+    let message_id = client.send_request(
+        conn_info.clone(),
+        Request::PutMData(MData::Unseq(mdata.clone())),
+    );
+    env.poll(&mut vault);
+
+    match client.expect_response(message_id) {
+        Response::Mutation(Ok(())) => (),
+        x => unexpected!(x),
+    }
```
octol: For these mutation requests I noticed there is a `common::perform_mutation`

octol:
Quite a few `handle_` methods now, in `destination_elder` in particular. I think it would be easier to navigate if they were ordered the same as the `match` blocks. Alternatively is it time to move idata/mdata/adata to their own files?

src/destination_elder.rs
```diff
@@ -331,6 +369,464 @@ impl DestinationElder {
...
+        })
+    }
+
+    fn handle_delete_mdata_req(
+        &mut self,
+        requester: PublicId,
+        address: MDataAddress,
+        message_id: MessageId,
+    ) -> Option<Action> {
+        let requester_pk = *utils::own_key(&requester)?;
```
Fraser999: Shouldn't we try and be consistent for all these similar cases?  For `get_mdata_chunk`, we print an error message, which does seem like it'd be useful for all these cases.

However, maybe we could just print a more general error at the end of `handle_request` if any of them return `None`?  Or if there are cases where that's not an error, how about an `Action` variant indicating a won't-handle-this-request error, and return that instead of `None`, and log an error if we get that returned?

src/source_elder.rs
```diff
@@ -380,6 +386,90 @@ impl SourceElder {
...
+        &mut self,
+        request: Request,
+        client: &ClientInfo,
+        message_id: MessageId,
+        signature: Option<Signature>,
+    ) -> Option<Action> {
+        let owner = utils::owner(&client.public_id)?;
+        self.has_signature(&client.public_id, &request, &message_id, &signature)?;
+        if let Err(error) = self.withdraw(owner.public_key(), *COST_OF_PUT) {
+            // Note: in phase 1, we proceed even if there are insufficient funds.
```
Fraser999: It might be worth moving this comment and log line to inside `fn withdraw` and have it always return `Ok(())` for now to avoid duplicating this in a few places?

src/destination_elder.rs
```diff
@@ -331,6 +369,464 @@ impl DestinationElder {
...
+
+        self.mutate_mdata_chunk(&address, requester, message_id, move |mut data| {
+            match data {
+                MData::Seq(ref mut mdata) => mdata.mutate_entries(actions, requester_pk)?,
+                MData::Unseq(..) => {
+                    error!("Logic error - unexpected chunk stored at {:?}", address);
+                    return Err(NdError::NetworkOther(
+                        "Logic error - unexpected chunk".to_string(),
+                    ));
+                }
```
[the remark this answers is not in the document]
Fraser999: @octol Yes, you're right.  I was assuming this was checked and caught at the source elders, but it isn't.  I think it should be validated there (as much validation as possible should happen as early in the flow as possible I feel).  If we do that, then this would represent a logic error.

@madadam I think since we're not allowed to handle logic errors by panicking, then returning a "Logic error" to the client will help us identify the issue.  I see it as similar to a "500 Internal Server Error" http response.  Regardless of what we send back, there's nothing the client can really do to fix it, so sending back a message which implies this seems the best we can do.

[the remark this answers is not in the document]
madadam: OK, here is an idea: instead of having only `get` in the mutable store, have both `get_seq` and `get_unseq` where each would return the corresponding type (`SeqMutableData` or `UnseqMutableData`) and `Err(NoSuchData)` in case of mismatch. Then we wouldn't have to handle the mismatches all the time. Probably something for future PRs.

madadam:
Superseded by https://github.com/maidsafe/safe_vault/pull/762

PR #754, opened 2019-07-12
Handle Auth Keys

### LINKED ISSUE #710: Handle InsAuthKey, DelAuthKey and ListAuthKeysAndVersion

Fraser999:
Closes #710

### DIFF
--- a/src/source_elder.rs
+++ b/src/source_elder.rs
@@ -6,28 +6,31 @@
 // KIND, either express or implied. Please review the Licences for the specific language governing
 // permissions and limitations relating to use of the SAFE Network Software.
 
+mod account;
 mod balance;
 
-use self::balance::{Balance, BalanceDb};
+use self::{
+    account::AccountsDb,
+    balance::{Balance, BalancesDb},
+};
 use crate::{
     action::Action,
     quic_p2p::{self, Config as QuicP2pConfig, Event, NodeInfo, Peer, QuicP2p},
     rpc::Rpc,
     utils,
     vault::Init,
-    Error, Result, ToDbKey,
+    Error, Result,
 };
 use bytes::Bytes;
 use crossbeam_channel::{self, Receiver};
 use lazy_static::lazy_static;
 use log::{error, info, trace, warn};
-use pickledb::PickleDb;
 use safe_nd::{
     AppPermissions, Challenge, Coins, Error as NdError, IData, IDataAddress, IDataKind, Message,
     MessageId, NodePublicId, PublicId, PublicKey, Request, Response, Signature, Transaction,
     TransactionId, XorName,
 };
-use serde::{Deserialize, Serialize};
+use serde::Serialize;
 use std::{
     collections::HashMap,
     fmt::{self, Display, Formatter},
@@ -36,16 +39,9 @@ use std::{
 };
 use unwrap::unwrap;
 
-const ACCOUNTS_DB_NAME: &str = "accounts.db";
-
 lazy_static! {
     /// The cost to Put a chunk to the network.
     pub static ref COST_OF_PUT: Coins = unwrap!(Coins::from_nano(1_000_000_000));
-}
-
-#[derive(Serialize, Deserialize, Debug)]
-struct Account {
-    apps: HashMap<PublicKey, AppPermissions>,
 }
 
 #[derive(Clone, Debug)]
@@ -56,8 +52,8 @@ struct ClientInfo {
 
 pub(crate) struct SourceElder {
     id: NodePublicId,
-    accounts: PickleDb,
-    balances: BalanceDb,
+    accounts: AccountsDb,
+    balances: BalancesDb,
     clients: HashMap<SocketAddr, ClientInfo>,
     // Map of new client connections to the challenge value we sent them.
     client_candidates: HashMap<SocketAddr, Vec<u8>>,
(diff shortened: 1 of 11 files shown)


tests/integration.rs
```diff
@@ -509,3 +510,165 @@ fn delete_unpublished_immutable_data() {
...
+        Response::Mutation(Err(NdError::NoSuchBalance)) => (),
+        x => unexpected!(x),
+    }
+
+    message_id = owner.send_request(conn_info.clone(), Request::ListAuthKeysAndVersion);
+    env.poll(&mut vault);
+    match owner.expect_response(message_id) {
+        Response::ListAuthKeysAndVersion(Err(NdError::NoSuchBalance)) => (),
+        x => unexpected!(x),
+    }
```
octol: Would this be nice?
```
assert_eq!(
    common::request_expect_err(... , Request::ListAuthKeysAndVersion)), 
    NdError::NoSuchBalance
)
```

octol: The difficult part would be destructuring the reponse, but I guess it's fine to also include it in the return value
```
assert_eq!(
    common::request_expect_err(... , Request::ListAuthKeysAndVersion)), 
    Response::ListAuthKeysAndVersion(NdError::NoSuchBalance),
)
```

src/source_elder.rs
```diff
@@ -459,22 +583,20 @@ impl SourceElder {
         client: &ClientInfo,
         address: IDataAddress,
         message_id: MessageId,
-        signature: Option<Signature>,
     ) -> Option<Action> {
         if address.kind() == IDataKind::Pub {
             self.send_response_to_client(
                 &client.public_id,
                 message_id,
+                // TODO: consider changing this error
```
octol: Is this accidentally reinserting an old comment?

tests/integration.rs
octol: Alternatively just a more compact expression (macro?) to replace the `match expect_response` block. There are quite a few lines of `x => unexpected!(x)` now in that file...

src/utils.rs
```diff
@@ -128,3 +140,11 @@ pub(crate) fn dst_elders_address(request: &Request) -> Option<&XorName> {
...
     }
 }
+
+pub(crate) fn is_published(address: &ADataAddress) -> bool {
+    use ADataAddress::*;
+    match address {
+        PubSeq { .. } | PubUnseq { .. } => true,
+        UnpubSeq { .. } | UnpubUnseq { .. } => false,
+    }
+}
```
octol: Hey convenient :)
I created the same function in my branch, now I can get rid of that!
(This should go in `safe-nd` though right? If so should this be in a `workaround` module?)

Fraser999: Good point!

src/source_elder/account.rs
```diff
@@ -0,0 +1,126 @@
...
+use safe_nd::{
+    AppPermissions, AppPublicId, ClientPublicId, Error as NdError, PublicKey, Result as NdResult,
+};
+use serde::{Deserialize, Serialize};
+use std::{
+    collections::{BTreeMap, HashMap},
+    path::Path,
+};
+
+type AccountAsTuple = (BTreeMap<PublicKey, AppPermissions>, u64);
```
madadam: Why do we need this? Why not use `Account` directly?

tests/common/mod.rs
```diff
@@ -132,52 +149,48 @@ impl AsMutSlice<TestVault> for [TestVault] {
...
+        match self {
+            FullId::Client(full_id) => PublicId::Client(full_id.public_id().clone()),
+            FullId::App(full_id) => PublicId::App(full_id.public_id().clone()),
+        }
     }
+}
 
-    pub fn full_id(&self) -> &ClientFullId {
-        &self.full_id
+pub trait TestClientTrait {
```
madadam: What about instead of having client and app represented as two distinct types, have only one type with two constructors, say: `TestClient::disconnected_client` and `TestClient::disconnected_app`? That way we could avoid lot of unnecessary complexity (traits, generics) and duplication (the `TestClientTrait` impls of both `TestClient` and `TestApp` are pretty much identical).

src/source_elder/account.rs
madadam: Of course. OK.

tests/common/mod.rs
Fraser999: Yeah, I had deliberately avoided that so that there's less chance of a test using the wrong type.  For example, if someone later creates a TestClient via `disconnected_client()` and labels the variable `app`, it could be less apparent that we're testing the wrong thing than if these were separate types.  It also lets the client hold the specific ID type appropriate to it, where as if we join them into a single type, we'd have to fall back to returning the `PublicId` enum.

Maybe it's overkill though?

madadam: OK, I take your points. Still feels a bit overkill but I won't block the PR over it. We can possibly address this in a later PR if we think it's worth it.

Fraser999: Agreed.  It should be fairly trivial to change this - I think the biggest hit would just be in how we'd handle changing from the specific `PublicId` types.  E.g. return the top-level enum, or maybe safer to change to have two getters like:
```rust
fn client_public_id(&self) -> &ClientPublicId {
    match &self.public_id {
        Client(id) => id
        _ => panic!()
    }
}
```

PR #756, opened 2019-07-15
Add PutAData and DeleteAData

### LINKED ISSUE #702: Handle PutAData

octol:
Add `PutAData` and `DeleteAData`.

Closes #702, #705

### DIFF
--- a/tests/integration.rs
+++ b/tests/integration.rs
@@ -295,46 +295,76 @@ fn put_append_only_data() {
     let tag = 100;
     let mut adata = PubSeqAppendOnlyData::new(adata_name, tag);
     unwrap!(adata.append(vec![(b"more".to_vec(), b"data".to_vec())], 0));
-    common::perform_mutation(
-        &mut env,
-        &mut client,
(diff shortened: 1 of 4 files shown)


tests/integration.rs
```diff
@@ -322,6 +323,78 @@ fn coin_operations() {
...
+    common::perform_mutation(
+        &mut env,
+        &mut client,
+        &mut vault,
+        Request::PutAData(AData::UnpubUnseq(adata)),
+    );
+
+    // TODO - get the data to verify
+}
+
```
madadam: Would be good to also add test for the various failure modes of `PutAData` (data exists, access denied, ...). Can be done in a followup PR if you prefer.

octol: Yep I agree, both Put and Delete needs a bit more systematic testing of the non-happy paths. I'd prefer to do that in a follow-up PR so as not to block other PRs

madadam: OK, let's do it in a follow-up then.

PR #759, opened 2019-07-17
chore/all: refactored LoginPacket flows

Fraser999:
This changes the handler for login packets from `DestinationElder` to `SourceElder`.

It meant having `SourceElder` updated to handle Vault-level requests, since the flow for `CreateLoginPacket` involves sending from the client's elders to the elders of `login_packet.destination()`.

### DIFF
--- a/src/source_elder.rs
+++ b/src/source_elder.rs
@@ -15,11 +15,12 @@ use self::{
 };
 use crate::{
     action::Action,
+    chunk_store::{error::Error as ChunkStoreError, LoginPacketChunkStore},
     quic_p2p::{self, Config as QuicP2pConfig, Event, NodeInfo, Peer, QuicP2p},
     rpc::Rpc,
     utils,
     vault::Init,
-    Error, Result,
+    Config, Error, Result,
 };
 use bytes::Bytes;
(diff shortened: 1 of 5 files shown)


src/source_elder.rs
```diff
@@ -743,6 +771,81 @@ impl SourceElder {
...
+        trace!(
+            "{}: Received ({:?} {:?}) from src {} (client {:?})",
+            self,
+            request,
+            message_id,
+            src,
+            requester
+        );
+        // TODO - remove this
+        #[allow(unused)]
```
madadam: I think this can be removed now

src/source_elder.rs
```diff
@@ -743,6 +771,81 @@ impl SourceElder {
...
+            | AppendUnseq(_)
+            | TransferCoins { .. }
+            | GetBalance
+            | ListAuthKeysAndVersion
+            | InsAuthKey { .. }
+            | DelAuthKey { .. }
+            | CreateBalance { .. }
+            | CreateLoginPacketFor { .. }
+            | UpdateLoginPacket { .. }
+            | GetLoginPacket(..) => {
```
madadam: Can we do a wildcard match here?

PR #763, opened 2019-07-19
feat/all: use try_from in integration tests

Fraser999:
This also updates quic-p2p, but the mock's behaviour hasn't been changed to reflect the real quic-p2p behaviour regarding the new `Token` being passed in `send()`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,20 +17,20 @@ bincode = "=1.1.4"
 bytes = { version = "~0.4.12", features = ["serde"] }
 crossbeam-channel = "~0.3.8"
 directories = "~2.0.1"
-env_logger = "~0.6.1"
+env_logger = "~0.6.2"
 fxhash = { version = "~0.2.1", optional = true }
 hex = "~0.3.2"
 hex_fmt = { version = "~0.3.0", optional = true }
 lazy_static = "~1.3.0"
-log = "~0.4.6"
+log = "~0.4.7"
 pickledb = "~0.4.0"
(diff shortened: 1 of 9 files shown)


src/utils.rs
```diff
@@ -159,7 +159,7 @@ pub(crate) fn is_owner<T: AppendOnlyData<ADataUnpubPermissions>>(
     requester: PublicKey,
 ) -> NdResult<()> {
     adata
-        .fetch_owner_at_index(adata.owners_index() - 1)
+        .owner(adata.owners_index() - 1)
```
octol: I'll remove this function later since I'll switch to use the one in safe-nd in my branch

Cargo.toml
```diff
@@ -17,20 +17,20 @@ bincode = "=1.1.4"
...
+env_logger = "~0.6.2"
 fxhash = { version = "~0.2.1", optional = true }
 hex = "~0.3.2"
 hex_fmt = { version = "~0.3.0", optional = true }
 lazy_static = "~1.3.0"
-log = "~0.4.6"
+log = "~0.4.7"
 pickledb = "~0.4.0"
-quic-p2p = { version = "~0.1.1", git = "https://github.com/maidsafe/quic-p2p", optional = true }
+quic-p2p = { version = "~0.2.0", git = "https://github.com/maidsafe/quic-p2p", optional = true }
```
jeanphilippeD: Maybe worth using the crates.io version like routing does to avoid future breakage from change in the master branch.

Issue #758, opened 2019-07-16
Rename elder structs


Fraser999:
We need to make the following changes:
1. Rename `SourceElder` to `ClientHandler`
2. Split `DestinationElder` into `IDataHandler` (deals with the meta-info about `IData`), `IDataHolder` (owns the `IData` chunkstore), `MDataHandler`, and `ADataHandler`.
3. Rename `Account` to `AuthKeys` (ensure there are no instances of the term "account" left in the source code after the change)

madadam:
Closed via #764 and #765.

PR #765, opened 2019-07-21
Split and rename DestinationElder to DataHandler

octol:
Rename DestinationElder to DataHandler and extract out {I,A,M}DataHandler and IDataHolder.

Part of #758


src/data_handler/adata_handler.rs
```diff
@@ -0,0 +1,536 @@
...
+            Rc::clone(total_used_space),
+            init_mode,
+        )?;
+        Ok(Self {
+            id,
+            append_only_chunks,
+        })
+    }
+
+    pub(crate) fn handle_put_adata_req(
```
madadam: Another nitpick: All these methods (and the ones in the other handlers too) can be `pub(super)` as they are only called from `DataHandler` and not above that.

src/lib.rs
```diff
@@ -32,7 +31,6 @@
     unconditional_recursion,
     unknown_lints,
     unsafe_code,
-    unused,
```
madadam: Why remove this?

src/lib.rs
```diff
@@ -17,8 +17,7 @@
     exceeding_bitshifts,
     mutable_transmutes,
     no_mangle_const_items,
-    unknown_crate_types,
-    warnings
```
madadam: Why remove this?

src/vault.rs
```diff
@@ -34,7 +34,7 @@ const STATE_FILENAME: &str = "state";
 enum State {
     Elder {
         src: ClientHandler,
```
madadam: Maybe also rename these fields: `src` -> `client`, `dst` -> `data` ?

src/vault.rs
```diff
@@ -171,8 +171,8 @@ impl Vault {
         match action {
             ForwardClientRequest(message) => {
                 let requester_name = utils::rpc_elder_address(&message)?;
-                let dst_elders_address = if let Rpc::Request { ref request, .. } = message {
-                    match utils::dst_elders_address(&request) {
+                let data_handlers_address = if let Rpc::Request { ref request, .. } = message {
+                    match utils::data_handlers_address(&request) {
                         Some(address) => address,
                         None => {
                             error!("{}: Logic error - no dst address available.", self);
```
madadam: "...no data handler address available" - or something like that.

src/vault.rs
```diff
@@ -185,10 +185,10 @@ impl Vault {
                 };
 
                 // TODO - once Routing is integrated, we'll construct the full message to send
-                //        onwards, and then if we're also part of the dst elders, we'll call that
+                //        onwards, and then if we're also part of the data handlers, we'll call that
                 //        same handler which Routing will call after receiving a message.
 
-                if self.self_is_elder_for(&dst_elders_address) {
+                if self.self_is_elder_for(&data_handlers_address) {
```
madadam: Maybe rename `self_is_elder_for` -> `self_is_handler_for` ?

src/lib.rs
octol: Good catch

src/data_handler.rs
```diff
@@ -274,11 +274,6 @@ impl DataHandler {
             //
             // ===== Coins =====
             //
```
madadam: You can now remove this comment (and the "Login packets" one below)

octol: Updated

PR #770, opened 2019-07-23
Check app permissions

### LINKED ISSUE #761: Check App permissions
We need to update the vaults so that when handling requests from Apps they check the following:
* if the request is a mutation of any kind, the app should have an entry in the `Account` for its owner, and the value of `transfer_coins` must be true to allow the mutation.
* if the request is non-mutating, but for unpublished data (`UnpubImmutableData`, `Unpub{Seq|Unseq}AppendOnlyData`, or `{Seq|Unseq}MutableData`), the app should have an entry in the `Account` for its owner (the value of `transfer_coins` doesn't affect this).

For published data, non-mutating requests don't need such checks.

madadam:
Closes #761 .

### DIFF
--- a/src/utils.rs
+++ b/src/utils.rs
@@ -12,7 +12,8 @@ use log::{error, trace};
 use pickledb::{PickleDb, PickleDbDumpPolicy};
 use rand::{distributions::Standard, thread_rng, Rng};
 use safe_nd::{
-    ADataAddress, ClientPublicId, Error as NdError, PublicId, PublicKey, Request, Response, XorName,
+    ADataAddress, ClientPublicId, Error as NdError, IDataAddress, PublicId, PublicKey, Request,
+    Response, XorName,
 };
 use serde::Serialize;
 use std::{fs, path::Path};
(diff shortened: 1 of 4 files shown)


src/utils.rs
```diff
@@ -225,6 +226,82 @@ pub fn to_error_response(request: &Request, error: NdError) -> Response {
...
+    MutUnpub,
+}
+
+// Returns the type of authorisation needed for the given request.
+pub(crate) fn authorisation_kind(request: &Request) -> AuthorisationKind {
+    use AuthorisationKind::*;
+    use Request::*;
+
+    match request {
+        PutIData(_)
```
Fraser999: Can't `PutIData` cover `MutPub` too?

madadam: Not sure. I based this on where we used to call `has_signature()` and we did call it unconditionally for `PutIData`: https://github.com/maidsafe/safe_vault/blob/master/src/client_handler.rs#L243

Fraser999: Oh - I missed what I think is an error above.  At line 1241 of client_handler.rs, I think `AuthorisationKind::MutPub` should also be setting `signature_required` to true.  AFAIK, the only types not requiring a signature are get/list requests for published types.  I wonder if it's even worth having `MutPub` and `MutUnpub` actually?

madadam: Originally I didn't had only `Mut`, but then I noticed that `AppendSeq` and `AppendUnseq` didn't require signature if the data is published.

madadam: ...which I'm not sure is correct actually :thinking:

Fraser999: Hmm - yes, I think they should require the signature too.

PR #771, opened 2019-07-23
Notify client when balance is increased

### LINKED ISSUE #746: Send message to client when its balance is increased
When a coin balance is increased, the vaults should send a message to the balance's owner if it is connected.

This should be a new variant in safe_nd::Message: `Notification(Transaction)`.

octol:
Closes #746

### DIFF
--- a/src/client_handler.rs
+++ b/src/client_handler.rs
@@ -28,9 +28,9 @@ use lazy_static::lazy_static;
 use log::{error, info, trace, warn};
 use safe_nd::{
     AData, ADataAddress, AppPermissions, AppPublicId, Challenge, Coins, Error as NdError, IData,
-    IDataAddress, IDataKind, LoginPacket, MData, Message, MessageId, NodePublicId, PublicId,
-    PublicKey, Request, Response, Result as NdResult, Signature, Transaction, TransactionId,
-    XorName,
+    IDataAddress, IDataKind, LoginPacket, MData, Message, MessageId, NodePublicId, Notification,
+    PublicId, PublicKey, Request, Response, Result as NdResult, Signature, Transaction,
+    TransactionId, XorName,
 };
 use serde::Serialize;
 use std::{
@@ -170,6 +170,12 @@ impl ClientHandler {
                     info!(
                         "{}: {} invalidly sent {:?}",
                         self, client.public_id, response
+                    );
+                }
+                Ok(Message::Notification { notification, .. }) => {
+                    info!(
+                        "{}: {} invalidly sent {:?}",
+                        self, client.public_id, notification
                     );
                 }
                 Err(err) => {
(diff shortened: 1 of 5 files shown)


src/client_handler.rs
```diff
@@ -171,6 +172,12 @@ impl ClientHandler {
                         self, client.public_id, response
                     );
                 }
+                Ok(Message::Notification { .. }) => {
+                    info!(
+                        "{}: Notification from {} not handled",
```
Fraser999: Maybe using a similar log message as for Message::Response (i.e. saying "invalidly sent") or something like that would be more helpful?  I might interpret this as the vault doing the wrong thing (choosing to not handle the client message) rather than the client being at fault.

src/client_handler.rs
```diff
@@ -1041,6 +1058,20 @@ impl ClientHandler {
         self.quic_p2p.send(recipient, msg, 0)
     }
 
+    fn send_notification_to_client(&mut self, client_id: PublicId, notification: Notification) {
+        let peer_addr = if let Some(peer_addr) = self.lookup_client_peer_addr(&client_id) {
+            *peer_addr
+        } else {
+            info!("{}: client {} not found", self, client_id);
```
Fraser999: This could maybe also include why we're trying to find the client?  Something like `"can't notify {} as it's not connected"`

src/client_handler.rs
```diff
@@ -1070,6 +1101,13 @@ impl ClientHandler {
             .map(|(peer_addr, _)| peer_addr)
     }
 
+    fn lookup_client_id(&self, name: &XorName) -> Option<PublicId> {
+        self.clients
+            .values()
+            .find(|client| client.public_id.name() == name)
```
Fraser999: This will match the client or any apps where the owner is the relevant client.  Maybe we should only be notifying actual `Client` types and never `App`s - or maybe just notify all of those that match?

octol: Notifying app owners seems reasonable no?

Fraser999: Yes, I'd think we should at least notify the owner.  Just not sure if we should be notifying all the connected apps too.

octol: What do you think of something like
```
    fn lookup_client_and_owner_ids(&self, name: &XorName) -> Vec<PublicId> {
        self.clients
            .values()
            .filter(|client| client.public_id.name() == name)
            .filter_map(|client_info| utils::owner(&client_info.public_id))
            .map(|public_id| PublicId::Client(public_id.clone()))
            .collect::<Vec<_>>()
    }
```
?

Fraser999: Won't that just return the owner's ID repeated?

I was thinking something more like:
```rust
fn lookup_owner_and_its_apps(&self, name: &XorName) -> Vec<&PublicId> {
    self.clients
        .values()
        .filter_map(|client| {
            if client.public_id.name() == name {
                Some(&client.public_id)
            } else {
                None
            }
        })
        .collect()
}
```

PR #782, opened 2019-07-31
Build and Release in Jenkins

jacderida:
Adds a build and release process in Jenkins.

## Build & Test Stage

* In parallel:
    - Run clippy
    - Build and test on Linux
    - Build with musl on Linux
    - Build and test on macOS
    - Build and test on Windows
* Upload build artifacts

The reason for having a separate path for building with musl is because we can't run the tests against a musl version: they pull in dependencies that result in a linker failure related to libc. However, I had a conversation with Nikita and he said if `cargo build --release --target x86_64-unknown-linux-musl` works, then we should be able to distribute that just fine.

## Deployment Stage

* Clone the code with the tags fetched (not done by default)
* Retrieve build artifacts from first stage
* If the commit message is a version change:
    - Get the version number
    - Package the artifacts with version number in file name
    - Create a tag for this version
    - Publish a GitHub release
    - Publish to crates.io
* Otherwise:
    - Package the artifacts with commit hash in file name
    - Upload the artifacts to the safe-vault S3 bucket

### Deployment without Version Change

[Build log](https://jenkins.maidsafe.net/blue/organizations/jenkins/pipeline-safe_vault/detail/PR-782/4/pipeline)

Artifacts produced:
```
aws s3 ls s3://safe-vault | sort
2019-08-01 11:55:00    9226240 safe_vault-17bcf39-x86_64-apple-darwin.tar
2019-08-01 11:55:01   12400640 safe_vault-17bcf39-x86_64-unknown-linux-gnu.tar
2019-08-01 11:55:01   20428800 safe_vault-17bcf39-x86_64-pc-windows-gnu.tar
```

### Deployment with Version Change

[Build log](https://jenkins.maidsafe.net/blue/organizations/jenkins/pipeline-safe_vault/detail/PR-782/13).

[Release](https://github.com/jacderida/safe_vault/releases/tag/0.18.1) on my fork.

Since you can't delete anything from crates.io, I wasn't able to run a real publish. However, I have tested the same code on a sample project:

* [Here](https://crates.io/crates/jenkins_sample_lib) are some crates that were uploaded.
* [Here](https://jenkins.maidsafe.net/blue/organizations/jenkins/pipeline-sandbox/detail/pipeline-sandbox/3/pipeline/) is the build that produced crate 0.0.3.
* [Here](https://github.com/maidsafe/jenkins_sample_lib/blob/42f98a4a9d2954d1e063808651b9dd90b4a38dd1/Jenkinsfile#L41) is the Jenkinsfile for the sample project

Cheers,

Chris

### DIFF
--- a/Dockerfile.build
+++ b/Dockerfile.build
@@ -10,7 +10,11 @@ RUN addgroup --gid 1001 maidsafe && \
 # We could perhaps put this into a base container at a later stage.
 RUN USER=maidsafe && \
     GROUP=maidsafe && \
-    curl -SsL https://github.com/boxboat/fixuid/releases/download/v0.4/fixuid-0.4-linux-amd64.tar.gz | tar -C /usr/local/bin -xzf - && \
+    echo "06b3e053be5aaccc91dd5a45faf2356f  fixuid-0.4-linux-amd64.tar.gz" > fixuid-0.4-linux-amd64.tar.gz.md5 && \
+    curl -OSsL https://github.com/boxboat/fixuid/releases/download/v0.4/fixuid-0.4-linux-amd64.tar.gz && \
+    md5sum -c fixuid-0.4-linux-amd64.tar.gz.md5 && \
+    tar -C /usr/local/bin -xzf fixuid-0.4-linux-amd64.tar.gz && \
+    rm fixuid-0.4-linux-amd64.tar.gz && \
     chown root:root /usr/local/bin/fixuid && \
     chmod 4755 /usr/local/bin/fixuid && \
     mkdir -p /etc/fixuid && \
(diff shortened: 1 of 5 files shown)


Dockerfile.build
```diff
@@ -0,0 +1,40 @@
...
+    adduser --uid 1001 --ingroup maidsafe --home /home/maidsafe --shell /bin/sh --disabled-password --gecos "" maidsafe && \
+    # The parent container sets this to the 'staff' group, which causes problems
+    # with reading code stored in Cargo's registry.
+    chgrp -R maidsafe /usr/local
+
+# Install fixuid for dealing with permissions issues with mounted volumes.
+# We could perhaps put this into a base container at a later stage.
+RUN USER=maidsafe && \
+    GROUP=maidsafe && \
+    curl -SsL https://github.com/boxboat/fixuid/releases/download/v0.4/fixuid-0.4-linux-amd64.tar.gz | tar -C /usr/local/bin -xzf - && \
```
octol: Is the checksum of this checked?

jacderida: @octol Sorry, I had my head in something else the last few days. Added an md5 verification.

PR #795, opened 2019-08-12
Create balance as part of creating LoginPacket

### LINKED ISSUE #790: Pre-loaded balances should be created automatically during identity creation

octol:
At least during phase 1 we want pre-loaded balances to be created
automatically during identity creation.

Resolves #790


octol:
Closed in favour of #796

PR #801, opened 2019-08-14
Hello from betterthantrav from the Safe Network Forum.

JoeTheCodeMonkey:
Hello this is betterthantrav from the Safe Network Forum. I am learning to code
and use Git/Github. Just wanted to say hello and thank you for all that you
guys are doing.


dirvine:
You are more than welcome. We move fast here, but will try and hekp

JoeTheCodeMonkey:
Thank you. I wanted to participate and tried learning rust, but it was to much to take on. So I decided web development might be more my speed. Thanks again and good luck with the safe network.

PR #802, opened 2019-08-15
Make payment strict

octol:
- Make payment strict, that is don't allow sufficient balance
- Update integration tests to reflect this
- Update `COST_OF_PUT` to match `safe_client_libs`
- Remove deprecated `has_balance`

### DIFF
--- a/tests/common/mod.rs
+++ b/tests/common/mod.rs
@@ -577,3 +577,7 @@ impl IntoCoins for u64 {
         unwrap!(Coins::from_nano(self))
     }
 }
+
+pub fn multiply_coins(coins: Coins, factor: u64) -> Coins {
+    unwrap!(Coins::from_nano(coins.as_nano() * factor))
+}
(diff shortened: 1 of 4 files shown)


tests/common/mod.rs
```diff
@@ -577,3 +577,11 @@ impl IntoCoins for u64 {
         unwrap!(Coins::from_nano(self))
     }
 }
+
+pub fn multiply_coins(coins: Coins, factor: isize) -> Coins {
```
jeanphilippeD: Maybe 
```
pub fn multiply_coins(coins: Coins, factor: isize) -> Coins {
    Coins::from_nano((coins.as_nano() * factor) as u64)
}
```

Unfortunate that Coins does not do multiplication by a scalar.

octol: Yes good point, also I don't know for how large factor the monadic chaining scales to. I'll change to your suggestion.

PR #809, opened 2019-08-19
Handle put duplicate unpublished immutable data

### LINKED ISSUE #808: Putting unpublished immutable data at the occupied address should return DataExists
In accordance with the [RFC text](https://github.com/maidsafe/rfcs/blob/master/text/0055-unpublished-immutable-data/0055-unpublished-immutable-data.md)

octol:
Resolves #808

### DIFF
--- a/src/data_handler/idata_handler.rs
+++ b/src/data_handler/idata_handler.rs
@@ -76,12 +76,18 @@ impl IDataHandler {
         };
 
         if self.metadata.exists(&(*kind.address()).to_db_key()) {
-            trace!(
-                "{}: Replying success for Put {:?}, it already exists.",
-                self,
-                kind
-            );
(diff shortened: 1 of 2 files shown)


src/data_handler/idata_handler.rs
```diff
@@ -76,12 +76,16 @@ impl IDataHandler {
...
-            return respond(Ok(()));
+            if kind.is_pub() {
+                trace!(
+                    "{}: Replying success for Put {:?}, it already exists.",
+                    self,
+                    kind
+                );
+                return respond(Ok(()));
+            } else {
+                return respond(Err(NdError::DataExists));
```
nbaksalyar: Maybe just add a quick comment clarifying that for immutable data it's expected that we return `DataExists`?

PR #814, opened 2019-08-20
Report back errors to client for DeleteUnpubIData

### LINKED ISSUE #807: Elders assume for IData delete requests that Adults always report success
This scenario is specific to immutable data as it's the only one where Elders delegate to Adults (and sometimes Elders). For phase 1 the Elder effectively delegates to itself, but the flow is generic in anticipation for future phases.

In `idata_holder.rs` is the logic run by Adults (and Elders) when putting and deleting immutable data. Currently that can return things like `NdError::AccessDenied` if you try to delete data you're not the owner of. However in `idata_handler.rs` the Elder always assume that mutation requests when delegated to Adults (and Elders) succeed.

See 
`IDataHandler::handle_put_idata_resp`
`IDataHandler::handle_delete_unpub_idata_resp`
where the `NdResult` is unused.

For Put requests this probably doesn't have a too big impact, but **for deletions this mean that you can delete other users `UnpublishedIData`**. The integration test for this is in `delete_immutable_data()` where it is disabled
```rust
    // Currently Elders always assume mutation requests (put/delete) to Adults succeed.
    if false {
        // Delete unpublished data without being the owner
        common::send_request_expect_err(
            &mut env,
            &mut client_b,
            Request::DeleteUnpubIData(IDataAddress::Unpub(unpub_idata_address)),
            NdError::AccessDenied,
        );
    }

```

octol:
Resolves #807

### DIFF
--- a/src/client_handler.rs
+++ b/src/client_handler.rs
@@ -107,11 +107,13 @@ impl ClientHandler {
             "Our connection info:\n{}\n",
             unwrap!(serde_json::to_string(&our_conn_info))
         );
-        if let Ok(connection_info_file) = write_connection_info(&our_conn_info) {
-            println!(
-                "Writing connection info to: {}",
-                connection_info_file.to_string_lossy()
-            );
+        if !cfg!(feature = "mock") {
+            if let Ok(connection_info_file) = write_connection_info(&our_conn_info) {
+                println!(
+                    "Writing connection info to: {}",
+                    connection_info_file.display()
+                );
+            }
         }
 
         Ok((quic_p2p, event_receiver))
--- a/src/data_handler/idata_handler.rs
+++ b/src/data_handler/idata_handler.rs
@@ -222,7 +222,7 @@ impl IDataHandler {
         let (idata_address, op_type) = self.idata_op_mut(&message_id).and_then(|idata_op| {
             let op_type = idata_op.op_type();
             idata_op
-                .handle_mutation_resp(sender, own_id, message_id)
+                .handle_mutation_resp(sender, result.clone(), own_id, message_id)
                 .map(|address| (address, op_type))
         })?;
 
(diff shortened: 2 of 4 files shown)


src/data_handler/idata_op.rs
```diff
@@ -92,9 +94,27 @@ impl IDataOp {
...
+            })
+            .filter(|(_, response)| {
+                if let Response::Mutation(Err(_)) = response {
+                    true
+                } else {
+                    false
+                }
+            })
+            .collect()
+    }
```
octol: This function seems clunky, must be a better way?

src/data_handler/idata_handler.rs
```diff
@@ -286,46 +286,64 @@ impl IDataHandler {
...
-            warn!(
-                "{}: Failed to get metadata from DB: {:?}",
-                self, idata_address
-            );
-            None
-        });
-
-        if let Some(mut metadata) = metadata {
-            if !metadata.holders.remove(&sender) {
+        // TODO - Only rudimentary checks for if requests to Adult nodes were successful. These
```
jeanphilippeD: Function is a bit on the bigger side, but I guess we can always refactor later.

src/data_handler/idata_op.rs
jeanphilippeD: Can you do something like:
```rust
             .filter_map(|(sender, state)| match state {
                RpcState::Actioned(response @ Response::Mutation(Err(_))) => Some((*sender, response.clone())),
                _ => None,
            }) 
```

octol: Interesting, that's new syntax I learned today :)

src/client_handler.rs
```diff
@@ -107,11 +107,13 @@ impl ClientHandler {
...
             unwrap!(serde_json::to_string(&our_conn_info))
         );
-        if let Ok(connection_info_file) = write_connection_info(&our_conn_info) {
-            println!(
-                "Writing connection info to: {}",
-                connection_info_file.to_string_lossy()
-            );
+        if cfg!(test) {
+            if let Ok(connection_info_file) = write_connection_info(&our_conn_info) {
+                println!(
```
madadam: Just out of curiosity (it's not part of this PR): why do we need this printed out in tests?

octol: It's a bug, I just pushed a fixup that inverts the check :)

octol:
> Looks good, however I would feel more confident if this PR also included tests for the new behaviour.

There is a test that I enabled as part of this PR though https://github.com/maidsafe/safe_vault/pull/814/files#diff-2a2d8cb84eb7f679ed0fdc62e67e9071R2153, you're thinking of something more extensive?

src/client_handler.rs
```diff
@@ -107,11 +107,13 @@ impl ClientHandler {
...
-        if let Ok(connection_info_file) = write_connection_info(&our_conn_info) {
-            println!(
-                "Writing connection info to: {}",
-                connection_info_file.to_string_lossy()
-            );
+        if !cfg!(feature = "mock") {
+            if let Ok(connection_info_file) = write_connection_info(&our_conn_info) {
+                println!(
+                    "Writing connection info to: {}",
+                    connection_info_file.to_string_lossy()
```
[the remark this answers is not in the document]
octol: I'll update the PR

madadam:
> There is a test that I enabled as part of this PR though https://github.com/maidsafe/safe_vault/pull/814/files#diff-2a2d8cb84eb7f679ed0fdc62e67e9071R2153, you're thinking of something more extensive?

Sorry, must have missed that. I think that should be sufficient for now.

PR #815, opened 2019-08-20
App coin operations

### LINKED ISSUE #811: App without transfer_coins permission shouldn't be allowed to use coin operations
Currently the `get_balance` operation passes through even if `transfer_coins` is set to `false`.

We should add an integration tests in vaults to check this behaviour.

madadam:
- Add tests for coin operations by apps. 
- Add stricter permissions requirement for `GetBalance`

Closes #811

### DIFF
--- a/src/client_handler.rs
+++ b/src/client_handler.rs
@@ -1504,7 +1504,9 @@ impl ClientHandler {
         signature: Option<Signature>,
     ) -> Option<()> {
         let signature_required = match utils::authorisation_kind(request) {
-            AuthorisationKind::GetUnpub | AuthorisationKind::Mut => true,
+            AuthorisationKind::GetUnpub
+            | AuthorisationKind::Mut
+            | AuthorisationKind::GetBalance => true,
             AuthorisationKind::GetPub => false,
         };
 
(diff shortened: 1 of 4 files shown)


src/client_handler.rs
```diff
@@ -1560,11 +1568,15 @@ impl ClientHandler {
...
+        check: impl FnOnce(AppPermissions) -> bool,
+    ) -> Result<(), NdError> {
         if self
             .auth_keys
             .app_permissions(app_id)
-            .map(|perms| perms.transfer_coins)
+            .map(check)
             .unwrap_or(false)
         {
             Ok(())
```
jeanphilippeD: Maybe could keep the `.ok_or(NdError::AccessDenied)` like it was in `authorise_app_for_unpublished_get`?

madadam: I'm not sure what you mean here. Could you elaborate? Note that in `authorise_app_for_unpublished_get` we didn't have to check the permission itself, just that the corresponding entry exists. That's not the case here.

jeanphilippeD: Yes, I did not realize we needed to get the conversion to () and also filter.
So we'd get something like that which is probably not better than the if/else.

```rust
     fn check_app_permissions(
        &self,
        app_id: &AppPublicId,
        check: impl FnOnce(AppPermissions) -> bool,
    ) -> Result<(), NdError> {
        self
            .auth_keys
            .app_permissions(app_id)
            .map(check)
            .filter(|value| value) // did not realize we needed that
            .map(|_| ()) // did not realize we needed that
            .ok_or(NdError::AccessDenied)
    } 
```

PR #817, opened 2019-08-21
Refactor flow for `CreateLoginPacketFor`

### LINKED ISSUE #804: CreateCoinBalanceFor should return a Transaction response
For a `CreateCoinBalanceFor` request the response type should be `Response::Transaction` for both success and error types.

### LINKED ISSUE #806: The requester of CreateCoinBalanceFor should be debited the cost of creating the balance
When sending a `CreateCoinBalanceFor` request with the amount set as `5 safecoin`. Then the requester is debited six coins. 5 for the new balance and 1 for the login packet insertion.

Currently, the newly created balance is debited the cost of storing the account.

lionel1704:
Resolves #804, resolves #806

### DIFF
--- a/src/client_handler.rs
+++ b/src/client_handler.rs
@@ -1283,12 +1283,14 @@ impl ClientHandler {
             self.send_response_to_client(
                 payer,
                 message_id,
-                Response::Mutation(Err(NdError::ExceededSize)),
+                Response::Transaction(Err(NdError::ExceededSize)),
             );
             return None;
         }
+        // The requester bears the cost of storing the login packet
+        let new_amount = amount.checked_add(*COST_OF_PUT)?;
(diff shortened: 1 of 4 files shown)


src/client_handler.rs
```diff
@@ -1283,12 +1283,15 @@ impl ClientHandler {
...
                 payer,
                 message_id,
-                Response::Mutation(Err(NdError::ExceededSize)),
+                Response::Transaction(Err(NdError::ExceededSize)),
             );
             return None;
         }
+        // The requester bears the cost of storing the login packet
+        let new_amount =
+            Coins::from_nano(amount.as_nano().checked_add(COST_OF_PUT.as_nano())?).ok()?;
```
madadam: `Coins` has `checked_add`, so you can do just this:
```rust
let new_amount = amount.checked_add(COST_OF_PUT)?;
```

PR #824, opened 2019-08-28
Add cost to create a balance

### LINKED ISSUE #818: Charge 1 nano coin for creating a new coin balance
- This issue is of lower priority (perhaps a Phase 2 task) because it doesn't cause any tests to fail as they are not expecting a charge currently.

lionel1704:
- cost the requester one nano coin for creating a balance
- allow creating a balance with zero coins
closes #818

### DIFF
--- a/src/client_handler.rs
+++ b/src/client_handler.rs
@@ -928,12 +928,13 @@ impl ClientHandler {
         transaction_id: TransactionId,
         message_id: MessageId,
     ) -> Option<Action> {
+        let request = Request::CreateBalance {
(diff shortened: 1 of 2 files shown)


src/client_handler.rs
```diff
@@ -978,7 +992,8 @@ impl ClientHandler {
                 }
             }
             Err(error) => {
-                // Send refund.
+                let amount = amount.checked_add(*COST_OF_PUT)?;
```
octol: I think we should probably map this to an `Error` but maybe that could be done later

PR #828, opened 2019-08-30
Tell the user about connections

octol:
So that it doesn't give the false impression that the vault is stuck


octol:
I think this is now superseded by #829

PR #829, opened 2019-08-30
Add verbose command line flag

octol:
Adds a command line argument `-v` that can be repeated like `-vv` to increase verbosity in the console output. It is equivalent to and overrides RUST_LOG if present.

### DIFF
--- a/src/config_handler.rs
+++ b/src/config_handler.rs
@@ -8,7 +8,7 @@
 
 use crate::{quic_p2p::Config as QuicP2pConfig, quic_p2p::NodeInfo, Result};
 use directories::ProjectDirs;
-use log::trace;
+use log::{trace, Level};
 use serde::{Deserialize, Serialize};
 use std::{
     env,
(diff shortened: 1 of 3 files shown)


src/config_handler.rs
```diff
@@ -69,13 +74,19 @@ impl Config {
             wallet_address: None,
             max_capacity: None,
             root_dir: None,
+            verbose: 0,
```
nbaksalyar: Maybe use a noun instead of an adjective? I.e.
```suggestion
            verbosity: 0,
```

octol: When passing the full flag I think `safe_vault --verbose` sounds slightly better than `safe_vault --verbosity`, but I'm no English speaker :)

nbaksalyar: Ah yes, for the CLI `--verbose` works better for sure. I meant more like a structure field, but I have no strong preference here.

octol: Yes as a structure field I agree. `structopt` wants the field the same as the flag I think though

nbaksalyar: That's what we can do as it seems:
https://github.com/rust-cli/clap-verbosity-flag/blob/ad073540fc6714e0d0bb186eef44eb5c1adf0c73/src/lib.rs#L35-L36

I.e.,
```rust
#[structopt(long = "verbose", short = "v", parse(from_occurrences))]
verbosity: u8,
```

octol: Ah yes, good point. I'll update

octol: I started changing but the smoke test https://github.com/maidsafe/safe_vault/pull/829/files#diff-0fb07a438eb34eab8647cbaa5b5a7482R293 doesn't seem to happy about it, so I'm inclined to keep as is?

nbaksalyar: Happy to keep it as it is now :+1: Thanks!

PR #830, opened 2019-08-30
chore/update: self update after config parsing

jacderida:
Running the update before this causes an update check (and hence connectivity to the internet) even if you just want to display the --help menu.


nbaksalyar:
I believe this has been fixed by https://github.com/maidsafe/safe_vault/pull/829 already, so closing this PR.

PR #849, opened 2019-09-11
add initial ParsecAction with immediate execution

### LINKED ISSUE #841: Extend Mock: Create continuation scheme
(1-2 weeks) Create continuation scheme to use in all Vault flows.
        - Need to design this early.
        - (5-7 days?) JP: Maybe use existing continuation mechanism to use new Action.
        - (7-10 days?) Maybe need to explore option into more chunk? Might be option.
        - N: Possibly call back or other things.

jeanphilippeD:
Closes #841 

Initial flow to allow Parsec to be used at the pay step.
All the similar data mutation are handled the same way with a vote that result in the pay and forward request action to be executed immediately.

To move to real parsec:
Convert that line to a vote `ParsecVote(action) => Some(ParsecConsensus(action))` and return None action. And on consensus create an action and handle it.

### DIFF
--- a/src/action.rs
+++ b/src/action.rs
@@ -8,9 +8,11 @@
 
 use crate::rpc::Rpc;
 use safe_nd::{Coins, MessageId, PublicId, Request, XorName};
+use serde::{Deserialize, Serialize};
 use std::collections::BTreeSet;
 
-#[derive(Debug)]
+#[derive(Debug, Clone, Serialize, Deserialize)]
+// Need to Serialize/Deserialize to go through Parsec
 pub(crate) enum ParsecAction {
     // Process pay for request and forward request to client.
     PayAndForwardClientRequest {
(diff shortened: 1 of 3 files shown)


jeanphilippeD:
Updated PR and commit titles/descriptions.

Fixed clippy format issue and rebase.
Also added additional commit to complete transformation for similar requests.

src/action.rs
```diff
@@ -7,12 +7,30 @@
...
+    },
+}
+
 #[derive(Debug)]
 #[allow(clippy::large_enum_variant)]
 pub(crate) enum Action {
+    // Trigger a vote for an event so we can process the defered action on consensus.
+    // (Currently immediatly.)
+    ParsecVote(ParsecAction),
+    // Process defered action (eventually will wait for Parsec consensus).
```
nbaksalyar:
```suggestion
    // Process deferred action (eventually will wait for Parsec consensus).
```

src/action.rs
```diff
@@ -7,12 +7,30 @@
...
+        client_public_id: PublicId,
+        message_id: MessageId,
+        cost: Coins,
+    },
+}
+
 #[derive(Debug)]
 #[allow(clippy::large_enum_variant)]
 pub(crate) enum Action {
+    // Trigger a vote for an event so we can process the defered action on consensus.
```
nbaksalyar:
```suggestion
    // Trigger a vote for an event so we can process the deferred action on consensus.
```

src/action.rs
```diff
@@ -7,12 +7,30 @@
...
+    // (Currently immediatly.)
```
nbaksalyar:
```suggestion
    // (Currently immediately.)
```

jeanphilippeD:
Fixed typos is respective commits.

PR #852, opened 2019-09-18
Add basic mock routing functions

### LINKED ISSUE #838: Initial integration: First Initial Mock Routing
(1-3 days) First Initial Mock Routing with only the minimum that will be need to be called (`try_next_evt`) does not actually do anything.

### LINKED ISSUE #839: Initial integration: Setup routing instance
(2-4 days) Setup the routing instance in `Vault`, probably in `State::Elder`, in the form of `routing::Node`. Make sure Routing is progressing (integration with event loop) (could be just logs). 
      => Is there problem with quic_p2p? probably not since vault is not really using vaults.
      => Command line argument to bootstrap? No, can just setup routing as first node (hard code everything).

### LINKED ISSUE #840: Extend Mock: Expose `vote_for`
(1-3 days) expose a `vote_for` method to insert payloads into Parsec and new output event to signal consensus about a custom event. (On Vote signale consensus reached) - Make sure it is sensible for routing as well. (No mock_parsec at this stage just pass through)
      => Decide type `Vec<u8>` or enum.
      => The current thinking leans towards accumulating events in routing to keep the API exposed to safe_vault simple.
      => Implement routing EventBox trait with a channel to fit well with current design in Vault.

nbaksalyar:
Closes #838, closes #839, closes #840.

### DIFF
--- a/src/vault.rs
+++ b/src/vault.rs
@@ -7,12 +7,13 @@
 // permissions and limitations relating to use of the SAFE Network Software.
 
 use crate::{
-    action::Action,
+    action::{Action, ConsensusAction},
     adult::Adult,
     client_handler::ClientHandler,
     coins_handler::CoinsHandler,
     data_handler::DataHandler,
     quic_p2p::{Event, NodeInfo},
+    routing::{Event as RoutingEvent, Node},
     rpc::Rpc,
     utils, Config, Error, Result,
 };
@@ -62,6 +63,7 @@ pub struct Vault {
     state: State,
     event_receiver: Receiver<Event>,
     command_receiver: Receiver<Command>,
+    routing_node: Node,
 }
 
 impl Vault {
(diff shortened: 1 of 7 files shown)


src/vault.rs
```diff
@@ -160,20 +165,47 @@ impl Vault {
...
             maybe_action = self.handle_action(action);
         }
     }
 
+    fn handle_routing_event(&mut self, event: RoutingEvent) -> Option<Action> {
+        let client_handler = self.client_handler_mut()?;
+        match event {
+            RoutingEvent::Consensus(custom_event) => {
+                match bincode::deserialize::<ParsecAction>(&custom_event) {
+                    Ok(parsec_action) => client_handler.handle_parsec_action(parsec_action),
```
jeanphilippeD: Maybe worth being a bit more precised : `handle_consensused_parsec_action`?

nbaksalyar: Maybe even `handle_consensused_action`? And rename `ParsecAction` to `ConsensusAction`?

jeanphilippeD: Seem good, there does not seem to be a need to mention Parsec in the vault code.
then best to make sure we do not use that word in other functions as well.

PR #854, opened 2019-09-20
Send requests for deletion and create login packet via consensus

### LINKED ISSUE #855: Send deletion requests via routing API
- There is no pay, but still consensus in Client-Handler and continue flow like put and mutation.
- smaller.

### LINKED ISSUE #856: Send CreateLoginPacket request via consensus
Modify LoginPacket flows to go through parsec at the pay part of the flow. Transaction goes through only after Parsec consensus (source Client-Handler) and then in 2.a & b the remaining flow proceed on all nodes (source client-handler).
    Involve hooking up to allow restarting the flow when Parsec event consensus.
    (Risk: call back or more simple way, once we solve that working. 1 week or more.)
    => Maybe good to do with Modify `AuthKey` `InsAuthKey`
    => Same as mutation maybe need to go to different Action.
    => One of the simplest one.
    => Code has similar flow with an existing Action

### LINKED ISSUE #857: Send CreateLoginPacketFor via consensus
- Same as for mutating Coin flows.
- Maybe the exact same code as mutating Coin.
         => Maybe some additonal action, combintion of login packet and coin flow (Depend on both these done.)

### LINKED ISSUE #858: Modify mutating Coin flows to go via consensus
These are all handled by the ClientHandler.

Modify Coin flows to go through parsec at the pay part of the flow. Transation goes through only after Parsec consensus (source Client-Handler)and then in 2.a the remaining flow proceed on all nodes (including destination client-handler).

-  Small number of flow (not just transfer), but different things for each
- May require new action depending on continuation framework

lionel1704:
Resolves #855 
Resolves #856
Resolves #857
Resolves #858

### DIFF
--- a/src/action.rs
+++ b/src/action.rs
@@ -12,6 +12,7 @@ use serde::{Deserialize, Serialize};
 use std::collections::BTreeSet;
 
 #[derive(Debug, Clone, Serialize, Deserialize)]
+#[allow(clippy::enum_variant_names)]
 // Need to Serialize/Deserialize to go through the consensus process.
 pub(crate) enum ConsensusAction {
     // Process pay for request and forward request to client.
@@ -26,6 +27,14 @@ pub(crate) enum ConsensusAction {
         request: Request,
         client_public_id: PublicId,
         message_id: MessageId,
+    },
+    // Process pay for request and proxy the request to a different client's handler.
+    // Only used by `CreateLoginPacketFor`
+    PayAndProxyClientRequest {
+        request: Request,
+        client_public_id: PublicId,
+        message_id: MessageId,
+        cost: Coins,
     },
 }
 
--- a/src/action.rs
+++ b/src/action.rs
@@ -21,6 +21,12 @@ pub(crate) enum ConsensusAction {
         message_id: MessageId,
         cost: Coins,
     },
+    // Process request that doesn't need a payment and forward request to client.
+    ForwardClientRequest {
+        request: Request,
+        client_public_id: PublicId,
+        message_id: MessageId,
+    },
 }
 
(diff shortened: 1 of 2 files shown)


src/client_handler.rs
```diff
@@ -1324,26 +1356,19 @@ impl ClientHandler {
...
                     new_login_packet: login_packet,
-                };
-                Some(Action::ProxyClientRequest(Rpc::Request {
-                    request,
-                    requester: payer.clone(),
-                    message_id,
-                }))
-            }
-            Err(error) => {
-                self.send_response_to_client(payer, message_id, Response::Transaction(Err(error)));
```
jeanphilippeD: seem like we had some error handling that is no longer there? Is that intentional?

lionel1704: Yes @jeanphilippeD. This is intentional.
After consensus `pay()` will be call and that in turn calls `withdraw()`. The errors will be handled there.

src/action.rs
```diff
@@ -12,6 +12,7 @@ use serde::{Deserialize, Serialize};
 use std::collections::BTreeSet;
 
 #[derive(Debug, Clone, Serialize, Deserialize)]
+#[allow(clippy::enum_variant_names)]
```
m-cat: I'm not familiar with the vault code yet, but instead of allowing this clippy lint wouldn't it be better to rename or refactor something here, or add a `TODO` comment to do so eventually? Every variant ending in "ClientRequest" seems like suboptimal design to me at first glance.

src/client_handler.rs
jeanphilippeD: Indeed, makes sense.

PR #863, opened 2019-09-27
refactor/client_handler: send authkeys operations via consensus

### LINKED ISSUE #862: Send InsAuthKey and DelAuthKey requests via Consensus
Modify `InsAuthKey` and `DelAuthKey` flows to go through parsec. Transition goes through only after Parsec consensus (source Client-Handler) and then in 2.a & b the remaining flow proceed on all nodes (source client-handler).

- `DelAuthKey` is not a paid operation anymore(was a paid op in Phase 1)
- There is no pay, but still consensus in Client-Handler and continue flow in Client-Handler

Yoga07:
Resolves #862

### DIFF
--- a/src/action.rs
+++ b/src/action.rs
@@ -12,6 +12,7 @@ use serde::{Deserialize, Serialize};
 use std::collections::BTreeSet;
 
 #[derive(Debug, Clone, Serialize, Deserialize)]
+#[allow(clippy::enum_variant_names)]
 // Need to Serialize/Deserialize to go through the consensus process.
 pub(crate) enum ConsensusAction {
     // Process pay for request and forward request to client.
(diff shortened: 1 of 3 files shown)


src/utils.rs
```diff
@@ -137,7 +137,7 @@ pub(crate) fn destination_address(request: &Request) -> Option<Cow<XorName>> {
         } => Some(Cow::Borrowed(new_login_packet.destination())),
         UpdateLoginPacket(login_packet) => Some(Cow::Borrowed(login_packet.destination())),
         GetLoginPacket(ref name) => Some(Cow::Borrowed(name)),
-        GetBalance | ListAuthKeysAndVersion | InsAuthKey { .. } | DelAuthKey { .. } => None,
+        _ => None,
```
m-cat: I prefer we avoid `_` in `match` statements as much as possible. This way if a new enum variant is ever added the compiler will warn us that it is not handled here. With `_`, the new variant would always take this branch, returning `None`. If this *is* what we want, I would like a comment here explicitly stating why `_` is fine. Hope that makes sense 🙂

Yoga07: >This way if a new enum variant is ever added the compiler will warn us that it is not handled here.

True! Thought making it `_` would be the generic way of writing it :)

Yoga07:
Closing in favour of #866

PR #865, opened 2019-10-03
fix/vault: change vault's default root dir from temp dir to data dir

### LINKED ISSUE #850: Do not store chunks in the temp directory by default
Currently, we store the actual data chunks in `$TMPDIR` by default, which works fine on Linux and Windows, but causes troubles on Macs: it gets purged occasionally and can result in loss of chunks stored by vaults.

We should store the chunks in one of the [data dirs](https://docs.rs/dirs/2.0.2/dirs/fn.data_dir.html) instead.

Yoga07:
Resolves #850

### DIFF
--- a/src/config_handler.rs
+++ b/src/config_handler.rs
@@ -11,7 +11,6 @@ use directories::ProjectDirs;
 use log::{trace, Level};
 use serde::{Deserialize, Serialize};
 use std::{
-    env,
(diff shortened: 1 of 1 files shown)


Yoga07:
Wasn't aware #864 was already raised for this issue! :) Hence closing this PR.

PR #866, opened 2019-10-03
refactor/client_handler: send insert and delete auth keys request via consensus

### LINKED ISSUE #862: Send InsAuthKey and DelAuthKey requests via Consensus

Yoga07:
Resolves #862

### DIFF
--- a/src/client_handler.rs
+++ b/src/client_handler.rs
@@ -429,9 +429,9 @@ impl ClientHandler {
                 key,
                 version,
                 permissions,
-            } => self.handle_ins_auth_key(client, key, version, permissions, message_id),
+            } => self.handle_ins_auth_key_client_req(client, key, version, permissions, message_id),
             DelAuthKey { key, version } => {
-                self.handle_del_auth_key(client, key, version, message_id)
+                self.handle_del_auth_key_client_req(client, key, version, message_id)
             }
         }
     }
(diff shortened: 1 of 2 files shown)


src/client_handler.rs
```diff
@@ -225,6 +225,27 @@ impl ClientHandler {
...
+                    client_public_id,
+                    key,
+                    version,
+                    permissions,
+                    message_id,
+                ),
+                Request::DelAuthKey { key, version } => {
+                    self.del_auth_key_as_src_client(client_public_id, key, version, message_id)
+                }
+                _ => None,
```
lionel1704: Should we return some kind of an error here (or atleast log it) in case some other request type is recieved here?

Yoga07: Yep, we can. Does it sound good to report back a `warn!(" Request: {:?} cannot be handled as a Src Client Handler.");`?

lionel1704: Yes something like that.
However I don't think this is a warning. Should we use `err!`  instead, as this is a logic error.

src/client_handler.rs
```diff
@@ -822,6 +822,14 @@ impl ClientHandler {
                 updated_login_packet,
                 message_id,
             ),
+            InsAuthKey {
+                key,
+                version,
+                permissions,
+            } => self.ins_auth_key_as_src_client(requester, key, version, permissions, message_id),
+            DelAuthKey { key, version } => {
+                self.del_auth_key_as_src_client(requester, key, version, message_id)
```
lionel1704: Do you want to rename this function to `handle_(ins/del)_auth_key_vault` request like the other functions?

PR #867, opened 2019-10-07
Support multi vault tests

### LINKED ISSUE #853: Run multiple vaults with mock_routing
(2-3 days)
- poll 5 vault as once (Maybe unkown.)
- mock routing vote_for could not change. 
- assumme client send same request to all vaults.
- Environments to setup how many nodes.

nbaksalyar:
Closes #853

This PR also covers the consensus for the `CreateBalance` request and partially addresses some other requirements for the multi-vault testing environment (e.g.: running the existing test suite with multiple vaults).

### DIFF
--- a/src/client_handler.rs
+++ b/src/client_handler.rs
@@ -173,6 +173,7 @@ impl ClientHandler {
 
     pub fn handle_consensused_action(&mut self, action: ConsensusAction) -> Option<Action> {
         use ConsensusAction::*;
+        trace!("{}: Consensused {:?}", self, action,);
         match action {
             PayAndForward {
                 request,
@@ -972,7 +973,11 @@ impl ClientHandler {
             .map(|key| key == &owner_key)
             .unwrap_or(false);
         if own_request {
-            return Some(action);
+            return Some(Action::ConsensusVote(ConsensusAction::Forward {
+                request,
+                client_public_id: requester.clone(),
+                message_id,
+            }));
         }
 
         self.pay(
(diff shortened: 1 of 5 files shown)


src/mock_routing.rs
```diff
@@ -7,11 +7,39 @@
...
+}
 
 /// Interface for sending and receiving messages to and from other nodes, in the role of a full routing node.
 pub struct Node {
-    events: VecDeque<Vec<u8>>,
+    events: EventsRef,
+    consensus_group: Option<Weak<RefCell<ConsensusGroup>>>,
 }
 
 impl Node {
```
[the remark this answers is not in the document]
nbaksalyar: >I don't see the point of the Node::builder() method

It's needed to conform with the public Routing API, but we might as well change it there after some consideration. 👍

src/client_handler.rs
```diff
@@ -968,7 +969,11 @@ impl ClientHandler {
...
         if own_request {
-            return Some(action);
+            return Some(Action::ConsensusVote(ConsensusAction::Forward {
+                request,
+                client_public_id: requester.clone(),
+                message_id,
+            }));
         }
 
         self.pay(
```
lionel1704: Shouldn't this `pay()` operation go via consensus as well ?
As all the client handlers need to be made aware of this operation to debit the client's balance in their coin balances db.

nbaksalyar: Yes, it definitely should! I wanted to tackle this in a separate PR though if that's OK.

nbaksalyar:
@m-cat, I addressed the review comments - thanks!

If this looks fine to you, I'll squash the fixup commit.

src/client_handler.rs
lionel1704: Sure! Sounds good.

nbaksalyar:
Thanks, rebased :+1:

PR #869, opened 2019-10-08
refactor/idata: introduce IDataRequest to eliminate unwraps

m-cat:
A small refactor and minor fixes.


src/action.rs
```diff
@@ -12,23 +12,23 @@ use serde::{Deserialize, Serialize};
 use std::collections::BTreeSet;
 
 #[derive(Debug, Clone, Serialize, Deserialize)]
-// Need to Serialize/Deserialize to go through the consensus process.
+/// Need to Serialize/Deserialize to go through the consensus process.
```
lionel1704: Doesn't look like this line is a doc string

PR #871, opened 2019-10-09
Send UpdateLoginPacket req via consensus

### LINKED ISSUE #870: Send UpdateLoginPacket via consensus
There is no pay, but still consensus in Client-Handler and continue flow to complete update.
=> Maybe go together with AuthKey DelAuthKey
=> A tiny bit different because no action, probably just need to send through Parsec, and do the full flow when it comes out.

lionel1704:
Resolves #870

### DIFF
--- a/src/client_handler.rs
+++ b/src/client_handler.rs
@@ -412,8 +412,8 @@ impl ClientHandler {
                 new_login_packet,
                 message_id,
             ),
-            UpdateLoginPacket(ref updated_login_packet) => self.handle_update_login_packet_req(
-                &client.public_id,
+            UpdateLoginPacket(updated_login_packet) => self.handle_update_login_packet_req(
+                client.public_id.clone(),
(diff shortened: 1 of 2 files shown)


src/client_handler.rs
```diff
@@ -1095,6 +1099,35 @@ impl ClientHandler {
...
+            )
+            .and_then(|_existing_login_packet| {
+                if !updated_login_packet.size_is_valid() {
+                    return Err(NdError::ExceededSize);
+                }
+                self.login_packets
+                    .put(&updated_login_packet)
+                    .map_err(|err| err.to_string().into())
+            });
+        Some(Action::RespondToClientHandlers {
```
nbaksalyar: While `Action::RespondToClientHandlers` is totally correct in this context, it might be slightly confusing when it's used in the `ClientHandler` module. I'm wondering if we should consider renaming it to something else?

PR #874, opened 2019-10-14
refactor/all: update to safe-nd 0.4.0

lionel1704:
- upgrade to safe-nd 0.4.0
- include granular app permissions for get_balance, transfer_coins and
perform_mutations

### DIFF
--- a/tests/integration.rs
+++ b/tests/integration.rs
@@ -455,7 +455,8 @@ fn coin_operations_by_app() {
     // Create initial balance.
     common::create_balance(&mut env, &mut client_a, None, 10);
 
-    // Create an app with permission to transfer coins.
+    // Create an app with all permissions.
+    // The `perform_mutations` permission is required to send a `CreateBalance` request.
     let mut app = env.new_disconnected_app(client_a.public_id().clone());
     common::perform_mutation(
         &mut env,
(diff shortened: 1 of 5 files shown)


tests/integration.rs
```diff
@@ -2306,7 +2312,9 @@ fn app_permissions() {
             key: *app_0.public_id().public_key(),
             version: 1,
             permissions: AppPermissions {
-                transfer_coins: true,
+                perform_mutations: true,
```
m-cat: The comment says "permission to transfer coins" but here I only see `perform_mutations` is `true`.

tests/integration.rs
```diff
@@ -2210,6 +2214,8 @@ fn auth_keys() {
 
     let permissions = AppPermissions {
         transfer_coins: true,
+        perform_mutations: true,
```
m-cat: Can we add a comment here saying what these permissions are for?

tests/integration.rs
```diff
@@ -465,6 +465,8 @@ fn coin_operations_by_app() {
             version: 1,
             permissions: AppPermissions {
                 transfer_coins: true,
```
m-cat: Can the comment here be updated to say "all permissions" (ideally with an explanation why)? 🙂

PR #877, opened 2019-10-15
Handle refunds when errors are returned from the data handlers

### LINKED ISSUE #873: Handle refunds for legit errors returned from data handlers
Currently a refund is processed only for the `TransferCoins` request. In cases such as when a mutation fails due to errors such as `NoSuchData` or `AccessDenied` at the data handler an error response is sent back to the client handlers and the error is returned to the client.

Assuming that all data handlers behave in the same way, based on the request type (mutation / get) and the response type (Ok / Error) we should process a refund at the ClientHandler before responding to the client.

lionel1704:
Rebased against #874

resolves #873 

Todo:
- [x] See if refunds for transfer coins and create balance can be handled using the same pattern, thus deprecating `Rpc::Refund`
- [x] Add more tests

### DIFF
--- a/src/client_handler.rs
+++ b/src/client_handler.rs
@@ -752,17 +752,7 @@ impl ClientHandler {
                 requester,
                 message_id,
                 refund,
-            } => {
-                if let Some(refund_amount) = refund {
-                    if let Err(error) = self.deposit(requester.name(), refund_amount) {
-                        error!(
-                            "{}: Failed to refund {} coins for {:?}: {:?}",
-                            self, refund_amount, requester, error,
-                        )
-                    };
-                }
-                self.handle_response(src, requester, response, message_id)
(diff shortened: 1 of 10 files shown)


src/client_handler.rs
```diff
@@ -750,14 +750,18 @@ impl ClientHandler {
...
-            Rpc::Refund {
-                requester,
-                amount,
-                transaction_id,
-                reason,
-                message_id,
-            } => self.handle_refund(src, requester, amount, transaction_id, reason, message_id),
+                refund,
+            } => {
+                if let Some(refund_amount) = refund {
```
m-cat: Handling the refund is something I would do in `handle_response`, not `handle_vault_rpc`. I would just add a `refund` parameter to the former function.

PR #879, opened 2019-10-17
feat/mock: make mock clients connect and send requests to 2/3 of the vaults

### LINKED ISSUE #876: Run multiple vaults with mock_routing and only send request to some vaults 2/3
- Mock routing vote_for could be a simple send first one to all vaults and filter other ones. (Alternative could use mock_parsec but maybe more complicated).
- Assume client send same request to 2/3 vaults.
- All vaults reply the same even if they did not get the requests.

Yoga07:
Resolves #876

### DIFF
--- a/tests/common/mod.rs
+++ b/tests/common/mod.rs
@@ -41,6 +41,8 @@ use unwrap::unwrap;
 
 /// Default number of vaults to run the tests with.
 const DEFAULT_NUM_VAULTS: usize = 5;
+/// Default number of vaults to which the test-clients connect to.
+pub const DEFAULT_NUM_CONNECTED_VAULTS: usize = 3;
 
 macro_rules! unexpected {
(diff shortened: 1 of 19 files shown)


tests/common/mod.rs
```diff
@@ -342,7 +349,7 @@ pub struct TestClient {
     rx: Receiver<Event>,
     full_id: FullId,
     public_id: ClientPublicId,
-    connected_vault: Option<NodeInfo>,
+    connected_vaults: Option<Vec<NodeInfo>>,
```
m-cat: Can this be simplified to just `Vec<NodeInfo>`? The `Option` part seems redundant to me, because a `None` would just be equivalent to an empty vector.

lionel1704:
@Yoga07 Can you please raise this PR to the `vNext`branch?

PR #880, opened 2019-10-17
Routing vault eventloop

ustulation:


Cargo.toml
```diff
@@ -15,19 +15,20 @@ base64 = "~0.10.1"
...
 bytes = { version = "~0.4.12", features = ["serde"] }
-crossbeam-channel = "~0.3.8"
+crossbeam = "~0.7.2"
 ctrlc = "~3.1.3"
 directories = "~2.0.1"
 env_logger = "~0.6.2"
 fxhash = { version = "~0.2.1", optional = true }
 hex = "~0.3.2"
 hex_fmt = { version = "~0.3.0", optional = true }
-lazy_static = "~1.3.0"
```
m-cat: We can downgrade this dependency to force things to compile, but it seems kind of backwards... I left this comment on Slack in backend-misc:

> My question to the channel is: does anyone have any objections to using lazy_static = "~1", so we don't run into this again? Do we gain anything by specifying the minor version? This is a very basic dependency and not worth spending so much time on continually updating. For now it would be sufficient to set lazy_static = "~1" in safe_vault.

I don't see the point in continuously wasting time wrestling with dependency issues for minor crates.

nbaksalyar:
Closed in favour of #888

PR #883, opened 2019-10-21
refactor/client_handler: send CreateBalance request via consensus

### LINKED ISSUE #882: Send CreateBalance request via consensus
When calling create balance, the client handlers should vote the event and then debit the requester's balance and forward the request to the destination address' section where the balance is created in each of the client handlers' balances DB.

This issue relates to #872. This is the last request that needs to be sent via consensus.

Yoga07:
Resolves #882

### DIFF
--- a/src/client_handler.rs
+++ b/src/client_handler.rs
@@ -968,44 +968,26 @@ impl ClientHandler {
             amount,
             transaction_id,
         };
-        let action = Action::ForwardClientRequest(Rpc::Request {
-            request: request.clone(),
-            requester: requester.clone(),
-            message_id,
-        });
-
-        // For phase 1 we allow owners to create their own balance freely.
+        // For phases 1 & 2 we allow owners to create their own balance freely.
         let own_request = utils::own_key(requester)
             .map(|key| key == &owner_key)
             .unwrap_or(false);
         if own_request {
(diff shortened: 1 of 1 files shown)


src/client_handler.rs
```diff
@@ -1024,18 +1005,40 @@ impl ClientHandler {
...
+                if amount.as_nano() == 0 || own_request {
+                    Rpc::Response {
+                        response: Response::Transaction(Ok(transaction)),
+                        requester,
+                        message_id,
+                    }
+                } else {
+                    // We debit amout from the owner if the coin balance
+                    // is being created for a non-owner
+                    match self.withdraw(requester.name(), amount) {
```
lionel1704: Perform the withdraw here is not exactly correct. Once a client is debited some amount for creating a balance the request is forwarded to (potentially) a different section where the new balance is created. Performing a `withdraw` for the source client at the destination section would return a no balance found error of sorts.

The right flow would be:
- `PayAndForward` will debit the entire amount: cost of balance creation + inital coins given to the new balance
- The request is then forwarded to the destination section and a balance is created there.

Refund:
     If there is an error in the destination section then an error is returned to the src client handers which returns the amount to the requester and then responds

Even if there is no error the response still goes via the src client handlers since they're the only ones who know the IP + PORT of the client.

PR #885, opened 2019-10-23
Upgrades to phase 1

Yoga07:
This PR contains multiple upgrades for Phase 1:
-  improves app authorisation
- fixes root_dir by using project's data dir by default
- updates to safe-nd 0.4.0
- refactors test to verify granulated app permissions
- changes usage of idata term "kind" to "data"
- introduces IDataRequest to eliminate unwraps
- handles refunds when errors are returned from the data
- deprecates Rpc::Refund to use Rpc::Response instead 
- test that refunds are processed
- introduces util function to calculate refund and move refund logic to handle_response
- removes explicitly-listed non-warn lints

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,7 @@ env:
   global:
     - RUST_BACKTRACE=1
     - PATH=$PATH:$HOME/.cargo/bin
+    - RUSTFLAGS="-D warnings"
 language: rust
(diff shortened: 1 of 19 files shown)


lionel1704:
@Yoga07 Can you also add 5aa17fd to this please?

PR #886, opened 2019-10-24
feat/mock: make mock clients connect and send requests to 2/3 of the vaults

### LINKED ISSUE #876: Run multiple vaults with mock_routing and only send request to some vaults 2/3

Yoga07:
Resolves #876

### DIFF
--- a/src/mock_routing.rs
+++ b/src/mock_routing.rs
@@ -8,7 +8,7 @@
 
 use crossbeam_channel::TryRecvError;
 use std::cell::RefCell;
-use std::collections::VecDeque;
+use std::collections::{btree_set::BTreeSet, VecDeque};
 use std::rc::{Rc, Weak};
 
 /// Default events `VecDeque` pre-allocated capacity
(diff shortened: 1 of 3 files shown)


Yoga07:
> Nice one @Yoga07!
> Everything looks good to me.
> 
> I have an idea as well. Let me know what you think.
> Instead of storing the entire event inside the `processed_events` vector can we store the`sha3` of  the event. It would be a lot more optimal since the event holds a lot of data including entire data payloads.

Nice one! This list could tend to grow large over a period of time and there can be chances where huge amounts of data are pushed. Therefore it makes sense to reduce the load stored in it. I'll get this done. Thanks!

Yoga07:
So the user might alter the number

> Looks good!
> Do you think it's a good idea to add an assertion in `Environment::with_multiple_vaults()` to verify that
> `DEFAULT_NUM_VAULTS > DEFAULT_NUM_CONNECTED_VAULTS` ?

Yes, that'll be a good assertion and also if the user tries to run only one vault by changing `DEFAULT_NUM_VAULTS` to `1`, he should also be altering `DEFAULT_NUM_CONNECTED_VAULTS` to `1` too, which is another case to be asserted.

Yoga07:
Closing this PR as it has been rendered invalid with the recent updates in the mock_routing module.

PR #889, opened 2019-11-01
chore/lint: forbid unsafe and add badge

m-cat:
I noticed that `unsafe_code` can actually be `forbid` (which is stronger than `deny`), since we don't use any unsafe code in this crate! So I changed it and added the `unsafe forbidden` badge.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,7 @@ env:
   global:
     - RUST_BACKTRACE=1
     - PATH=$PATH:$HOME/.cargo/bin
+    - RUSTFLAGS="-D warnings"
 language: rust
 rust: stable
 branches:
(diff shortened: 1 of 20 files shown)


m-cat:
Looks like the `crossbeam_channel::select` macro contains `#[allow(unsafe)]` 🙁 This scenario isn't possible to catch with grep...

m-cat:
This PR should be good to go once [this PR in crossbeam](https://github.com/crossbeam-rs/crossbeam/pull/431) gets merged.

lionel1704:
`select!` is being removed in #888.
Once that is merged, we will no longer have to wait for crossbeam-rs/crossbeam#431

Issue #898, opened 2019-11-24
Safe Vault binary: `unwrap_or` swallows errors is json config file


peter-wilkins:
errors from `serde_json::from_reader` are silently ignored and a default config is used instead. 

https://github.com/maidsafe/safe_vault/blob/efbe8d8de35df73f1428e05a1fc7e35dcbfe4615/src/config_handler.rs#L85

It might be preferable to exit if bad configuration file exists. it took me a while to figure out what was going on.

Also noticed that calls to trace!() in read_from_file() do nothing as the logger is not setup till later.

https://github.com/maidsafe/safe_vault/blob/efbe8d8de35df73f1428e05a1fc7e35dcbfe4615/src/bin/safe_vault.rs#L64-L72

PR #899, opened 2019-11-24
fix/vault: reading config file returns option

peter-wilkins:
disambiguate following cases:
- config file found and deserialized successfully
- config file found but error occurred (panic)
- no config file found (use default config)

update calls to trace with println as logging not available at this point
(logging setup requires config)

Resolves https://github.com/maidsafe/safe_vault/issues/898

### DIFF
--- a/src/bin/safe_vault.rs
+++ b/src/bin/safe_vault.rs
@@ -44,7 +44,7 @@ mod detail {
 
     /// Runs a SAFE Network vault.
     pub fn main() {
-        let mut config = Config::new();
+        let mut config = Config::new().expect("Error reading configuration file");
 
         if let Some(c) = &config.completions() {
             match c.parse::<clap::Shell>() {
--- a/src/config_handler.rs
+++ b/src/config_handler.rs
@@ -10,7 +10,7 @@ use crate::routing::{ConnectionInfo, NetworkConfig};
 use crate::Result;
 use directories::ProjectDirs;
 use lazy_static::lazy_static;
-use log::{trace, Level};
+use log::Level;
 use serde::{Deserialize, Serialize};
 use std::{
     fs::{self, File},
@@ -90,16 +90,8 @@ pub struct Config {
 impl Config {
     /// Returns a new `Config` instance.  Tries to read from the default vault config file location,
     /// and overrides values with any equivalent command line args.
-    pub fn new() -> Self {
-        let mut config = Self::read_from_file().unwrap_or(Self {
-            wallet_address: None,
-            max_capacity: None,
-            root_dir: None,
-            verbose: 0,
-            network_config: Default::default(),
-            first: false,
-            completions: None,
-        });
+    pub fn new() -> Result<Self> {
+        let mut config = Self::read_from_file()?.unwrap_or_default();
 
         let command_line_args = Config::clap().get_matches();
         for arg in &ARGS {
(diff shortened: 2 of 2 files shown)


src/config_handler.rs
```diff
@@ -82,13 +82,16 @@ impl Config {
...
     /// and overrides values with any equivalent command line args.
     pub fn new() -> Self {
-        let mut config = Self::read_from_file().unwrap_or(Self {
-            wallet_address: None,
-            max_capacity: None,
-            root_dir: None,
-            verbose: 0,
-            quic_p2p_config: Default::default(),
-        });
+        let mut config = match Self::read_from_file().unwrap() {
```
m-cat: I think you can do `.unwrap().unwrap_or_default()` here. We can just return `Default::default()` in the `None` case, and `unwrap_or_default()` lets you do that more concisely than by using a `match` statement.

src/config_handler.rs
```diff
@@ -82,13 +82,7 @@ impl Config {
...
     /// and overrides values with any equivalent command line args.
     pub fn new() -> Self {
-        let mut config = Self::read_from_file().unwrap_or(Self {
-            wallet_address: None,
-            max_capacity: None,
-            root_dir: None,
-            verbose: 0,
-            quic_p2p_config: Default::default(),
-        });
+        let mut config = Self::read_from_file().unwrap().unwrap_or_default();
```
m-cat: Just noticed something: we try to avoid `unwrap()` in our Rust code. You can read about this policy in our Rust style guide, linked to from [here](https://github.com/maidsafe/safe_client_libs/wiki/Guide-to-contributing#code-style). In this case, I think you can use `unwrap_or_default()` twice, but you'd have to play around with it.

peter-wilkins: ah yes, the library should not panic. It shouldn't swallow errors either. How about returning a Result so the caller?

m-cat: That's even better :)

src/bin/safe_vault.rs
```diff
@@ -43,7 +43,7 @@ mod detail {
 
     /// Runs a SAFE Network vault.
     pub fn main() {
-        let mut config = Config::new();
+        let mut config = Config::new().expect("Error reading configuation file");
```
m-cat: Hmm, looks like we just moved the panic from `Config::new` over to `main`. I think it would be best here to display any error that occurs, while instantiating a default config instead of panicking. What do you think?

peter-wilkins: Sure, if that's the behaviour you want - as long as it tells the user why it is ignoring the config in the file. In my experience it is safer to exit if there are config issues but it depends on the situation.
I will update. Thanks.

m-cat: That's true, users could be confused if the binary doesn't read their config and they don't see the error message.

m-cat: I'm fine with leaving it as it is now. There is a typo in the error message here though :)

peter-wilkins: ah, damn displexia

lionel1704:
Nice job guys! Happy to get this merged. However the master branch is for updates to Phase 1, could you please raise a PR with these changes to the `vNext` branch?

m-cat:
Sure lionel, I can change the target branch here, without opening a new PR. I'll do the rebase now, there are just conflicts that I need to take care of.

PR #901, opened 2019-11-29
feat\vault: use mock quic-p2p from routing

maqi:


tests/common/mod.rs
```diff
@@ -58,31 +55,9 @@ impl Environment {
     pub fn with_multiple_vaults(num_vaults: usize) -> Self {
         assert!(num_vaults > 0);
 
-        let do_format = move |formatter: &mut Formatter, record: &Record<'_>| {
```
jeanphilippeD: Why did that change for this PR?

tests/common/mod.rs
```diff
@@ -268,17 +243,37 @@ pub trait TestClientTrait {
     }
 
     fn expect_new_message(&self) -> (SocketAddr, Bytes) {
-        match self.rx().try_recv() {
-            Ok(Event::NewMessage { peer_addr, msg }) => (peer_addr, msg),
-            x => unexpected!(x),
+        let max_fetches = 10;
+        let mut steps = 0;
+        while steps < max_fetches {
```
jeanphilippeD: Why did that change for this PR?

tests/common/mod.rs
```diff
@@ -268,17 +243,37 @@ pub trait TestClientTrait {
...
+        panic!("haven't got a new message after {} fetches", max_fetches);
     }
 
     fn expect_no_new_message(&self) {
-        match self.rx().try_recv() {
-            Err(error) => assert!(error.is_empty()),
-            x => unexpected!(x),
+        let max_fetches = 10;
+        let mut steps = 0;
+        while steps < max_fetches {
```
jeanphilippeD: same?

tests/common/mod.rs
maqi: We can now use the Network from routing directly, and doesn't require this format anymore.

tests/common/mod.rs
maqi: there is new event `SentUserMessage` defined in routing, which will be received during normal communication and shall be ignored here

tests/common/mod.rs
maqi: yeah, same reason.
there is new event `SentUserMessage` defined in routing, which will be received during normal communication and shall be ignored here

tests/common/mod.rs
jeanphilippeD: Isn't it for log? isn't it up to vault how to decide how to format things?

tests/common/mod.rs
jeanphilippeD: Then why do we have a max_fetches and steps? Shouldn't we loop until either no more events are there or we get the one we look for?

tests/common/mod.rs
jeanphilippeD: same

tests/common/mod.rs
maqi: the logbuilder here conflicts with the log initializer inside the routing  :(

jeanphilippeD: Ok, not ideal, probably still needed when using mock_routing though?

jeanphilippeD: Is there a way it would not conflict?

jeanphilippeD: I guess the right question for @nbaksalyar is should we remove this or should it be used, and if it should be used what should it do.

maqi: >probably still needed when using mock_routing though

vault::mock_routing is using routing::mock_quick_p2p which has the log initializer. 

> Is there a way it would not conflict?

make routing::mock_quick_p2p using the log builder from vault? however I think it will make the current API exposing more stuff.

jeanphilippeD: Ok, make sense, if @nbaksalyar is happy, I'm ok with it.

nbaksalyar: >make routing::mock_quick_p2p using the log builder from vault? 

Maybe the other way around, @maqi? That is, instead of exposing & defining the logger in `mock-quic-p2p` directly, initialise it only at the use site, not in the library code. In this case the use site is `safe_vault::tests::Environment`, and in Routing I guess we can have something similar and initialise the logger in the Routing tests (within a wrapper, if needed)?

>should we remove this or should it be used, and if it should be used what should it do.

I agree this is not ideal. We should decouple test logs from Routing. Currently, routing depends on `maidsafe_utilities` which is deprecated, so eventually it needs to be replaced there too.

maqi: hmm...  the change will then requires the log initializer to be separated from the current routing::mock_quick_p2p::network constructor.  :(

I'd prefer it to be handled as separate issue/PR ?

nbaksalyar: Yeah, separate issue/PR is OK :+1:

PR #909, opened 2019-12-07
feat/vault: add --dump-completions flag similar to that of safe-cli

dan-da:
hi, for your consideration -- I added a --completions option, ie:

```
$ safe_vault --help | grep dump
        --completions <completions>
            dump shell completions for: [bash, fish, zsh, powershell, elvish]
```

Example Usage:

```
$ safe_vault --completions bash > /tmp/vault_completions.sh
$ source /tmp/vault_completions.sh
$ safe_vault --<tab>
--bootstrap-cache-dir       --idle-timeout-msec         --our-complete-cert         --version
--dump-completions          --ip                        --our-type                  --wallet-address
--first                     --keep-alive-interval-msec  --port                      
--hard-coded-contacts       --max-capacity              --root-dir                  
--help                      --max-msg-size-allowed      --verbose 
$ safe_vault --max-<tab>
--max-capacity          --max-msg-size-allowed
```

This is similar in spirit to https://github.com/maidsafe/safe-api/pull/328, though that one is implemented as a subcommand and also has the capability to dump completions for all supported shells at once as json.

The general idea is to make it easy to enable shell completions for the various safe CLI apps, probably at install time.  

note: Since these completions are generated directly by clap/structopt, the output of --completions always remains in sync with the code.

### DIFF
--- a/src/bin/safe_vault.rs
+++ b/src/bin/safe_vault.rs
@@ -192,13 +192,23 @@ mod detail {
         // get filename without preceding path as std::ffi::OsStr (C string)
         let exec_name_ffi = match exe_path.file_name() {
             Some(v) => v,
-            None => return Err(format!("Can't extract file_name of executable from path {}", exe_path.display())),
+            None => {
+                return Err(format!(
+                    "Can't extract file_name of executable from path {}",
+                    exe_path.display()
+                ))
+            }
         };
 
         // Convert OsStr to string.  Can fail if OsStr contains any invalid unicode.
         let exec_name = match exec_name_ffi.to_str() {
             Some(v) => v.to_string(),
-            None => return Err(format!("Can't decode unicode in executable name '{:?}'", exec_name_ffi)),
+            None => {
+                return Err(format!(
+                    "Can't decode unicode in executable name '{:?}'",
+                    exec_name_ffi
+                ))
+            }
         };
 
         // Generates shell completions for <shell> and prints to stdout
--- a/src/bin/safe_vault.rs
+++ b/src/bin/safe_vault.rs
@@ -192,13 +192,13 @@ mod detail {
         // get filename without preceding path as std::ffi::OsStr (C string)
         let exec_name_ffi = match exe_path.file_name() {
             Some(v) => v,
-            None => return Err("Can't extract file_name of executable".to_string()),
+            None => return Err(format!("Can't extract file_name of executable from path {}", exe_path.display())),
         };
 
         // Convert OsStr to string.  Can fail if OsStr contains any invalid unicode.
         let exec_name = match exec_name_ffi.to_str() {
             Some(v) => v.to_string(),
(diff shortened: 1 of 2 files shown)


dan-da:
renamed --dump-completions to --completions for consistency with safe-cli

src/bin/safe_vault.rs
```diff
@@ -166,6 +183,30 @@ mod detail {
...
+
+    fn gen_completions_for_shell(shell: clap::Shell) -> Result<Vec<u8>, String> {
+        // Get exe path
+        let exe_path =
+            std::env::current_exe().map_err(|err| format!("Can't get the exec path: {}", err))?;
+
+        // get filename without preceding path as std::ffi::OsStr (C string)
+        let exec_name_ffi = match exe_path.file_name() {
+            Some(v) => v,
+            None => return Err("Can't extract file_name of executable".to_string()),
```
maqi: better include the `exe_path` as part of the err ?

src/bin/safe_vault.rs
```diff
@@ -166,6 +183,30 @@ mod detail {
...
+        };
+
+        // Convert OsStr to string.  Can fail if OsStr contains any invalid unicode.
+        let exec_name = match exec_name_ffi.to_str() {
+            Some(v) => v.to_string(),
+            None => return Err("Can't decode unicode in executable name".to_string()),
```
maqi: shall the `exec_name_ffi` be included within the error string?

src/config_handler.rs
```diff
@@ -184,6 +194,8 @@ impl Config {
             self.network_config.ip = Some(unwrap!(value.parse()));
         } else if arg == ARGS[11] {
             self.network_config.our_type = unwrap!(value.parse());
+        } else if arg == ARGS[13] {
+            self.completions = Some(unwrap!(value.parse()));
```
maqi: seems `completions` is allowed to be None.
which means we could parse nothing from the `value`.
in that case, using `unwrap` will be a problem?

dan-da: maybe so, but this seems to be the convention in this area of the code.

take wallet_address: It is also an Option<String> and is initialized to None in Config::new().  and in ::set_value() it does `self.wallet_address = Some(unwrap!(value.parse()));`, as do 5 others that are Option<T>.    

So I simply followed convention.... am I missing something?  What do you suggest instead?

src/bin/safe_vault.rs
dan-da: Ok.

src/bin/safe_vault.rs
dan-da: Ok.

src/config_handler.rs
maqi: Ah, fine then. We could adjust it in a separate PR if necessary.

PR #910, opened 2019-12-11
test/vault: use real routing for the integration test, with multiple vaults

maqi:
This PR contains the work to make vault use real routing (with mock_parsec) during the integration test.
The testing environment contains a network of vaults, and the client requests are handled by the network, instead of just one vault.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,6 +19,7 @@ crossbeam-channel = "~0.3.8"
 ctrlc = "~3.1.3"
 directories = "~2.0.1"
 env_logger = "~0.6.2"
+fake_clock = "~0.3.0"
 fxhash = { version = "~0.2.1", optional = true }
 hex = "~0.3.2"
 hex_fmt = { version = "~0.3.0", optional = true }
@@ -47,4 +48,5 @@ doc = false
 
 [features]
 mock = ["routing/mock_base", "fxhash", "hex_fmt"]
+mock-parsec = ["routing/mock_parsec", "routing/mock_crypto", "fxhash", "hex_fmt", "phase-one"]
 phase-one = []
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,6 +28,7 @@ log = "~0.4.7"
 pickledb = "~0.4.0"
 quick-error = "~1.2.2"
 rand = "~0.6.5"
+rand_chacha = "~0.1.1"
 routing = { git = "https://github.com/maidsafe/routing.git", branch = "fleming" }
 safe-nd = "~0.7.0"
 self_update = "0.5.1"
@@ -39,7 +40,6 @@ unwrap = "~1.2.1"
 
 [dev_dependencies]
 maplit = "~1.0.1"
-rand_chacha = "~0.1.1"
 tempdir = "~0.3.7"
 
 [[bin]]
@@ -47,6 +47,6 @@ name = "safe_vault"
 doc = false
 
 [features]
-mock = ["routing/mock_base", "fxhash", "hex_fmt"]
-mock-parsec = ["routing/mock_parsec", "routing/mock_crypto", "fxhash", "hex_fmt", "phase-one"]
-phase-one = []
+mock_base = ["routing/mock_base", "fxhash", "hex_fmt"]
+mock = ["mock_base"]
+mock_parsec = ["routing/mock_parsec", "routing/mock_crypto", "mock_base"]
--- a/src/client_handler.rs
+++ b/src/client_handler.rs
@@ -25,6 +25,7 @@ use crate::{
 use bytes::Bytes;
 use lazy_static::lazy_static;
 use log::{debug, error, info, trace, warn};
+use rand::{CryptoRng, Rng};
 use safe_nd::{
     AData, ADataAddress, AppPermissions, AppPublicId, Coins, ConnectionInfo, Error as NdError,
     HandshakeRequest, HandshakeResponse, IData, IDataAddress, IDataKind, LoginPacket, MData,
@@ -57,6 +58,7 @@ pub(crate) struct ClientHandler {
     balances: BalancesDb,
     clients: HashMap<SocketAddr, ClientInfo>,
     pending_msg_ids: HashMap<MessageId, SocketAddr>,
+    pending_actions: HashMap<MessageId, Response>,
     // Map of new client connections to the challenge value we sent them.
     client_candidates: HashMap<SocketAddr, (Vec<u8>, PublicId)>,
     login_packets: LoginPacketChunkStore,
@@ -87,6 +89,7 @@ impl ClientHandler {
             balances,
             clients: Default::default(),
             pending_msg_ids: Default::default(),
+            pending_actions: Default::default(),
             client_candidates: Default::default(),
             login_packets,
             routing_node,
@@ -177,7 +180,12 @@ impl ClientHandler {
         }
     }
 
-    pub fn handle_client_message(&mut self, peer_addr: SocketAddr, bytes: Bytes) -> Option<Action> {
+    pub fn handle_client_message<R: CryptoRng + Rng>(
+        &mut self,
+        peer_addr: SocketAddr,
+        bytes: Bytes,
+        rng: &mut R,
+    ) -> Option<Action> {
         if let Some(client) = self.clients.get(&peer_addr).cloned() {
             match bincode::deserialize(&bytes) {
                 Ok(Message::Request {
@@ -185,11 +193,27 @@ impl ClientHandler {
                     message_id,
                     signature,
                 }) => {
+                    // We could receive a consensused vault rpc contains a client request,
+                    // before receiving the request from that client directly.
+                    if let Some(response) = self.pending_actions.remove(&message_id) {
+                        self.send(
+                            peer_addr,
+                            &Message::Response {
+                                response,
+                                message_id,
+                            },
+                        );
+                        return None;
+                    }
+
                     if let Entry::Vacant(ve) = self.pending_msg_ids.entry(message_id) {
                         let _ = ve.insert(peer_addr);
                         return self.handle_client_request(&client, request, message_id, signature);
                     } else {
-                        info!("Pending MessageId reused - ignoring client message.");
+                        info!(
+                            "Pending MessageId {:?} reused - ignoring client message.",
+                            message_id
+                        );
                     }
                 }
                 Ok(Message::Response { response, .. }) => {
@@ -217,7 +241,7 @@ impl ClientHandler {
                     self.handle_bootstrap_request(peer_addr, client_id);
                 }
                 Ok(HandshakeRequest::Join(client_id)) => {
-                    self.handle_join_request(peer_addr, client_id);
+                    self.handle_join_request(peer_addr, client_id, rng);
                 }
                 Ok(HandshakeRequest::ChallengeResult(signature)) => {
                     self.handle_challenge(peer_addr, signature);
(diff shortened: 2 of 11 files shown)


src/config_handler.rs
```diff
@@ -185,7 +185,7 @@ impl Config {
         } else if arg == ARGS[11] {
             self.network_config.our_type = unwrap!(value.parse());
         } else {
-            #[cfg(not(feature = "mock"))]
+            #[cfg(not(any(feature = "mock", feature = "mock-parsec")))]
```
jeanphilippeD: seem like it would be better have a `mock_base` feature `["routing/mock_base", "fxhash", "hex_fmt"]
And only config on this.

Then
mock = [mock_base]
mock_parsec = ["routing/mock_parsec", "routing/mock_crypto", "mock_base", "phase-one"]

Then no need to have these more complicated feature gate.

Cargo.toml
```diff
@@ -47,4 +48,5 @@ doc = false
 
 [features]
 mock = ["routing/mock_base", "fxhash", "hex_fmt"]
+mock-parsec = ["routing/mock_parsec", "routing/mock_crypto", "fxhash", "hex_fmt", "phase-one"]
 phase-one = []
```
jeanphilippeD: phase_one?
mock_parsec?

for consistency with routing at least?

src/vault.rs
```diff
@@ -188,8 +196,18 @@ impl Vault {
...
+        let mut max_poll = 0;
         loop {
+            max_poll += 1;
+            if processed || max_poll > 1000 {
+                return processed;
+            }
+
+            #[cfg(feature = "mock-parsec")]
+            {
+                processed = self.routing_node.borrow_mut().poll();
```
jeanphilippeD: Why is this needed, why is `handle_selected_operation`, and should we make it sufficient somehow if not?

src/vault.rs
```diff
@@ -188,8 +196,18 @@ impl Vault {
     /// Returns whether at least one event was processed.
     pub fn poll(&mut self) -> bool {
         let mut processed = false;
-
+        let mut max_poll = 0;
         loop {
+            max_poll += 1;
+            if processed || max_poll > 1000 {
```
jeanphilippeD: do we really need that within that function, is poll not call in a loop itself?

src/vault.rs
```diff
@@ -219,12 +237,12 @@ impl Vault {
...
                     }
                     idx => {
-                        if let Err(err) = self
+                        if let Err(_err) = self
                             .routing_node
                             .borrow_mut()
                             .handle_selected_operation(idx)
                         {
-                            warn!("Could not process operation: {}", err);
+                            // warn!("Could not process operation: {}", err);
```
jeanphilippeD: why?

Cargo.toml
maqi: yeah, good suggestion, I will change that

tests/common/mod.rs
```diff
@@ -76,6 +86,43 @@ impl Environment {
...
+
+        let network = Network::new(Default::default());
+
+        let mut env = Self {
+            rng: rng::new(),
+            network,
+            vaults: Default::default(),
+        };
+
+        if num_vaults > 1 {
```
jeanphilippeD: only 1 vault with this PR? so can remove that part of the if?

src/config_handler.rs
maqi: yeah, I will try to change

tests/common/mod.rs
```diff
@@ -93,6 +141,20 @@ impl Environment {
...
+    // Poll the mock network and the environment's vault.
+    pub fn poll(&mut self) {
+        let mut max_poll = 0;
+        while max_poll < 100 {
+            max_poll += 1;
+            self.network.poll();
+            let _ = self.vaults.iter_mut().for_each(|vault| {
+                let _ = vault.inner.poll();
+            });
+            FakeClock::advance_time(1001);
```
jeanphilippeD: I see why we need that `FakeClock::advance_time(1001);`, I'm not clear why it need to ignore the progress part?
It seem a bit strange to always move the test forward 100 seconds each time we call poll

tests/common/mod.rs
```diff
@@ -154,9 +218,54 @@ impl TestVault {
...
+
+        Self {
+            inner,
+            _root_dir: root_dir,
+            _command_tx: command_tx,
+        }
+    }
+
+    #[cfg(feature = "mock-parsec")]
+    fn new_with_real_routing_first() -> Self {
```
jeanphilippeD: Seem to be mostly duplicate with new_with_real_routing

src/vault.rs
maqi: only `poll` checks the timer advancing, `handle_selected_operation` doesnt

tests/common/mod.rs
maqi: I will carry out a de-depulication

src/vault.rs
jeanphilippeD: What do you mean by check the timer advancing?
And why is handle_selected_operation not, should it?

tests/common/mod.rs
maqi: will have a double check if this still applies with the other changes.

tests/common/mod.rs
maqi: there are multiple vaults.
I think we will support multiple vaults with real routing soon, so don't want to change backwards and forwards.
The existing code already take multple vaults with mock_routing. So just want to ensure the code behave the same for both.

src/vault.rs
maqi: there are too many logs for this so got it disabled.
I think it is due to the `try_ready` method used

src/vault.rs
maqi: I remember I tried, it doesn't work. will have a double check again.

tests/common/mod.rs
jeanphilippeD: It does not make sense in the context of this PR, it would be best to either complete the work so it work with multiple vault, or remove the code that is not get us to fully functional multiple vaults.

maqi: I thought PR 903 (https://github.com/maidsafe/safe_vault/pull/903) is to deal with `purely one vault network`

Yeah, probably this PR shall be better work with multiple vaults.
I will add a WIP label then.

src/vault.rs
maqi: the `routing::Node::poll` eventually calls `routing::StateMachine::try_step`, which checks the timed_out_tokens
However, the `handle_selected_operation ` doesn't, as it is just for mock_routing.

Cargo.toml
```diff
@@ -38,13 +40,13 @@ unwrap = "~1.2.1"
...
 [[bin]]
 name = "safe_vault"
 doc = false
 
 [features]
-mock = ["routing/mock_base", "fxhash", "hex_fmt"]
-phase-one = []
+mock_base = ["routing/mock_base", "fxhash", "hex_fmt"]
+mock = ["mock_base"]
+mock_parsec = ["routing/mock_parsec", "routing/mock_crypto", "mock_base"]
```
octol: So `mock` means `mock_routing`, yet `mock_parsec` depends on `mock` but still uses real routing?

src/vault.rs
```diff
@@ -188,10 +208,15 @@ impl Vault {
 
     /// Processes any outstanding network events and returns. Does not block.
     /// Returns whether at least one event was processed.
+    #[allow(unused_assignments)]
```
octol: Maybe mark the variable `_processed` to limit the scope of this rule?

tests/common/mod.rs
```diff
@@ -128,15 +194,20 @@ trait AsMutSlice<T> {
...
 impl TestVault {
     /// Create a test Vault within a group.
-    fn new(consensus_group: Option<ConsensusGroupRef>, mut rng: TestRng) -> Self {
+    #[cfg(feature = "mock")]
+    fn new_with_mock_routing(
+        consensus_group: Option<ConsensusGroupRef>,
+        network: &Network,
+    ) -> Self {
         let root_dir = unwrap!(TempDir::new("safe_vault"));
+        trace!("creating a test vault at root_dir {:?}", root_dir);
```
octol: s/creating/Creating ?

src/vault.rs
```diff
@@ -189,9 +209,13 @@ impl Vault {
     /// Processes any outstanding network events and returns. Does not block.
     /// Returns whether at least one event was processed.
     pub fn poll(&mut self) -> bool {
-        let mut processed = false;
-
+        let mut _processed = false;
```
jeanphilippeD: why the _

maqi: just trying to limit the scope of rule, as mentioned at: https://github.com/maidsafe/safe_vault/pull/910#discussion_r359381896

Without `_`, we will have to use `#[allow(unused_assignments)]` above the function

jeanphilippeD: I don't get it, we return the value.
Your problem is that you probably should just have:

`let mut processed;`

maqi: I tried `let mut processed;`
there is a complain of `value assigned to processed is never read`
I think it's because the `returning` is not considered as a `read`, as there is an initial value assigned, however may not be used before being set to `true` ?

src/client_handler.rs
```diff
@@ -1555,7 +1582,7 @@ impl ClientHandler {
         let valid = if let Some(signature) = signature {
             self.is_valid_client_signature(public_id, request, &message_id, &signature)
         } else {
-            warn!(
+            println!(
```
nbaksalyar: Is there a reason to change from one of the log macros to `println!`?

src/vault.rs
```diff
@@ -189,9 +209,13 @@ impl Vault {
...
         loop {
+            #[cfg(feature = "mock_parsec")]
```
nbaksalyar: Probably something like
```rust
if cfg!(feature = "mock_parsec") {
  ...
}
```
would be more idiomatic?

tests/common/mod.rs
```diff
@@ -96,6 +142,20 @@ impl Environment {
...
+    // Poll the mock network and the environment's vault.
+    pub fn poll(&mut self) {
+        let mut processed = true;
+        while processed {
+            processed = false;
+            self.network.poll();
+            self.vaults
+                .iter_mut()
+                .for_each(|vault| processed = processed || vault.inner.poll());
+            FakeClock::advance_time(1001);
```
nbaksalyar: `1001` looks like an arbitrary magic number, so maybe put it into a constant or add a comment?

tests/common/mod.rs
```diff
@@ -220,23 +333,27 @@ pub trait TestClientTrait {
...
-            Ok(Event::ConnectedTo {
-                peer: Peer::Node { ref node_info },
-            }) if node_info == conn_info => (),
-            x => unexpected!(x),
+        loop {
+            match self.rx().try_recv() {
+                Ok(Event::ConnectedTo {
+                    peer: Peer::Node { ref node_info },
+                }) if node_info == conn_info => return,
+                // _ => (),
```
nbaksalyar: Looks like dead code

src/client_handler.rs
maqi: Ah, missed this one. a left over from the debuggings. will change it back.

tests/common/mod.rs
maqi: just used the same as in routing. will put a comment in

tests/common/mod.rs
maqi: forgot to remove, will do

src/vault.rs
maqi: Cannot do that.

Although the code won't be executed when compile with `mock`, the compiler will still looking for routing_node having a `poll` function, which is not provided (needed) by mock_routing.

PR #911, opened 2019-12-11
New bootstrap handshake format

octol:


Cargo.toml
```diff
@@ -28,7 +28,8 @@ pickledb = "~0.4.0"
 quick-error = "~1.2.2"
 rand = "~0.6.5"
 routing = { git = "https://github.com/maidsafe/routing.git", branch = "fleming" }
-safe-nd = "~0.4.0"
+# safe-nd = "~0.5.0"
+safe-nd = { path = "../safe-nd" }
```
octol: I'll need to update this before upgrading the draft to a PR

src/client_handler.rs
```diff
@@ -647,33 +635,56 @@ impl ClientHandler {
...
-                {
-                    warn!("{}: Could not disconnect client: {:?}", self, err);
+    /// If a client requests the section info, we also send it.
+    fn handle_challenge(&mut self, peer_addr: SocketAddr, signature: Signature) {
+        if let Some((challenge, public_id)) = self.client_candidates.remove(&peer_addr) {
+            let public_key = match utils::own_key(&public_id) {
+                Some(pk) => pk,
+                None => {
+                    info!(
+                        "{}: Client on {} identifies as a node: {}",
```
maqi: this log better to be a `warn` ? also the wording can be :
`{}: Client on {} identifies as a node: {}, hence disconnect from it.`

src/client_handler.rs
```diff
@@ -1106,6 +1117,77 @@ impl ClientHandler {
...
+                    .collect::<Vec<_>>(),
+                Err(e) => {
+                    info!("Could not handle bootstrap request: {:?}", e);
+                    return;
+                }
+            };
+
+            if closest_known_elders.is_empty() {
+                warn!(
+                    "{}: No closest known elders in any section we know of",
```
maqi: this sounds more like meaning this vault knows no elders ?
I'd prefer a log_or_panic here as it's very suspicious.

src/client_handler.rs
```diff
@@ -1106,6 +1117,77 @@ impl ClientHandler {
...
+                    self
+                );
+            } else {
+                self.send(peer_addr, &HandshakeResponse::Join(closest_known_elders));
```
maqi: just to confirm, so this means we are sending the `HandshakeResponse::Join(closest_known_elders)` back to the client,
and he shall try to connect to them, right?

src/mock_routing/mod.rs
```diff
@@ -94,6 +99,26 @@ impl Node {
         Ok(())
     }
 
+    /// Find out if the given XorName matches our prefix.
+    pub fn matches_our_prefix(&self, _name: &routing::XorName) -> Result<bool, RoutingError> {
```
maqi: better import `XorName` in the above use block ?
and same to the other places

tests/common/mod.rs
```diff
@@ -226,28 +236,56 @@ pub trait TestClientTrait {
...
+
+        let handshake_response: HandshakeResponse = unwrap!(bincode::deserialize(&bytes));
+        let _payload = match handshake_response {
+            HandshakeResponse::Join(payload) => payload,
+            _ => panic!("Unexpected HandshakeResponse"),
+        };
+
+        // TODO: handle payload
+        // Currently we are connected to a single vault. At this stage we should disconnect from it
+        // and connect to the new elders we are given as payload.
```
maqi: Shall we simulate the process even with just one vault?
I am asking this as the `self.send(&request)` actually only sends to the connected vaults.
which seems the test is not covering the whole procedure?

src/client_handler.rs
octol: Don't think we use `log_or_panic` in this crate though

src/client_handler.rs
octol: That's correct

tests/common/mod.rs
octol: This is phase 2B though for multiple sections?

src/client_handler.rs
maqi: ok, then

tests/common/mod.rs
maqi: I mean, this flow indicates that if within the same section, then probably don't need to carry out the disconnection and connect to the new.
So, maybe worth to mentionin the comment that:
`Disconnect if we are not among the elders` ?

octol: Good point, I'll update the comment

PR #912, opened 2019-12-12
test/vault: make vault rng seeded from routing's

maqi:


tests/common/rng.rs
```diff
@@ -6,28 +6,13 @@
 // KIND, either express or implied. Please review the Licences for the specific language governing
 // permissions and limitations relating to use of the SAFE Network Software.
 
-use rand::{self, Rng, SeedableRng};
```
m-cat: Is the `SeedableRng` import still needed?

maqi: yes, the `from_rng` is in the SeedableRng trait

PR #914, opened 2019-12-16
fix/test: resolve non-producable issue

maqi:
The non-producable issue is resolved by fixing the rng initialization within mock tests.

### DIFF
--- a/src/vault.rs
+++ b/src/vault.rs
@@ -19,6 +19,7 @@ use crate::{
 use bincode;
 use crossbeam_channel::{Receiver, Select};
 use log::{error, info, trace, warn};
+use rand::{CryptoRng, Rng};
 use safe_nd::{NodeFullId, Request, XorName};
 use std::borrow::Cow;
 use std::{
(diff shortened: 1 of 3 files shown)


src/vault.rs
```diff
@@ -68,26 +70,75 @@ pub struct Vault {
...
+        )
+    }
+
+    #[cfg(not(feature = "mock"))]
+    /// Create and start vault. This will block until a `Command` to free it is fired.
+    pub fn new(
+        routing_node: Node,
+        event_receiver: Receiver<RoutingEvent>,
+        config: Config,
+        command_receiver: Receiver<Command>,
```
jeanphilippeD: This does not seem ideal, would it make more sense to have that rng argument for both mock and non mock and it be a generic parameter on Rng?

jeanphilippeD: Probably don't need to expose TestRng and chacha if done with generic Rng pram

maqi: the current rng got from routing's network.new_rng doesn't support RngCrypto trait, which is required by safe-nd::NodeFullId .  :(
So we will need this TestRng unless upgrad routing's rand version (or impl RngCrypto there).

For non-mock, I don't mind to pass a rng down, though don't think it make more sense.
Unless you mean the feature guarded code block can be avoided within vault.rs   :)

jeanphilippeD: TestRng implement Rng? does'nt it?

maqi: Yes. But safe_nd::NodeFullId::new function requires the passed in rng implements RngCrypto trait additionally.

And, the current safe-nd used for vault is 0.4, meanwhile the latest safe-nd version is already 0.6.1, which doesn't even have NodeFullId 
So, I'd prefer not touch that part too much and leave it for the later on PR of dealing with the client-vault handshake (which will eforce a version upgrade).

jeanphilippeD: see https://maidsafe.slack.com/archives/C072L6WGH/p1576505822002700?thread_ts=1576502063.002200&cid=C072L6WGH

maqi: updated according to your comment

PR #916, opened 2019-12-16
[vNext] chore/ci: add caching and other changes to GHA

m-cat:
This PR includes all of the GHA commits from `master`.

+ Note that the `master` workflow has been removed.
+ The `test-publish` step from the `pr` workflow has been removed. `vNext` points to a routing git branch which causes `cargo publish` to fail.


.github/workflows/pr.yml
```diff
@@ -0,0 +1,88 @@
+# PR workflow.
```
jeanphilippeD: What was master workflow? `Note that the master workflow has been removed.` 
Should it have been replaced by a vNext workflow?

scripts/clippy
```diff
@@ -2,6 +2,5 @@
 
 set -e -x
 
-cargo fmt -- --check
```
maqi: why the format check is skipped?

.github/workflows/pr.yml
```diff
@@ -0,0 +1,88 @@
...
+        with:
+          profile: minimal
+          toolchain: stable
+          override: true
+          components: rustfmt, clippy
+      # Check if the code is formatted correctly.
+      - uses: actions-rs/cargo@v1
+        with:
+          command: fmt
+          args: --all -- --check
```
jeanphilippeD: This is part of scripts/clippy (mostly not have the --all), should it be there as well?

jeanphilippeD: sorry saw its been removed from the clippy script. (Not ideal because now the script when running locally will not say whether we failed formating).

.github/workflows/pr.yml
m-cat: I considered adding a vNext workflow @jeanphilippeD but I'm not sure if it would be useful.

The `master` workflow does the following:

+ Generates build artifacts
+ Uploads artifacts to s3
+ Generates a release, including release artifacts, if a version change commit was detected
+ Performs an automatic version publish if a version change commit was detected

If you think this is useful for vNext I'd be happy to add it, minus all of the version-change stuff.

scripts/clippy
m-cat: We run `cargo fmt` in the `pr.yml` workflow now, and I didn't think it made sense to have this check in a script called `clippy`. If you relied on this for your workflow, we could add a new script called `checks` which runs fmt and clippy -- what do you think?

.github/workflows/pr.yml
m-cat: See my response to @maqi about adding a script called `checks` and let me know what you think 🙂 I don't think cargo fmt should run in a script called `clippy` imo

scripts/clippy
maqi: I am not using that script, however not sure about others. 
It's ok to me, as long as it will be checked in CI.

.github/workflows/pr.yml
jeanphilippeD: I guess we can do it later.

scripts/clippy
jeanphilippeD: Can always tweak it later. (I'm using it for my flow, avoid having many things to run (i.e on static analyser build that include code formating and one test run is a prety good combination).

But this PR solve the problem we have now so will merge it assuming @maqi approve it

PR #917, opened 2019-12-24
chore/test: enable tests with feature of mock

maqi:
This PR contains the work of re-enabling integration tests when using `mock` feature.
Also resolved some clippy error due to update.

### DIFF
--- a/src/client_handler.rs
+++ b/src/client_handler.rs
@@ -965,7 +965,7 @@ impl ClientHandler {
             .unwrap_or(false);
         if own_request {
             return Some(Action::ConsensusVote(ConsensusAction::Forward {
-                request: request.clone(),
+                request,
                 client_public_id: requester.clone(),
                 message_id,
             }));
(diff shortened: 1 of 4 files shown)


src/client_handler.rs
```diff
@@ -965,7 +965,7 @@ impl ClientHandler {
             .unwrap_or(false);
```
jeanphilippeD: is CI testing it?

maqi: Yes. `cargo test --release --features=mock --no-default-features` was in the test scripts already.

jeanphilippeD: How did it not fail then since this PR is re-enabling it?

maqi: I remember within the last PR that disabling it, I changed the testing flow from carry out certain extra polls to a while loop. This may resolve the previous failures with feature mock automatically.
However, I didn't carry out extra check at that time. (maybe at that time, we can enable it back already)

PR #919, opened 2020-01-02
vault/refactor Small tidy up of imports for routing API cleanup

dirvine:


dirvine:
Good point Marcin, I will push a small change and have pub use at the top.

PR #925, opened 2020-01-09
Use mock-quic-p2p crate, not routing module

madadam:
This PR is a draft, because it uses path/git dependencies until PRs https://github.com/maidsafe/routing/pull/1999 and https://github.com/maidsafe/quic-p2p/pull/87 are merged.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,10 +25,12 @@ hex = "~0.3.2"
 hex_fmt = { version = "~0.3.0", optional = true }
 lazy_static = "~1"
 log = "~0.4.7"
+mock-quic-p2p = { git = "https://github.com/maidsafe/quic-p2p", optional = true }
 pickledb = "~0.4.0"
 quick-error = "~1.2.2"
 rand = "~0.6.5"
 rand_chacha = "~0.1.1"
+routing_rand_core = { package = "rand_core", version = "~0.5.1" }
 routing = { git = "https://github.com/maidsafe/routing.git", branch = "fleming" }
 safe-nd = "~0.7.0"
 self_update = "0.5.1"
@@ -47,6 +49,6 @@ name = "safe_vault"
 doc = false
 
 [features]
-mock_base = ["routing/mock_base", "fxhash", "hex_fmt"]
+mock_base = ["routing/mock_base", "fxhash", "hex_fmt", "mock-quic-p2p"]
 mock = ["mock_base"]
 mock_parsec = ["routing/mock", "mock_base"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,7 +18,7 @@ bytes = { version = "~0.4.12", features = ["serde"] }
 crossbeam-channel = "~0.3.8"
 ctrlc = "~3.1.3"
 directories = "~2.0.1"
-env_logger = "~0.6.2"
+env_logger = "~0.7.1"
 fake_clock = "~0.3.0"
 fxhash = { version = "~0.2.1", optional = true }
 hex = "~0.3.2"
@@ -43,7 +43,6 @@ unwrap = "~1.2.1"
 [dev_dependencies]
 maplit = "~1.0.1"
 tempdir = "~0.3.7"
-pretty_env_logger = "~0.3.1"
 
 [[bin]]
 name = "safe_vault"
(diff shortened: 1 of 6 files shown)


tests/common/rng.rs
```diff
@@ -6,20 +6,94 @@
...
+
+    fn try_fill_bytes(&mut self, dest: &mut [u8]) -> Result<(), Error> {
+        self.0.try_fill_bytes(dest)
+    }
+}
+
+impl CryptoRng for TestRng {}
+
+// Compatibility with routing.
+// TODO: remove this when we update rand to the same version that routing uses.
```
jeanphilippeD: Why was this not needed before?

tests/common/rng.rs
```diff
@@ -6,20 +6,94 @@
...
+        self.0.fill_bytes(dest);
+        Ok(())
+    }
+}
+
+pub fn from_rng<R: RngCore>(rng: &mut R) -> TestRng {
+    TestRng(ChaChaRng::from_seed(rng.gen()))
+}
+
+pub fn from_seed(seed: u64) -> TestRng {
```
jeanphilippeD: Why is that back to using a u64 seed?

tests/common/rng.rs
madadam: Before we used `Network::new_rng` which already returned rng compatible with routing. But that function no longer exists, as it is unrelated to the purpose of the mock-quic-p2p crate.

tests/common/rng.rs
madadam: Previously vault just used the rng mechanism of routing which uses `[u32; 4]` as seed, but that mechanism is not part of the `mock-quic-p2p` crate, so vault now need to do this on its own. And using `u64` is simpler (no need for custom parsing logic).

tests/common/rng.rs
jeanphilippeD: Is the Rng used for both routing and vault. How do you make sure the seed is reproductible, do we need 2 seeds now?

madadam: The rng passed to routing is generated from the vault one, so its enough to seed just the vault one.

Cargo.toml
```diff
@@ -25,10 +25,12 @@ hex = "~0.3.2"
 hex_fmt = { version = "~0.3.0", optional = true }
 lazy_static = "~1"
 log = "~0.4.7"
+mock-quic-p2p = { git = "https://github.com/maidsafe/quic-p2p", optional = true }
 pickledb = "~0.4.0"
 quick-error = "~1.2.2"
 rand = "~0.6.5"
 rand_chacha = "~0.1.1"
+routing_rand_core = { package = "rand_core", version = "~0.5.1" }
```
jeanphilippeD: Might be worth exposing rand_core in routing since it is part of its interface, then we don't have to import a dependency that must mach routing, we can just use it? (I did ask the same for threshold_crypto, but never got around to raise the PR).

Cargo.toml
```diff
@@ -41,12 +43,13 @@ unwrap = "~1.2.1"
 [dev_dependencies]
 maplit = "~1.0.1"
 tempdir = "~0.3.7"
+pretty_env_logger = "~0.3.1"
```
nbaksalyar: I think it makes sense to remove the `env_logger` dependency if we prefer this logger?

Cargo.toml
jeanphilippeD: can be sorted later.

Cargo.toml
jeanphilippeD: I guess this could be sorted later, maybe it make sense to have different logging in production or tests, probably not though?

Cargo.toml
madadam: Maybe create an issue for this in routing?

Cargo.toml
madadam: No, I just missed we were already using env_logger. I can remove this or replace env_logger with pretty_env_logger. No strong preference though.

jeanphilippeD: Whichever you prefer. I guess we can remove pretty_env_logger for now if that change nothing to the code and add it later if we want, then I can just merge that PR.

madadam:
Removed the pretty_env_logger dependency.

PR #930, opened 2020-01-22
Replace AppendOnlyData with Sequence

oetyng:
- Removes `adata_handler.rs`.
- Adds `sequence_handler.rs`.
- Replaces the `adata` `request` and `response` APIs with the equivalent of the `sequence` data type.


src/data_handler/sequence_handler.rs
```diff
@@ -0,0 +1,611 @@
+// Copyright 2020 MaidSafe.net limited.
```
lionel1704: @oetyng could you hold-off from renaming the file right away? It will make the review a little simpler. We could do the renaming at the end (just before merging) if that's okay.

PR #937, opened 2020-02-12
Log only to stdout

iancoleman:
* set target to stdout instead of stderr
* set default log level to INFO (ie verbose=2)
* replace `println!` with appropriate log macros
* replace inconsistent `log::info!` with `info!`
* replace inconsistent `debug!` with `info!` for self update messages

This was motivated by some current work I'm doing to create a vault GUI by parsing / streaming vault log files. It's possible to achieve the same effect with something like
`RUST_LOG=trace ./safe_vault >>vault.log 2>&1`
but these changes make logging more consistent and use sensible defaults.

### DIFF
--- a/src/bin/safe_vault.rs
+++ b/src/bin/safe_vault.rs
@@ -33,8 +33,8 @@ fn main() {
 
 #[cfg(not(feature = "mock"))]
 mod detail {
-    use env_logger::{fmt::Formatter, Builder as LoggerBuilder};
-    use log::{self, Level, Record};
+    use env_logger::{fmt::Formatter, Builder as LoggerBuilder, Target};
+    use log::{error, info, Level, Record};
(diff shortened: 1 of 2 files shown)


src/bin/safe_vault.rs
```diff
@@ -91,19 +92,19 @@ mod detail {
         match update() {
             Ok(status) => {
                 if let Status::Updated { .. } = status {
-                    println!("Vault has been updated. Please restart.");
```
maqi: not sure about this. 
we still need this for CLI interactions?

iancoleman:
Looks like #940 has this change covered. I'll close this one.

bochaco:
I'm no fully sure @iancoleman , let's revisit this one once we have #940 merged, I started with 940 mainly for being able to (optionally) direct the logs to file.

PR #939, opened 2020-02-12
pull routing and quic_p2p with separate client channel

jeanphilippeD:
Depends on https://github.com/maidsafe/quic-p2p/pull/95 and https://github.com/maidsafe/routing/pull/2052

We can now access directly the client messages and do not need to
rely on Routing to provide us with these. We select from the new
channel that routing returns and handle client event direclty.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1427,7 +1427,7 @@ dependencies = [
 [[package]]
 name = "mock-quic-p2p"
 version = "0.1.0"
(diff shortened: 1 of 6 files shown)


jeanphilippeD:
Push update fixing the tests with mock routing.

jeanphilippeD:
Updated to use the merged versions of routing and quic-p2p. Should be ready to merge.

jeanphilippeD:
Fixed formatting.

PR #940, opened 2020-02-18
feat/cli: support --log-dir arg to optionally log to a file instead of stdout/stderr

bochaco:


tests/common/mod.rs
```diff
@@ -67,8 +66,6 @@ impl Environment {
     pub fn with_multiple_vaults(num_vaults: usize) -> Self {
         assert!(num_vaults > 0);
 
-        logging::init();
```
madadam: Does this mean we no longer log in tests? That could make debugging difficult.

src/bin/safe_vault.rs
```diff
@@ -66,27 +66,41 @@ mod detail {
...
         };
-        let mut logger = LoggerBuilder::from_default_env();
-        let _ = logger.format(do_format).is_test(false);
-        if config.verbose() != Level::Error {
-            let _ = logger.filter(
-                Config::clap().get_bin_name(),
-                config.verbose().to_level_filter(),
-            );
+
+        // Let's filter out logs which are not from this crate, unless level chosen is ERROR
```
madadam: This might not be ideal. Sometimes it can be helpful to see the other crate logs when debugging. For example in testnet when something goes wrong the actual error can be anywhere in our stack and this would make it hard to track.

src/bin/safe_vault.rs
```diff
@@ -66,27 +66,41 @@ mod detail {
             config.listen_on_loopback();
         }
 
-        let do_format = move |formatter: &mut Formatter, record: &Record<'_>| {
-            let now = formatter.timestamp();
-            writeln!(
-                formatter,
+        // Custom formatter for logs
+        let do_format = move |write: &mut dyn Write, now: &mut DeferredNow, record: &Record| {
```
maqi: these `write` `now` is a bit keyword like, maybe better use different variable name?
say `writer` and `clock` ?

src/bin/safe_vault.rs
```diff
@@ -66,27 +66,41 @@ mod detail {
...
-        let mut logger = LoggerBuilder::from_default_env();
-        let _ = logger.format(do_format).is_test(false);
-        if config.verbose() != Level::Error {
-            let _ = logger.filter(
-                Config::clap().get_bin_name(),
-                config.verbose().to_level_filter(),
-            );
+
+        // Let's filter out logs which are not from this crate, unless level chosen is ERROR
+        let verbosity = config.verbose().to_level_filter();
```
maqi:
```rust
let filter = {
    let verbosity = config.verbose().to_level_filter();
    let mut filter = verbosity.to_string();
    if verbosity != Level::Error {
        if let Some(crate_name) = Config::clap().get_bin_name() {
            filter = format!("{}={}", crate_name, verbosity);
        }
    }
    filter 
};
```

tests/common/mod.rs
maqi: just to confirm, this is removed because no longer required?

maqi: yeah, same doubt here

src/bin/safe_vault.rs
maqi: agreed. maybe better make this as an configurable option as well.

src/bin/safe_vault.rs
bochaco: yeah, true, I'll rename them

tests/common/mod.rs
bochaco: I'll investigate to make sure logs still work on tests

src/bin/safe_vault.rs
bochaco: I agree, I was simply trying to keep the same logic it had, from my understanding that's what previous `logger.filter(...)` was doing, and it can be overriden with RUST_LOG, but yeah, I can remove the filter to this crate's only

S-Coyle:
I'm not sure why it's not being picked up by the Files tab diff, but in your `Cargo.toml` the `version` needs to be updated to v0.20.1 to match current upstream/fleming.
The result at the moment is that if a vault from this branch is launched it auto updates and the changes from this branch are lost in the updated running vault.

bochaco:
> I'm not sure why it's not being picked up by the Files tab diff, but in your `Cargo.toml` the `version` needs to be updated to v0.20.1 to match current upstream/fleming.
> The result at the moment is that if a vault from this branch is launched it auto updates and the changes from this branch are lost in the updated running vault.

I have no idea, but I've rebased it now with fleming branch, self-update won't happen anyway as I'm also disabled it and added a `--update` flag.

I think moving forward we should also migrate to use https://crates.io/crates/structopt , based on the experience in CLI is much simpler and a cleaner code.

tests/common/mod.rs
bochaco: It's required, I just fixed that, thanks @maqi

bochaco: You were right, I just fixed that

PR #941, opened 2020-02-19
Use latest version of self_update

lionel1704:
- use latest version of self_update with the rustls feature. Enabling
this feature will remove the requirement for the openssl dependency
which in turn will help re-enable MUSL builds


iancoleman:
Is Dockerfile.build still in use? If so these lines for openssl should also be removed:

https://github.com/maidsafe/safe_vault/blob/b80899e5c84c9ab7467401805491242a5155138e/Dockerfile.build#L34-L40

PR #942, opened 2020-02-19
Update version to 0.20.1

lionel1704:
Version 0.20.1 of the safe_vault was released from the master branch. This change updates the version so that the binary isn't replaced by self_update.


Cargo.toml
```diff
@@ -8,7 +8,7 @@ license = "GPL-3.0"
 name = "safe_vault"
 readme = "README.md"
 repository = "https://github.com/maidsafe/safe_vault"
-version = "0.19.2"
+version = "0.20.1"
```
maqi: A non-blocking comment:
as there was no 0.20.0 before, maybe this doesn't need to be 0.20.1 ?

lionel1704: The version in the master branch is `0.20.1` so if the version doesn't match then self_update will replace the binary with the one released from master.

PR #963, opened 2020-03-25
feat/sequence: first draft implementation of a Sequence data type as CRDT

bochaco:


Thierry61:
We are we are losing the consistency of previous naming convention that I mentioned here https://safenetforum.org/t/rfc-data-types-refinement/30508/42?u=tfa.

For example, I see the following translations of previous names:
-	`AData` becomes `Sequence`
-	`get_adata` becomes `get_seq_data`

This is not consistent internally, but also externally with the 2 other data structures (`IData` and `MData`).

To keep consistency, why not call the sequence structure `SData` and associated suffix `_sdata` for names that need snake case?

The single letter I mentioned in the post would become `S`, `M` or `I` for respectively `Sequence`, `Mutable` or `Immutable`.

bochaco:
This is superseded by PR #982

PR #975, opened 2020-04-17
Modularize client handler

oetyng:
Addressing sprawly levelling, with all kinds of detail, high to low, and fuzzy boundaries and structures in `ClientHandler`. 

We identify a top level, `ClientHandler`, and within that a few larger areas; `messaging`, `auth`, `login_packets` and `data`. These components encapsulate their details. This way, a quick glance on top level, gives a fair view of what it does, and keeping this practice down the line, will give the same down those levels.
It's about breaking up problems in sub-problems, and organize them together in components where they are necessary and relevant to each other, and only expose a minimum of interface between these components.

***

`ClientHandler` has got split up into the following sub-modules:
- `Messaging`
- `Auth`
- `LoginPackets`
- `ElderData`

(Along the previously existing `AuthKeys` and `Balance`)

`TransactionId` is taken to simply be `MessageId`.
`CoinsHandler` is removed.

The temporarily removed feature is `refund`. I think there might be some additional behavioral change in addition to that, but it really is all good in testnets IMO, and no problem to introduce it at a later point when it is prioritized.

As to separate the refactor from the feat impl of `at2`, next step is that `Balance` is removed and `Banking` is introduced instead.


src/chunk_store.rs
```diff
@@ -42,7 +42,7 @@ const MAX_CHUNK_FILE_NAME_LENGTH: usize = 104;
 pub(crate) type ImmutableChunkStore = ChunkStore<IData>;
 pub(crate) type MutableChunkStore = ChunkStore<MData>;
 pub(crate) type AppendOnlyChunkStore = ChunkStore<AData>;
-pub(crate) type LoginPacketChunkStore = ChunkStore<LoginPacket>;
+pub(crate) type LoginPacketDb = ChunkStore<LoginPacket>;
```
lionel1704, line 45: This isn't exactly a DB. LoginPackets are stored as files on disk so maybe this will cause some confusion?

oetyng: Yeah I saw that. Before I saw that I just thought the name was unnecessarily unwieldy.
Just putting a "Db" postfix conveys that it is some sort of storage in a concise and commonly understood way.

The fact now that it is stored as files on disk, I'm not sure that is something we would want to express in the name?

Also I wondered when I saw that, why are we writing to files on disk, as opposed to some other way?

Well.. this is not very important atm, but this "ChunkStore" thing doesn't seem that clean to me actually.

madadam: What about just `Store`?

src/client_handler.rs
```diff
@@ -43,22 +43,24 @@ use std::{
...
-    auth_keys: AuthKeysDb,
-    balances: BalancesDb,
-    clients: HashMap<SocketAddr, ClientInfo>,
-    pending_msg_ids: HashMap<MessageId, SocketAddr>,
-    pending_actions: HashMap<MessageId, Response>,
-    // Map of new client connections to the challenge value we sent them.
-    client_candidates: HashMap<SocketAddr, (Vec<u8>, PublicId)>,
-    login_packets: LoginPacketChunkStore,
-    routing_node: Rc<RefCell<Node>>,
+    messaging: Rc<RefCell<Messaging>>,
```
madadam, line 48: Why is this in `Rc<RefCell>`?

src/client_handler.rs
```diff
@@ -43,22 +43,24 @@ use std::{
...
+    auth: Auth,
+    login_packets: LoginPackets,
+    data: ElderData,
+}
+
+pub struct Responder {
+    messaging: Rc<RefCell<Messaging>>,
```
madadam, line 55: Same here, why the `Rc<RefCell>`?

madadam: OK, I see. Because it's shared between `ClientHandler` and `Responder`. Wonder if there is a better way...

src/client_handler.rs
```diff
@@ -182,79 +161,25 @@ impl ClientHandler {
...
-                Err(err) => {
-                    info!(
-                        "{}: Unable to deserialise handshake request from {}: {}",
-                        self, peer_addr, err
-                    );
-                }
-            }
+        match self
+            .messaging
+            .clone()
```
madadam, line 166: Don't think this `clone` is needed.

src/client_handler.rs
oetyng: Yeah me too. This is exactly one area where I suspect it could be done better.

So, first of all, there are 3 layers of nested `Rc<RefCell>>`; Node, Messaging, Responder.. I wonder if that can be done differently.

Responder was a way to encapsulate communication that is made from within the methods.
I tried to pass in function pointer (didn't work since it wraps an environment variable and would require passing along the messaging instance anyway). Passing in a closure just got bogged down in endless lifetime specifiers, didn't get that sorted out.

So, I created the Responder, which at least was less fuss than that.

But... all that said, I think the mutation within the methods is something to move away from.
I want to move to side-effect free calls to the sub-modules, and get events/cmds returned that are passed upwards and applied or sent on the wire in dedicated modules.

src/client_handler.rs
oetyng: Unfortunately it was, well, maybe not with some other solution, but just removing it yields this:

```
cannot borrow `*self` as mutable because it is also borrowed as immutable
   --> src\client_handler.rs:170:29
    |
164 |            match self
    |   _______________-
    |  |_______________|
    | ||
165 | ||             .messaging
    | ||______________________- immutable borrow occurs here
166 | |              //.clone()
167 | |              .borrow_mut()
    | |__________________________- a temporary with access to the immutable borrow is created here ...
...
170 |                Some(result) => self.initiate_client_request(
    |  ______________________________^
171 | |                  &result.client,
172 | |                  result.request,
173 | |                  result.message_id,
174 | |                  result.signature,
175 | |              ),
    | |______________^ mutable borrow occurs here
...
178 |        }
    |        - ... and the immutable borrow might be used here, when that temporary is dropped and runs the destructor for type `std::cell::RefMut<'_, client_handler::messaging::Messaging>`
    |
    = note: The temporary is part of an expression at the end of a block. Consider forcing this temporary to be dropped sooner, before the block's local variables are dropped. For example, you could save the expression's value in a new local variable `x` and then make `x` be the expression at the end of the block.
```

src/client_handler.rs
```diff
@@ -71,108 +73,85 @@ impl ClientHandler {
...
+                message_id,
+            } => self.finalize_client_request(src, requester, request, message_id),
+            Rpc::Response {
+                response,
+                requester,
+                message_id,
+            } => self
+                .messaging
+                .borrow_mut()
+                .relay_data_handler_reponse_to_client(src, &requester, response, message_id),
```
lionel1704, line 127: This is not always the data handler. It might be Elders of a different section when it comes to coin operations.

oetyng: OK. Took it from the comments in the previous code:

```
/// Handle response from the data handlers.
    fn handle_response(
```

(this is the method called from that branch arm in `handle_vault_rpc`)

src/client_handler/elder_data.rs
```diff
@@ -0,0 +1,351 @@
...
+    NodePublicId, Request, Response,
+};
+use std::{
+    cell::RefCell,
+    fmt::{self, Display, Formatter},
+    rc::Rc,
+};
+
+#[derive(Clone)]
+pub(crate) struct ElderData {
```
lionel1704, line 27: Looks like this is used to only validate the data requests before they are passed on to the Data Handlers.
Can this be just `DataRequestValidator` which validates requests for all three types of data.

src/client_handler/elder_data.rs
```diff
@@ -0,0 +1,354 @@
...
+        self.idata
+    }
+
+    pub fn mdata(self) -> ElderMData {
+        self.mdata
+    }
+
+    pub fn adata(self) -> ElderAData {
+        self.adata
+    }
```
madadam, line 52: This seems weird. So you are basically consuming this struct every time you want to use it which forces you to clone it. Why not remove these accessors altogether and instead make the fields themselves public? Alternatively change the accessors to return mutable references instead:
```rust
pub fn idata(&mut self) -> &mut ElderIData {
    &mut self.idata
}
```
Although I would not recommend that. Just exposing the fields is simpler, straightforward and involves less code.

src/client_handler/elder_data.rs
lionel1704: Also. This seems to be cloned every time it is used. That should be avoidable.

src/client_handler.rs
madadam: So just to elaborate: I usually prefer to avoid `RefCell` if possible, because it changes the borrow checking from compile-time to runtime and so makes it harder to reason about the correctness of the code. Where without it you let the compiler do the hard work, with it now you have to do it (writing more unit tests, etc...).

Not saying it's necessarily wrong doing it this way. Just something worth thinking about.

src/client_handler/elder_data.rs
oetyng: It's not even validating actually. It's just a higher level grouping to keep down the noise at `ClientHandler` level.

I didn't want to replicate all methods from the sub components in `ElderData`, so ended up just letting them be gotten via a method call.
I think the few extra lines in that file is worth it to keep it tidier at upper level.

src/client_handler.rs
```diff
@@ -182,79 +161,25 @@ impl ClientHandler {
...
-                    );
-                }
-            }
+        match self
+            .messaging
+            .clone()
+            .borrow_mut()
+            .try_initiate_client_request(peer_addr, bytes, rng)
+        {
+            Some(result) => self.initiate_client_request(
```
lionel1704, line 170: "Initiate" seems a liiiitle out of place here. Perhaps this should be `process_client_request` since we're not exactly initiating the request here. And `try_initiate_client_request` can be `try_fetch_client_request`. Since it implicitly handles the handshake requests and only returns requests for data/coin/auth operations. Would be nice to make that distinction here.

src/client_handler.rs
madadam: I see. Maybe separate the `try_initiate_client_request` call from the `match`:
```rust
let result = self.messaging.borrow_mut().try_initiate_client_request(...);
match result {
    Some(result) => self.initiate_client_request(...),
    None => None,
}
```
?

src/client_handler/elder_data.rs
oetyng: Yeah I thought something was iffy there.
What I would want is really a public read only getter. If I've understood it correctly, using the method call like you describe (so `&mut`, that's what I actually wanted) is the closest we get to that in rust?

Not sure if there are other ways that this is dealt with in rust, but I'm used to be able to have immutable structs, where outside cannot change fields. There's not really any state to be changed here, and can't set them to null so.. seems maybe not that much sense in trying to achieve that here. Just a habit to always protect the fields.

Anyway, seems pub fields would be fine here.

src/client_handler.rs
oetyng: I agree, I thought this as well, wanted to follow the pattern of the subsequent method calls.. but you're right `process_client_request` is better here.

About `try_fetch_client_request`. The name is maybe good. But, there's something about that that I'm not fully clear about now when you mention 
> implicitly handles the handshake requests and only returns requests for data/coin/auth operations

Seems that would be a bit too much responsibility there.
Will take your suggestions but need to look more into that method.

src/client_handler/auth.rs
```diff
@@ -0,0 +1,300 @@
...
+            .list_auth_keys_and_version(utils::client(&client.public_id)?));
+
+        self.responder
+            .borrow_mut()
+            .respond_to_client(message_id, Response::ListAuthKeysAndVersion(result));
+        None
+    }
+
+    // on client request
+    pub fn initiate_auth_key_insert(
```
lionel1704, line 104: Same comment about initate here. Finalize fits in superbly. I'll try to think of something to replace initiate.
How about `commence_insert_auth_key` ?

src/client_handler.rs
madadam: The events/command thing might be way to go. Alternatively, you can try making the messaging/responder to be owned only by `ClientHandler` and pass them as arguments to the functions that need them. This is what I did in routing with the `Core` struct. Not saying it's particularly pretty but at least it plays nicely with the borrowing/ownership rules of rust.

src/client_handler/auth.rs
oetyng: Yeah, it's the node process steps I want to verbalize and make clear in the code naming.

I think it's fine to say that the node initiates the auth key insert. The client _requests_ it, and the node initiates the actual key insert process, which is a multi step process. 

So,` insert auth key` is to be interpreted as a _process_, and `initiate` refers to the process.

And so, the Client requests an initiation of the process.

This is quite common when doing `eventsourcing/cqrs`, the cmds are called things like `InitiateProcess`, and the actual method that is called after the cmd has been deserialized at recipient, has the same name as the cmd.

All that said, commence is not a bad word either, but I think it shoots besides the target and is not necessary.

oetyng:
> Looks mostly good. The `ElderData` struct is a bit weird though. Also I would prefer avoiding the shared ownership of `Messaging` and `Responder` if possible.

>So just to elaborate: I usually prefer to avoid RefCell if possible, because it changes the borrow checking from compile-time to runtime and so makes it harder to reason about the correctness of the code. Where without it you let the compiler do the hard work, with it now you have to do it (writing more unit tests, etc...).

> Not saying it's necessarily wrong doing it this way. Just something worth thinking about.

Aha. Good to know. I didn't know actually what implications the use of RefCell would have.
I would happily see a better solution to the shared ownership.

I think my ultimate goal will let us do that. So, to clarify: this componentization is a multi step process, and I could have done more, but stayed here for the first step.

**First step**
As we separate messaging out from the big ball of mud, we necessarily need a way to call the specific messaging methods, since those are used in all the other components. I wanted to just pass in the actual method calls to the other components, because it's not right that they would know of the entire messaging component with all its interface - since they are not supposed to use all that.

**Second step**
Now, the way I would prefer to do all this, is that each call to a component returns a msg (i.e. an event or cmd), which is routed to the appropriate module.
That way the decoupling is complete.

But, that seemed like a step too far for this first step. Would be happy to carry it on soon though, I think this system needs it dearly.. it's a pattern I think we should have, and that way there is no shared reference of any module.

src/client_handler/elder_data.rs
lionel1704: I was just thinking if we could make the change to something like:
```
struct ClientHandler {
     data_request_agent: DataRequestAgent,
     ...
}
struct DataRequestAgent {
      ....
}
impl DataRequestAgent {
 fn commence_*data_operation() {
          validate_request()
          check_permissions()
          send_for_payment()
          ...
      }
}
```

So just a single struct for all data types since they don't do too much. Will reduce the number of copies of ID and Responder being held.

src/client_handler/elder_data.rs
lionel1704: > pub fields would be fine here.

Or even functions that take `&self` and return `&WhatWeNeed`.
If we need a copy `WhatWeNeed` then it can be cloned by whoever does the call.

src/client_handler/login_packets.rs
```diff
@@ -0,0 +1,258 @@
...
+            .login_packet(utils::own_key(client_id)?, address)
+            .map(LoginPacket::into_data_and_signature);
+        self.responder
+            .borrow_mut()
+            .respond_to_client(message_id, Response::GetLoginPacket(result));
+        None
+    }
+
+    // on client request
+    pub fn initiate_login_packet_creation(
```
lionel1704, line 58: Initiate fits in well here :)
We are initiating the creation and not the request itself so it makes full sense. :+1:

src/client_handler/elder_data.rs
oetyng: I think the efficiency we're trying to get by doing that is short term though. The main goal is to separate concerns, not to bundle them all up as trying to get (maybe negligible) performance gains.

This is a step one of componentization.
If we fulfil the steps, we won't be holding any `Responder`. It is a temporary workaround and not what I would consider the best solution. But necessary as to split up, while not going all the way in one single PR.

About the naming there, `DataRequestAgent`, I really don't think it's good. I'm trying to move away from that type of technical generic stuff and be as direct as possible. This is about Elder data, so `ElderData` seemed fairly accurate and concise (even though probably can be better).

`DataRequestAgent` makes me think of this:
EC49F39WwAA1Z3Y

src/client_handler.rs
oetyng: > messaging/responder to be owned only by ClientHandler and pass them as arguments to the functions that need them
> plays nicely with the borrowing/ownership rules of rust.

Yeah, with regards to racing conditions, that is top notch. But it doesn't solve encapsulation and keeping minimal available interfaces / possible paths at any given level. Why do I say that? Because if we pass in the messaging instance, we give the recipient component access to all the methods there, even though that is certainly not what we want. We just need to allow the component to use one of the messaging API:s.

This is where I see messaging (events/cmds) shine. You inverse the dependencies, since the components do not hold the other components, they are not coupled to the types. We call a method on such a component, and it does evaluation, side effect free, and returns some kind of msg (event/cmd) depending on what we want there. If we want a state change on that instance it is an event that we apply to that instance. If we want to send some msg on the wire, it might be a cmd returned by the method call (like `NotifyClient`) instead of that call happening within the method. That cmd is then sent to the messaging component.

None of this breaks the rust borrowing/ownership rules, and we do not need shared references.

So, this `Responder` thing is a intermediate step that I do not like, but while not going the full way with the decoupling in one single PR, there is no other way to split these components up, than to pass in something to them. And so I preferred a wrapper that minimizes the interface, instead of passing in the full API of messaging where it is not intended to be used.

src/client_handler/elder_data.rs
oetyng: > Or even functions that take &self and return &WhatWeNeed.

Yeah, that would probably be the closest to a public read only getter  :+1:, this is what I really wanted.

lionel1704:
Nice work with this. After taking the points raised into consideration we can try and get this merged so that all future work is based on this structure.

One additional concern I'd like to raise is the removal of balances completely. There is a `balance.rs` but it isn't used anywhere. I'd prefer to keep the `coins_handler` file in there too. It will be unused till we start implementing Safecoin farming, but I'd still like to keep it in there with a comment saying that's it's kept for farming.

src/client_handler/login_packets.rs
oetyng: Yeah, so maybe all the methods should be clear about being about a process that is initiated.
The request is sort of at a higher level. From this point of view, we're like already in the machinery and we're in another context now. The machinery only knows what buttons was pressed on it, not what happened outside of that (such as _the client initiating a request_, it has no cognition of that). That's why the naming here is just about this context: This machine initiates and completes multi-step processes.

So: `initiate_login_packet_creation`, and `initiate_auth_key_insertion`.

The insertion and creation is something the machine does a process it is responsible for, and that's why **the machine** is initiating that process.

The client just sends a cmd _requesting_ the machine to initiate this or that. So, it's not about initiating a request, but about initiating the process.

As you can notice above, I now saw that `initiate_auth_key_insert` should be renamed to `initiate_auth_key_insertion` to fully reflect what I'm talking about above. Will be looking at all the other methods and make sure they follow the same pattern.

oetyng:
> Nice work with this. After taking the points raised into consideration we can try and get this merged so that all future work is based on this structure.
> 
> One additional concern I'd like to raise is the removal of balances completely. There is a `balance.rs` but it isn't used anywhere. I'd prefer to keep the `coins_handler` file in there too. It will be unused till we start implementing Safecoin farming, but I'd still like to keep it in there with a comment saying that's it's kept for farming.

Awesome, nice.

About the `balance.rs`, yeah. It should have been removed :+1:, just forgot about it actually.

`coins_handler.rs` though it has almost nothing in there (so nothing lost by deleting it), and is not even used. So I think it's much better to not assume how farming will be implemented. We'll look at farming with fresh eyes when we get to that, and not try fit it into what ever `coins_handler.rs` was assuming.

src/client_handler.rs
oetyng: Changing it to this, if that's more correct:

```
/// Relay response from other node to the client.
    pub fn relay_reponse_to_client(
```

src/client_handler.rs
oetyng: Aah..! Much better, that did the trick :)

src/chunk_store.rs
oetyng: I'll leave it unchanged for now, and we can look at this later.  I'm still wondering why are we writing to files on disk, as opposed to some other way.

I suspect this "chunkstore" thing will need to be revised when we **A.** use crdt **B.** potentially introduce the "everything is a chunk" data hierarchy refinement.

src/client_handler.rs
oetyng: Went with `try_parse_client_request`. Seems appropriate / good enough.

tests/integration.rs
```diff
@@ -338,7 +338,7 @@ fn update_login_packet() {
 ////////////////////////////////////////////////////////////////////////////////
 
 #[test]
-fn coin_operations() {
+fn balances() {
```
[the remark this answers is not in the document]
oetyng: It was to begin with, but I actually saw it and thought it was an OK rename considering the module now has that name.
No biggie so can just as well revert it.

oetyng:
> the ElderData struct.
> 
> * I think the name doesn't fit very well. This module holds the code for the validation done at the client handler for incoming data requests. I don't think the current title delivers the right message. Let's try to think of something else.

So, this is a deeper discussion.

To begin with:
These modules all contain validation, but none of them are about validation per se, and they don't do _just_ validation, they are actually more of _decision makers_ than merely validators. (If it wasn't for the decision making aspect of them, I'd be OK with calling them validators for example, but now it's missing the point).
The contexts of them are auth keys, balances, data requests, messaging and login packets.

The data requests module is a bit anemic at the moment, there's hardly any behavior there (many methods do no validation or decision making at all). In fact, in it's current form it's mostly about making the code in `ClientHandler` a bit tidier, so just separating things and organizing them within coherent categories. But this bucket doesn't have much in it.

About `ElderData` name:
Absolutely, it is not optimal. As I said before there should be a better name available. Putting `Elder` there was an attempt to start moving away from the `ClientHandler/DataHandler` idea and start using the `Elder` and `Adult` concepts more, something David has requested.

I'm totally OK with deferring that and do a larger overhaul on that, the only problem with that kind of deferring is that it perpetuates what needs to change, and every line of code added with the old mindset will make it harder to move to the newer.. (Some time small steps is the cure, but I'm okay with doing a big bang effort as well, the important thing is to actually do something.)

`ElderData` is supposed to signify that this is the data context at Elders. It has risk of ambiguous meaning ("Is it data of Elders?"), I agree. We could call the sub-module `data_requests`, and the implementation `Evaluation` (even though that is also inadequate IMO). Describing the impl behavior instead of a noun is a deliberate choice.

I think that makes an OK description within this:
**ClientHandler has decision making sub-modules, they make decisions in the contexts of auth keys, balances, data requests, messaging and login packets.**

It's not perfect either. Data request evaluation is a good description, but it's a long name for an implementation (hence allowing for the module/type hierarchy to speak for itself; `data_requests::Evaluation`). 
It's a compromise, but maybe we can be happy for now and spend more time on higher priority things. I have a feeling much of this will change once we start getting CRDTs in place, so doesn't seem worth too much effort at this point.


> the ElderData struct.
> 
> * I agree that it's good to have separate structs for the  IData / MData / AData request validators. But I think they can be abstracted at the upper level by something like this:
> 
> ```
> struct ElderData {
>      idata: ElderIData,
>      mdata: ElderMData,
>      adata: ElderAData
> }
> impl ElderData {
>      fn do_idata_stuff(&self) {
>           self.idata.do_stuff();
>      }
> }
> ```
> 
> So in the client handler, we can do something like:
> 
> ```
> Request::PutIData() => {
>      self.data.do_idata_stuff()
> }
> ```
> 
> What do you think?

Yeah we could do that, but I don't think much is won by that. There are drawbacks with both ways IMO, and I don't feel it is very important at the moment. That module is a bit strange from the beginning I think, also it's likely things will change there with crdts, but I think the main purpose of cleaning up `ClientHandler` has been achieved now and that we can be happy with that as a first step.

lionel1704:
Agreed @oetyng. In that case let's just address the other comments. Most of them are trivial, but [this one](https://github.com/maidsafe/safe_vault/pull/975#discussion_r417771585) is a bug.
Once that's updated let's get this merge :+1:

oetyng:
> Agreed @oetyng. In that case let's just address the other comments. Most of them are trivial, but [this one](https://github.com/maidsafe/safe_vault/pull/975#discussion_r417771585) is a bug.
> Once that's updated let's get this merge 👍

Ah, I had this committed to the other branch, hold on I'll merge it into this.

PR #977, opened 2020-05-04
refactor/handlers: adapt to changes made on safe-nd Request enum

bochaco:


src/data_handler.rs
```diff
@@ -30,7 +29,6 @@ use std::{
 
 pub(crate) struct DataHandler {
     id: NodePublicId,
-    idata_holder: IDataHolder,
     idata_handler: Option<IDataHandler>,
     mdata_handler: Option<MDataHandler>,
     adata_handler: Option<ADataHandler>,
```
lionel1704, line 34: The `IDataHandler` and `IDataHolder` are two different personas that a Vault may have. An Adult will have an `IDataHolder` but it will not have any of the other three. That's why they are optional and the holder is not.

With this change the logic for Adults will break so could we bring back the IDataHolder to this level please?

src/client_handler/balances/mod.rs
```diff
@@ -32,19 +32,62 @@ impl Balances {
...
     }
 
-    pub(super) fn initiate_creation(
+    // on client request
+    pub(super) fn process_client_request(
+        &mut self,
+        client: &ClientInfo,
+        request: CoinsRequest,
+        message_id: MessageId,
+        messaging: &mut Messaging,
```
oetyng, line 41: Passing in the messaging module like this into the other client handler modules, is something we wanted to avoid by the previous refactor. It's simple to do, but it's also undoing the efforts to make this part of the code more modular.

src/client_handler/login_packets.rs
```diff
@@ -69,7 +103,76 @@ impl LoginPackets {
...
+                    "{}: Should not receive {:?} as a client handler.",
+                    self, request
+                );
+                None
+            }
+        }
+    }
+
+    // on consensus
+    fn finalize_creation(
```
oetyng, line 175: If there's a renaming of `finalize` to `finalise` going on here as well, this is one more. I would search to make sure all is replaced.

src/client_handler/balances/mod.rs
oetyng: There's no easy solution to this within the scope of this PR, but on the other hand, that's how code degrade. It's easily done and overlooked, and the difficult solution is more or less never within scope of the PR at hand.

There are plans for next steps in decoupling these modules, which includes side effect free methods returning messages intended for the other modules to act on.

bochaco: sounds good I'll be looking forward to that PR

PR #980, opened 2020-05-11
store and retrieve idata from adult node

ravinderjangra:


src/data_handler.rs
```diff
@@ -102,37 +102,37 @@ impl DataHandler {
...
                             self.handle_idata_request(|idata_handler| {
                                 idata_handler.handle_put_idata_req(requester, data, message_id)
                             })
                         }
                     }
                     IDataRequest::Get(address) => {
-                        if &src == address.name() {
+                        if matches!(requester, PublicId::Node(_)) {
                             // The message was sent by the data handlers to us as the one who is supposed to store
                             // the chunk. See the sent Get request below.
```
lionel1704, line 120: This comment needs to be updated.

src/data_handler.rs
```diff
@@ -102,37 +102,37 @@ impl DataHandler {
...
                                 idata_handler.handle_get_idata_req(requester, address, message_id)
                             })
                         }
                     }
                     IDataRequest::DeleteUnpub(address) => {
-                        if &src == address.name() {
-                            // Since the src is the chunk's name, this message was sent by the data handlers to us
+                        if matches!(requester, PublicId::Node(_)) {
+                            // Since the requester isa node, this message was sent by the data handlers to us
                             // as a single data handler, implying that we're a data handler where the chunk is
```
lionel1704, line 132:
```suggestion
                            // as a single data handler, implying that we're a data holder where the chunk is
```

src/data_handler/idata_holder.rs
```diff
@@ -63,7 +66,8 @@ impl IDataHolder {
 
         let refund = utils::get_refund_for_put(&result);
         Some(Action::RespondToOurDataHandlers {
-            sender: *self.id.name(),
+            // this is actually the target
+            sender: source,
```
lionel1704, line 70: The `sender` field in this Action variant should be renamed to `target`.

PR #981, opened 2020-05-11
WIP updated get idata_holder/idata flow

ravinderjangra:
Blocked by #980


ravinderjangra:
Needs more discussion on the security aspects when the client/app tries to fetch the content.

PR #982, opened 2020-05-12
feat/data: support for Sequence data type and associated requests

bochaco:


bochaco:
This is now sent as PR #1004 against master branch.

PR #983, opened 2020-05-13
auto generate dependency graphs and list duplicate dependencies via CI

S-Coyle:
**Clippy CI warnings causing fail** - these have been resolved in the vaults `phase 2b` branch, so I assume I don't need to resolve here as well.

**Dependency Graphs**
Adding dependency graphs will hopefully help the community and beyond visualise how our crates are interacting with each other, and how they are interacting with non-MaidSafe crates.

These graphs will update every night via CI, if any changes.

Note that with safe vault the `all dependencies` graph is very complex, not sure how much value it offers, but if it's minimised in a section then IMO there's no harm in giving readers the option to view if they want. Plus not all crates will be this complex so if this PR is accepted and we roll this out to other crates it would be good to be consistent and give the option to view all dependencies in all of them.

Tested on a fork & the CI ran & generated the graphs successfully, which were then displayed in the readme.

Preview of internal MaidSafe safe_vault dependencies graph:
image

Preview of all safe_vault dependencies graph:
image

**List Duplicate Dependencies**
I considered listing dependencies using `cargo tree` but thought the graphs generated by `cargo deps` would be better suited for the target audience in a README. `cargo tree` however has a useful `--duplicates` option which may be useful for our developers seeing and being aware of so I've added a stage in the PR CI which will list all duplicate dependencies. This will not cause any failures if duplicates are found, the list is for information only.
`List Duplicate Dependencies` can be viewed in the CI results for this PR.

### DIFF
--- a/scripts/generate_dependency_graphs
+++ b/scripts/generate_dependency_graphs
@@ -1,0 +1,11 @@
+#!/bin/bash
+
+set -e -x
+
+rm -rf images
+mkdir images
+
+cargo install cargo-deps
+
+cargo deps --all-deps --include-orphans --filter safe_vault safe-nd parsec maidsafe_utilities routing mock-quic-p2p quic-p2p lru_time_cache fake_clock | dot -T png -Nfontname=Iosevka -Gfontname=Iosevka -o images/safe_vault_maidsafe_dependencies.png
+cargo deps | dot -T png -o images/safe_vault_all_dependencies.png
--- a/scripts/generate_dependency_graphs
+++ b/scripts/generate_dependency_graphs
@@ -7,5 +7,5 @@ mkdir images
 
 cargo install cargo-deps
 
-cargo deps --all-deps --include-orphans --filter safe_vault safe-nd parsec maidsafe_utilities routing mock-quic-p2p quic-p2p lru_time_cache fake_clock | dot -T png -Nfontname=Iosevka -Gfontname=Iosevka -o images/safe_vault_maidsafe_dependencies.png
+cargo deps --all-deps --include-orphans --filter safe_vault safe-nd parsec maidsafe_utilities routing mock-quic-p2p quic-p2p lru_time_cache fake_clock xor_name bls_dkg safe-network-signature-aggregator | dot -T png -Nfontname=Iosevka -Gfontname=Iosevka -o images/safe_vault_maidsafe_dependencies.png
 cargo deps | dot -T png -o images/safe_vault_all_dependencies.png
--- a/.github/workflows/generate_dependency_graphs.yml
+++ b/.github/workflows/generate_dependency_graphs.yml
@@ -1,0 +1,30 @@
+name: Build and Deploy
+on:
+  push:
+    branches:
+      - master
+  schedule:
+    # Every day at 1 AM
+    - cron:  '0 1 * * *'
+
+jobs:
+  build-and-deploy:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2 # If you're using actions/checkout@v2 you must set persist-credentials to false in most cases for the deployment to work correctly.
+        with:
+          persist-credentials: false
+
+      - name: Install dot
+        run: sudo apt-get install graphviz
+
+      - shell: bash
+        run: ./scripts/generate_dependency_graphs
+
+      - name: Deploy
+        uses: JamesIves/github-pages-deploy-action@releases/v3
+        with:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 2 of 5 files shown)


scripts/generate_dependency_graphs
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+set -e -x
+
+rm -rf images
+mkdir images
+
+cargo install cargo-deps
```
dan-da, line 8: This looks very slick!  Thanks for tackling it.   Can you post a sample of generated output please?

I see you went with cargo-deps instead of cargo-tree.   I'm guessing that is because you wanted to create an SVG instead of plain text?

I'm wondering if cargo-deps also can show the dependencies that are included multiple times with different versions, like cargo-tree does.   To me, that is a useful thing to track, as it points out some fairly low hanging-fruit for reducing binary footprint and removing dup/older code.

scripts/generate_dependency_graphs
```diff
@@ -0,0 +1,11 @@
...
+
+cargo deps --all-deps --include-orphans --filter safe_vault safe-nd parsec maidsafe_utilities routing mock-quic-p2p quic-p2p lru_time_cache fake_clock | dot -T png -Nfontname=Iosevka -Gfontname=Iosevka -o images/maidsafe_dependencies.png
```
dan-da, line 10: I don't understand why this cargo deps command is so long.   What does the filter arg do?  I guess that command just seems brittle to me, like something that needs to be maintained over time.

With cargo tree, one simply enters the crate of choice and runs `cargo tree`.

m-cat: It limits included dependencies to the ones specified, in this case only MaidSafe dependencies. Yeah, it's brittle, I want it to support filtering by Cargo.toml attributes like "author", but I don't have much time or interest to add new features.

scripts/generate_dependency_graphs
m-cat: Does --include-versions (iirc) do what you want?

S-Coyle: Ah yes, previews would definitely help - now added to the description.

Yes I used `cargo tree` and that was fine, but I thought that the internal MaidSafe safe_vault dependencies just looked easier on the eye when visualised in a graph. The all dependencies 1 is a difficult read no matter what way it's represented because of its complexity, but hopefully not all repos will be this complex.
I concluded that the target audience for this, if added to a readme, was probably not our own internal development team, it was more likely to be newcomers to the project or external developers looking for some sort of insight. This was also the reason I didn't include `cargo tree --duplicate` - it just didn't feel to me like it belonged in a readme.
PS As per MArcin's suggestion, I'll post the results of using `--include-versions      Include the dependency version on nodes` to generate these graphs in a sec.

All just 100% my own opinions, I can be talked into just sticking with `cargo tree` and adding `cargo tree --duplicate`, adding these elsewhere, etc. Let me know what you think.

scripts/generate_dependency_graphs
S-Coyle: Yep so the actual command is just `cargo deps` but for line 10 above I wanted to filter down to only show the MaidSafe repos. I was just thinking that some people may like to see the internal MaidSafe deps only. Doesn't look like I can filter like this using `cargo tree`.

Agreed this means a maintenance overhead in future, so it's a pro's vs con's debate and given the discussions on the forum I judged the pro's outweighed the con's for this. 

You can see in line 11 where the default graph is produced it is a much shorter command.

scripts/generate_dependency_graphs
S-Coyle: examples of `cargo deps --include-versions`
image

image

m-cat: > I'm wondering if cargo-deps also can show the dependencies that are included multiple times with different versions, like cargo-tree does. To me, that is a useful thing to track, as it points out some fairly low hanging-fruit for reducing binary footprint and removing dup/older code.

@dan-da I just remembered that cargo-deps will detect when a dependency is included twice with different versions and include the version in their labels, even if --include-versions isn't passed.

S-Coyle:
@dan-da I've made a few changes and done a bit of tidying around this PR since you last commented, if you get a chance to take another look:
- I've added `cargo tree -d` to a `List Duplicate Dependencies` CI stage which runs against every PR.
- I've displayed the internal MaidSafe safe_vault dependencies graph in the README (was previously hidden in a section), keeping the all deps graph hidden in a section.
- I've added a link from the README to a graph with an overview of the interdependencies between all of the main MaidSafe components.
- I've switched the dependency graphs to be generated on a new upstream branch `png_generator` as I discovered that trying to get CI to push to a protected upstream branch was a minefield.

PR #999, opened 2020-06-15
Fix wrong / meaningless log messages

lionel1704:


src/data_handler/idata_handler.rs
```diff
@@ -617,7 +618,7 @@ impl IDataHandler {
         let is_idata_copy_op = self.idata_elder_ops.contains(&message_id);
 
         let action = if is_idata_copy_op {
-            info!("Got a copy action for IData");
+            info!("Adult(s) responded with chunk for duplication");
```
ravinderjangra, line 621: maybe be an elder as well?

PR #1000, opened 2020-06-15
update routing dep and use updated our_elders_sorted_by_distance_to API

ravinderjangra:


tests/common/mod.rs
```diff
@@ -174,6 +176,13 @@ impl Environment {
             self.poll();
         }
     }
+
+    #[cfg(feature = "mock_parsec")]
+    pub fn get_closest(&mut self, target: &XorName) -> Vec<XorName> {
```
lionel1704, line 181: What are we getting here?
Adults / Elders / Data?

ravinderjangra: elders, I'll update the PR once the routing PR is merged

PR #1008, opened 2020-06-24
Remove caching of ahead of time responses and implement BLS signature accumulation in Vaults.

lionel1704:


src/vault.rs
```diff
@@ -371,27 +390,95 @@ impl<R: CryptoRng + Rng> Vault<R> {
...
+                        };
+                        self.data_handler_mut()?.handle_vault_rpc(
+                            SrcLocation::Section(prefix),
+                            accumulated_request,
+                            Some(signature),
+                        )
+                    } else {
+                        info!("Insufficient sigs");
+                        None
+                    }
```
ravinderjangra, line 478: This section seems to be repeated multiple times maybe we can convert this into a local function here

src/data_handler/mod.rs
```diff
@@ -170,32 +208,83 @@ impl DataHandler {
             src,
             requester
         );
-        match request {
+        match request.clone() {
             IData(idata_req) => {
                 match idata_req {
                     IDataRequest::Put(data) => {
                         if src.is_section() {
```
ravinderjangra, line 215: Since we are not using the section to send/receive msgs, will this still work? same for the next two request types.

src/data_handler/mod.rs
```diff
@@ -170,32 +208,83 @@ impl DataHandler {
...
+                                None
+                            }
+                        } else if matches!(requester, PublicId::Node(_)) {
+                            if self
+                                .routing_node
+                                .borrow()
+                                .public_key_set()
+                                .ok()?
+                                .public_key()
+                                .verify(accumulated_signature.as_ref()?, utils::serialise(&address))
```
ravinderjangra, line 270: How the `accumulated_signature` verification is working when we are getting the request only from a single node in the section?

src/accumulator.rs
```diff
@@ -0,0 +1,183 @@
...
+
+                        let (_, _, signatures) = self.messages.get(&message_id)?;
+                        let public_key_set =
+                            self.routing_node.borrow().public_key_set().ok()?.clone();
+                        info!(
+                            "Got {} signatures. We need {}",
+                            signatures.len(),
+                            public_key_set.threshold() + 1
+                        );
+                        if signatures.len() > public_key_set.threshold() {
```
ravinderjangra, line 79: Should this be `>=`?

src/data_handler/idata_handler.rs
```diff
@@ -159,92 +155,30 @@ impl IDataHandler {
...
-        } else {
-            None
-        }
+        let signature = self
+            .routing_node
+            .borrow()
+            .secret_key_share()
+            .map_or(None, |key| Some(key.sign(utils::serialise(&request))));
+        let signature =
+            signature.map(|sig| (self.routing_node.borrow().our_index().unwrap_or(0), sig));
```
ravinderjangra, line 164: Repeated multiple times in the same file. Maybe we can extract this into a function.

src/accumulator.rs
```diff
@@ -0,0 +1,183 @@
...
+    IDataAddress,
+    BTreeSet<XorName>,
+    Vec<(usize, SignatureShare)>,
+);
+
+pub struct Accumulator {
+    routing_node: Rc<RefCell<Node>>,
+    messages: HashMap<MessageId, RequestInfo>,
+    duplications: HashMap<MessageId, DuplicationInfo>,
+    completed: HashSet<MessageId>,
```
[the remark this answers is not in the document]
lionel1704: I was thinking of writing this to disk or something. But this is just a temporary implementation. So didn't want to go that far. 
If you want we can write it to a local DB on disk.

src/accumulator.rs
lionel1704: We need threshold + 1 signatures.

src/data_handler/mod.rs
lionel1704: https://github.com/maidsafe/safe-vault/blob/a3e9458183b3dcd56ae49cca1909c07c728cd767/src/accumulator.rs#L46-L59

src/data_handler/mod.rs
ravinderjangra: but that's only when we (the current node/vault) is one of the targets otherwise we use send_to_peer function where the src & dest both are Node.

PR #1021, opened 2020-07-07
Refactor CI, add scheduled audit, and deploy `latest` to S3 for CI use

S-Coyle:
Updated version of the checkout action being used to v2.
Tidied and condensed the caching.
Added a checkout depth of '0' when adding a tag to the repo to
ensure all commits are retrieved.
Switched the code coverage to it's own job so it can run in
parallel, and set it to only run against PRs.
Added a scheduled security audit.

UPDATE - in order to support CI cross multiple repos where we want to pull the latest released vault I've also added a commit where I've updated the release CI to publish a version of each zip & tar with `latest` alongside the file with a version number. Example of a PR currently in flight which will use this can be viewed [here](https://github.com/maidsafe/safe-client-libs/pull/1224/files#diff-d9db351496552ae42395e3057e47624eR12).
This change will mean we don't need to update multiple CI file vault version numbers manually after each vault release.

### DIFF
--- a/.github/workflows/master.yml
+++ b/.github/workflows/master.yml
@@ -4,7 +4,6 @@
 #
 # 1. Generates a release build.
 # 2. If the last commit is a version change, publish.
-# 3. Gather coverage stats and push to coveralls.io
 
 name: Master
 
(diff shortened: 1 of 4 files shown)


S-Coyle:
Added DO NOT MERGE flag - as discussed with @lionel1704, we would like to not merge this until the tests have been switched over to run with non-mock routing

S-Coyle:
Re-added `--features=mock_parsec` flag to code coverage check to unblock the other improvements offered by this PR - the code coverage can be switched to run against --release in a separate PR.

Also rebased PR

PR #1035, opened 2020-07-27
chore/logs: impl Debug and add info logs

Yoga07:


src/node/mod.rs
```diff
@@ -152,9 +158,19 @@ impl<R: CryptoRng + Rng> Node<R> {
...
+            RunAsAdult(duty) => {
+                info!("Handling Adult duty: {:?}", duty);
+                self.duties.adult_duties()?.process(&duty)
+            }
+            RunAsElder(duty) => {
+                info!("Handling Elder duty: {:?}", duty);
+                self.duties.elder_duties()?.process(duty)
+            }
+            RunAsNode(duty) => {
+                info!("Handling Adult duty: {:?}", duty);
```
oetyng, line 170: Should be "Handling **Node** duty: {:?}"

src/node/mod.rs
```diff
@@ -114,8 +114,14 @@ impl<R: CryptoRng + Rng> Node<R> {
...
             let result = match self.receiver.next_event() {
-                Received::Client(event) => ProcessClientEvent(event).into(),
-                Received::Network(event) => ProcessNetworkEvent(event).into(),
+                Received::Client(event) => {
+                    info!("Received a Client Event from quic-p2p: {:?}", event);
+                    ProcessClientEvent(event).into()
+                }
+                Received::Network(event) => {
+                    info!("Received a Network Event from routing: {:?}", event);
+                    ProcessNetworkEvent(event).into()
```
oetyng, line 123: Don't both of these come from `quic-p2p`?

src/node/node_ops.rs
```diff
@@ -162,10 +176,33 @@ impl Into<NodeOperation> for MessagingDuty {
...
+            Self::SendHandshake { address, .. } => write!(
+                f,
+                "SendHandshake [ address: {:?}, response: (...) ]",
+                address
+            ),
+            Self::SendToAdults { targets, msg } => {
+                write!(f, "SendToAdults [ target: {:?}, msg: {:?} ]", targets, msg)
+            }
+            Self::SendToNode(env) => write!(f, "SendToNode [ msg: {:?} ]", env),
+            Self::SendToSection(env) => write!(f, "SendToSection [ msg: {:?} ]", env),
```
oetyng, line 194: I wondered first if you meant `environment` here :) (Then I understood it is `envelope`.) Maybe better call it `msg`?

src/node/node_ops.rs
```diff
@@ -367,6 +410,8 @@ impl Into<NodeOperation> for PaymentDuty {
 /// metadata only, but could include
 /// chunks, and are then relayed to
 /// Adults (i.e. chunk holders).
+///
```
oetyng, line 413: Surplus comment line.

src/node/mod.rs
Yoga07: Technically yes, but it's internal to routing which handles all Network/Node messages for us. In the case of clients, it's a first-hand direct connection between them and us.

PR #1042, opened 2020-07-28
feat(transfers): store transfers to disk

oetyng:
Uses PickleDb to store transfers to disk.

### DIFF
--- a/src/node/elder_duties/key_section/transfers/replica_manager.rs
+++ b/src/node/elder_duties/key_section/transfers/replica_manager.rs
@@ -201,13 +201,12 @@ impl ReplicaManager {
     }
 
     fn persist(&mut self, event: ReplicaEvent) -> NdResult<()> {
-        match self.store.try_append(event.clone()) {
-            Ok(_) => {
+        self.store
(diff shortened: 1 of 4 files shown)


src/node/elder_duties/key_section/transfers/replica_manager.rs
```diff
@@ -185,10 +200,14 @@ impl ReplicaManager {
         }
     }
 
-    fn persist(&mut self, event: ReplicaEvent) -> Result<()> {
-        self.store.try_append(event.clone())?;
-        self.replica.apply(event);
-        Ok(())
+    fn persist(&mut self, event: ReplicaEvent) -> NdResult<()> {
+        match self.store.try_append(event.clone()) {
```
Yoga07, line 204: Can do a `map` and `map_err` here. Must be careful not to shadow underlying the underlying errors though :)

oetyng: Yeah I failed to remember which to use, maybe you can PR it into this one, and I can approve it? I am not at the desk now to write it myself.

PR #1048, opened 2020-07-31
feat(section_funds): set new actor on elder churn

oetyng:
- Every time the Elder constellation changes, the
key of the section funds change as well. Therefore, we must transfer all money
from previous key to the new key. This is done by
transitioning to a new actor.
- Importantly though, any reward payouts in flight, must be
completed first, requiring the transition to be deferred.
- Also, any reward payouts coming in during transition, or while any is
already in flight, must be enqueued, to be executed once possible again.

### DIFF
--- a/src/node/elder_duties/data_section/rewards/mod.rs
+++ b/src/node/elder_duties/data_section/rewards/mod.rs
@@ -10,7 +10,7 @@ mod farming;
 mod section_funds;
 mod validator;
 
-use self::section_funds::SectionFunds;
+use self::section_funds::{Payout, SectionFunds};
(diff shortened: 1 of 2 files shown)


oetyng:
> LGTM! 👍 But must have this tested with soon, as there are a lot of panics in between the flow(Must handle them later once we set the logic in stone)

Yep, I think I have put that in the comments as well. Where it panics is where we don't know if the node can actually continue when the error occurs, the behavior after that point might be undefined. So that's what we need to find out, how to lead it onto a path of defined behavior, and if it's not possible, well only then should we resort to crashing.

PR #1049, opened 2020-07-31
feat(transfers): keep key subset at replicas on section split

oetyng:
- When section splits, the Replicas in either resulting section
also split the responsibility of the accounts.
Thus, both Replica groups need to drop the accounts that
the other group is now responsible for.

### DIFF
--- a/src/node/elder_duties/data_section/mod.rs
+++ b/src/node/elder_duties/data_section/mod.rs
@@ -84,6 +84,11 @@ impl DataSection {
         self.rewards.transition(actor)
     }
 
+    // At section split, all Elders get their reward payout.
+    pub fn section_split(&mut self) -> Option<NodeOperation> {
+        None
+    }
+
     pub fn relocated_member_joined(
         &mut self,
         old_node_id: XorName,
--- a/src/node/elder_duties/key_section/transfers/replica_manager.rs
+++ b/src/node/elder_duties/key_section/transfers/replica_manager.rs
@@ -15,6 +15,7 @@ use safe_nd::{
     TransferValidated,
 };
 use safe_transfers::{get_genesis, TransferReplica as Replica};
+use std::collections::BTreeSet;
 use threshold_crypto::{PublicKeySet, SecretKeyShare};
 
 use routing::SectionProofChain;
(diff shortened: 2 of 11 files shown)


src/node/elder_duties/data_section/mod.rs
```diff
@@ -84,6 +84,11 @@ impl DataSection {
         self.rewards.transition(actor)
     }
 
+    // At section split, all Elders get their reward payout.
+    pub fn section_split(&mut self) -> Option<NodeOperation> {
+        None
```
Yoga07, line 89: Would be nice to add a TODO/FIXME stating to fix this once we have section splits?

src/node/elder_duties/key_section/transfers/replica_manager.rs
```diff
@@ -82,6 +87,28 @@ impl ReplicaManager {
...
+        self.check_init_status()?;
+
+        // Drops the streams from db.
+        self.store
+            .drop(accounts)
+            .map_err(|e| NdError::NetworkOther(e.to_string()))?;
+
+        // Replays the kept streams
+        // on a new instance of a Replica.
+        self.churn(
```
Yoga07, line 104: Makes sense to create a new one here, but should we necessarily? Won't dropping the non-matching ones be sufficient?

src/node/elder_duties/data_section/mod.rs
oetyng: I see what you mean, but [my upcoming PR has code in this method](https://github.com/oetyng/safe-vault/blob/8585fc1516e76a3a7ea874c5962640497856fccf/src/node/elder_duties/data_section/mod.rs#L90).

src/node/elder_duties/key_section/transfers/replica_manager.rs
oetyng: What we drop is from the database, but the replica in our field is a materialized view from the events in the database, so we need to reload it, or it is essentially like keeping a stale cache.

Yoga07: Right, better to create a new one :)

PR #1063, opened 2020-08-12
feat(rewards): set node reward wallet at startup

oetyng:
- Allows registration of node reward wallet at startup.
- Updates comments for the reward simplification.
- Adds some missing comments.

### DIFF
--- a/src/node/elder_duties/data_section/rewards/mod.rs
+++ b/src/node/elder_duties/data_section/rewards/mod.rs
@@ -34,12 +34,13 @@ pub struct Rewards {
     wrapping: ElderMsgWrapping,
 }
 
+// Node age
 type Age = u8;
 
 pub enum RewardAccount {
     /// When added.
     AwaitingStart(Age),
-    /// After having received the account id, the
+    /// After we have received the account id, the
     /// stage of the RewardAccount is `Active`.
     Active { id: AccountId, age: Age },
     /// After a node leaves the section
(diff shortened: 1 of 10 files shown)


src/node/elder_duties/data_section/rewards/mod.rs
```diff
@@ -34,18 +34,22 @@ pub struct Rewards {
...
+// Node age
 type Age = u8;
 
 pub enum RewardAccount {
-    /// When added.
-    AwaitingStart(Age),
-    /// After having received the account id, the
+    /// When a new node joins the network.
+    NewNode,
+    //Adult(AccountId),
```
Yoga07, line 43: Can remove this.

src/node/elder_duties/data_section/rewards/mod.rs
```diff
@@ -141,16 +155,40 @@ impl Rewards {
...
+        let _ = self.node_accounts.insert(node_id, RewardAccount::NewNode);
         None
     }
 
-    /// 1. When a node is relocated to our section, we add the account
-    /// and send a cmd to old section, for claiming the rewards.
-    fn add_relocated_account(
+    /// 1. A new node registers an account id for future reward payout.
+    /// ... or, an active node updates its account.
+    fn set_node_account(&mut self, node_id: XorName, id: AccountId) -> Option<MessagingDuty> {
```
Yoga07, line 169: Better to make the return type a boolean? We don't seem to returning anything except `None`

oetyng: This is the same all over, when there is no more operations to do, it is None returned. I think this was the same previously as well.

A bool would not fit IMO, because we have no handling of bools at upper layers. There are ops or no ops.

Yoga07: Hmm, yea, would make better sense to get things follow the same pattern, cools :+1:

PR #1071, opened 2020-08-20
refactor(node): adapt code to make use of routing async API

oetyng:
Rebasing `bochaco/async-routing-api` onto `maidsafe/master`.

Supersedes #1068.


oetyng:
Superseded by #1072.

PR #1077, opened 2020-08-24
Remove excessive cloning

lionel1704:
One change that I still have doubts about is to the Reading and Writing structs in adult duties.
A struct itself might not be required there, so perhaps we can simplify it to be `reading::do_something(....)` and `writing::do_something(...)`. What do you think?


joshuef (approved):

joshuef:
> so perhaps we can simplify it to be reading::do_something(....) and writing::do_something(...). What do you think?

I'm not sure what you're proposing here. 

The structs seem pretty clear to me. What do we gain by dropping them? (Also perhaps a good convo to have on another channel . another PR w/ the proposal there?)

PR #1082, opened 2020-08-25
Remove Reading and Writing structs and remove cloning of MsgEnvelope

lionel1704:


src/node/adult_duties/chunks/reading.rs
```diff
@@ -9,55 +9,38 @@
...
             None
         }
-    }
-
-    fn verify_msg(&self) -> bool {
-        match self.msg.most_recent_sender() {
-            MsgSender::Section { .. } => self.msg.verify(),
-            _ => false,
-        }
+    // } else if matches!(self.requester, PublicId::Node(_)) {
```
dirvine, line 29: @lionel1704  did you mean to leave this commented block?

src/node/adult_duties/chunks/writing.rs
```diff
@@ -6,53 +6,36 @@
...
         }
-    }
-
-    fn verify_msg(&self) -> bool {
-        match self.msg.most_recent_sender() {
-            MsgSender::Section { .. } => self.msg.verify(),
-            _ => false,
+        DeletePrivate(address) => {
+            if msg.verify() {
+                // really though, for a delete, what we should be looking at is the origin signature! That would be the source of truth!
```
dirvine, line 33: This is worth "fixing"

src/node/adult_duties/chunks/reading.rs
lionel1704: Yes @dirvine 
This code is a part of the data duplication process when a node leaves the network. That part has been temporarily disabled in the refactor. So we'll need to enable this later.

src/node/adult_duties/chunks/writing.rs
lionel1704: Yes. But the change was not a part of this PR. So I believe it should be in a PR of it's own with other relevant changes?

Yoga07 (approved):

PR #1090, opened 2020-09-07
Async statedb file io

### LINKED ISSUE #1089: Make File I/O async in vault state_db
**Is your feature request related to a problem? Please describe.**
File i/o is currently synchronous but could be made asynchronous in the state_db.rs file handling.

Add the "fs" flag to the `tokio` dependency and make the `store_` and `get_` functions async.

**Additional context**
See forum discussion [here](https://forum.safedev.org/t/still-using-synchronous-file-i-o-in-vaults/2881).

Scorch-Dev:
fixes #1089 
~~This should compile, but I'm not entirely sure as the dependencies are being reworked so I haven't been able to test it locally quite yet. As such, the `Cargo.lock` file will probably need an extra update later. I'll leave this here as a draft."~~

### DIFF
--- a/src/node/mod.rs
+++ b/src/node/mod.rs
@@ -55,15 +55,15 @@ impl<R: CryptoRng + Rng> Node<R> {
             None => {
                 let secret = SecretKey::random();
                 let public = secret.public_key();
-                store_new_reward_keypair(root_dir, &secret, &public)?;
+                store_new_reward_keypair(root_dir, &secret, &public).await?;
                 PublicKey::Bls(public)
             }
         };
-        let age_group = if let Some(age_group) = get_age_group(&root_dir)? {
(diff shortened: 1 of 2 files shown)


Scorch-Dev:
Actually I found a couple other places where std::fs was being used, so I may go ahead and patch those while I'm at it. Will expand the scope of the commit a bit though...

dirvine:
Excellent @Scorch-Dev this is all very helpful and we are stabalising as fast as possible. Some async work in routing but almost there now. Thank you again though

joshuef (approved):

joshuef:
Ah, this needs a rebase, @Scorch-Dev . Looks good though. So if you can update I'll get it merged in, thanks!

joshuef (approved):

PR #1103, opened 2020-09-17
Make the vault fully async and fix the Vault Test suite to work with async/await.

lionel1704:


Cargo.toml
```diff
@@ -58,10 +60,10 @@ xor_name = "1.1.0"
 [dev_dependencies]
 maplit = "1.0.1"
 tempdir = "~0.3.7"
-safe_core = { git = "https://github.com/maidsafe/safe_client_libs", branch = "master" }
-tokio = { version="~0.2.21", features=["rt-core", "blocking", "stream", "rt-util"] }
+#safe_core = { git = "https://github.com/maidsafe/safe_client_libs", branch = "master" }
```
joshuef, line 63:
```suggestion
safe_core = { git = "https://github.com/maidsafe/safe_client_libs", branch = "master" }
```

Cargo.toml
```diff
@@ -58,10 +60,10 @@ xor_name = "1.1.0"
...
+safe_core = { path = "../safe_client_libs" }
```
joshuef, line 64:
```suggestion
#safe_core = { path = "../safe_client_libs" }
```

joshuef (changes requested):

Yoga07 (approved):

joshuef (changes requested):
Would be good for this to work with network launcher tool before merging.

We really need to stick to one method of bootstrapping a network, and I view that as our goto. If we need changes from any other script we should be incorporating them there

src/chunk_store/tests.rs
```diff
@@ -88,22 +89,22 @@ fn successful_put() {
...
         root.path(),
         u64::MAX,
-        Rc::clone(&used_space),
+        Arc::clone(&used_space),
         Init::New
     ));
 
     let mut put = |data: &Data, size| {
-        let used_space_before = used_space.get();
+        let used_space_before = used_space.lock().unwrap();
```
joshuef, line 101: @lionel1704 I don't think we should be adding in all these `unwraps` given we've been trying to remove them recently.

Better to be using `futures::lock::Mutex` and then this will look like `used_space.lock().await`, no `unwrap()` needed 👍

lionel1704:
Superseeded by #1106

Issue #1094, opened 2020-09-09
Make Chunk Storage File I/O Async


Scorch-Dev:
**Is your feature request related to a problem? Please describe.**
On the heels of #1089 it seems there's room for improvement by using the asynchronous file i/o API for chunk storage as well. Compared to the other issue, this is a bit more of a sweeping change, so I made a separate issue for this.

**Describe the solution you'd like**
Use `tokio;:fs` to access files instead of `std::fs`

**Describe alternatives you've considered**
Leaving file access as `std::fs` for the moment and fixing this later is a possibility to increase stability I suppose.

**Additional context** 
None

Scorch-Dev:
Just as an update, the original PR got really muddied and was hard to read, so I'm splitting it into two parts. The first part is to make the tracking of used space safe to use asynchronously. 

The second part is to implement `tokio::fs` storage of the actual chunks and make that access async-safe. The first part is pretty much implemented now (on a [separate branch](https://github.com/Scorch-Dev/safe-vault/tree/async-used-space), not the one linked to the closed PR draft), and I'll pull in the changes from the second part once I know that the async used space tracking works (e.g. when safe-vault tests become stable again).

PR #1105, opened 2020-09-19
enhancement(UsedSpace): Synchronise Individual ChunkStore UsedSpace Access

Scorch-Dev:
Simplifies the passing of `used_space` to chunk stores. It creates a shared `UsedSpace` instance which is a thin wrapper to an `Arc<Mutex<UsedSpace>>` and synchronizes access to the global amount of used space (`total_value` and `max_capacity` in code today) and the individual used space for each chunk store. 

**Quick Benefits**
* Paves the way for concurrency within a single `ChunkStore`, not just across `ChunkStore`s (see "Motivation" section below)
* Less implementation bleeding to other modules
    * e.g. no more passing around `Arc<Mutex<u64>> used_space` everywhere. Instead just becomes `used_space.clone()`
* reduced redundancy of shared state and less prone to related bugs
    * e.g. no more need to pass `maximum_capacity` to `ChunkStore` objects, giving each there own copy, which could potentially fall out of sync if it were ever needed to be changed

## Motivation

This doesn't seem to be covered in the big [asysnc PR](https://github.com/maidsafe/safe-vault/pull/1103). This PR seems to cover mostly things related to qp2p and synchronizes access to the global `used_space`, but  not the local used space tracking for any given `ChunkStore<T>`. Currently, the struct `ChunkStore<TChunk>` is not async safe. So while many of their functions may be denoted `async` after that linked PR is merged, the vault doesn't currently have the flexibility to leverage concurrency for multiple vault operations. 

Thus, statements like the following, or potential architectural changes of similar nature would not be possible, limiting the vault's potential for performance:

    // Currently, not possible (even after PR merges)
    let tasks = Vec::new();
    for _ in 0..10 {
        tasks.push( chunk_store.do_concurrent_operation() );
    }
    future::futures::try_join_all(&tasks).await?

Instead, we would need something like the following, code, which would `await` each operation, which could otherwise be run concurrently, before running the next.

    // Only this is possible right now (even after PR merges)
    for _ in 0..10 {
        chunk_store.do_operation().await;
   }

So the current architecture allows concurrency for connections, it's still limiting for chunk store operations.

The first step to enabling chunk store operations to run concurrently, is to simplify the interface to `UsedSpace`, so the `local_value` of the `UsedSpace` struct that each `ChunkStore<T>` holds, as well as the file `used_space` for each chunk store, tracks properly across multiple operations on potentially separate threads.

## Related

I mentioned this briefly in a couple of places, and it's currently not a part of the [`async` PR](https://github.com/maidsafe/safe-vault/pull/1103) as I mentioned above. I did mention it [here](https://forum.safedev.org/t/still-using-synchronous-file-i-o-in-vaults/2881) in the forums and in a git issue I raised earlier, #1094 and alluded to in and PR #1090. This PR is part of #1094  but doesn't fully close it per say, but it would be the first step.

### DIFF
--- a/src/chunk_store/tests.rs
+++ b/src/chunk_store/tests.rs
@@ -9,13 +9,11 @@
 use super::{
     chunk::{Chunk, ChunkId},
     error::Error,
-    ChunkStore, Subdir,
+    ChunkStore, Subdir, UsedSpace,
 };
 use crate::{utils::Init, Result, ToDbKey};
-use futures::lock::Mutex;
 use rand::{distributions::Standard, rngs::ThreadRng, Rng};
 use serde::{Deserialize, Serialize};
-use std::sync::Arc;
 use std::{path::Path, u64};
 use tempdir::TempDir;
 use unwrap::unwrap;
@@ -90,13 +88,9 @@ async fn successful_put() -> Result<()> {
     let chunks = Chunks::gen(&mut rng);
 
     let root = temp_dir();
-    let used_space = Arc::new(Mutex::new(0));
-    let mut chunk_store = unwrap!(ChunkStore::<Data>::new(
-        root.path(),
-        u64::MAX,
-        Arc::clone(&used_space),
-        Init::New
-    ));
+    let used_space = UsedSpace::new(u64::MAX);
+    let mut chunk_store =
+        unwrap!(ChunkStore::<Data>::new(root.path(), used_space.clone(), Init::New).await);
 
     for (index, (data, size)) in chunks.data_and_sizes.iter().enumerate().rev() {
         let the_data = &Data {
@@ -131,13 +125,9 @@ async fn failed_put_when_not_enough_space() -> Result<()> {
     let mut rng = new_rng();
     let root = temp_dir();
     let capacity = 32;
-    let used_space = Arc::new(Mutex::new(0));
-    let mut chunk_store = unwrap!(ChunkStore::new(
-        root.path(),
-        capacity,
-        used_space,
-        Init::New
-    ));
+    let used_space = UsedSpace::new(capacity);
+    let mut chunk_store =
+        unwrap!(ChunkStore::new(root.path(), used_space.clone(), Init::New).await);
 
     let data = Data {
         id: Id(rng.gen()),
@@ -161,8 +151,8 @@ async fn delete() -> Result<()> {
     let chunks = Chunks::gen(&mut rng);
 
     let root = temp_dir();
-    let used_space = Arc::new(Mutex::new(0));
-    let mut chunk_store = ChunkStore::new(root.path(), u64::MAX, used_space, Init::New)?;
+    let used_space = UsedSpace::new(u64::MAX);
+    let mut chunk_store = ChunkStore::new(root.path(), used_space.clone(), Init::New).await?;
 
     for (index, (data, size)) in chunks.data_and_sizes.iter().enumerate() {
         let the_data = &Data {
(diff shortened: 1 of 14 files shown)


joshuef:
@Scorch-Dev thanks for this.

And just so you're aware. Currently digging in to CI e2e failures. Once we have that in / straightened out we'll be able to verify this against the CI better (the ubuntu fail is almost certainly the same).

Will get to this asap after that 👍

Scorch-Dev:
@joshuef No worries, and thanks for the heads up 👍

src/chunk_store/tests.rs
```diff
@@ -90,13 +88,9 @@ async fn successful_put() -> Result<()> {
...
-    let used_space = Arc::new(Mutex::new(0));
-    let mut chunk_store = unwrap!(ChunkStore::<Data>::new(
-        root.path(),
-        u64::MAX,
-        Arc::clone(&used_space),
-        Init::New
-    ));
+    let used_space = UsedSpace::new(u64::MAX);
+    let mut chunk_store =
+        unwrap!(ChunkStore::<Data>::new(root.path(), used_space.clone(), Init::New).await);
```
joshuef, line 93: @Scorch-Dev we're currently striving to cleanup tests, to return a `Result` type. Would be great to use that here instead of `unwrap`.
```suggestion
        ChunkStore::<Data>::new(root.path(), used_space.clone(), Init::New).await?;
```

src/chunk_store/tests.rs
```diff
@@ -131,13 +125,9 @@ async fn failed_put_when_not_enough_space() -> Result<()> {
...
-    let used_space = Arc::new(Mutex::new(0));
-    let mut chunk_store = unwrap!(ChunkStore::new(
-        root.path(),
-        capacity,
-        used_space,
-        Init::New
-    ));
+    let used_space = UsedSpace::new(capacity);
+    let mut chunk_store =
+        unwrap!(ChunkStore::new(root.path(), used_space.clone(), Init::New).await);
```
joshuef, line 130:
```suggestion
        ChunkStore::new(root.path(), used_space.clone(), Init::New).await?;
```

src/chunk_store/tests.rs
```diff
@@ -186,13 +176,9 @@ async fn put_and_get_value_should_be_same() -> Result<()> {
...
-    let used_space = Arc::new(Mutex::new(0));
-    let mut chunk_store = unwrap!(ChunkStore::new(
-        root.path(),
-        u64::MAX,
-        Arc::clone(&used_space),
-        Init::New
-    ));
+    let used_space = UsedSpace::new(u64::MAX);
+    let mut chunk_store =
+        unwrap!(ChunkStore::new(root.path(), used_space.clone(), Init::New).await);
```
joshuef, line 181:
```suggestion
        ChunkStore::new(root.path(), used_space.clone(), Init::New).await?;
```

src/chunk_store/tests.rs
```diff
@@ -239,13 +224,9 @@ async fn overwrite_value() -> Result<()> {
...
-    let used_space = Arc::new(Mutex::new(0));
-    let chunk_store: ChunkStore<Data> = unwrap!(ChunkStore::new(
-        root.path(),
-        u64::MAX,
-        used_space,
-        Init::New
-    ));
+    let used_space = UsedSpace::new(u64::MAX);
+    let chunk_store: ChunkStore<Data> =
+        unwrap!(ChunkStore::new(root.path(), used_space.clone(), Init::New).await);
```
joshuef, line 229:
```suggestion
        ChunkStore::new(root.path(), used_space.clone(), Init::New).await?;
```

src/chunk_store/used_space.rs
```diff
@@ -9,87 +9,346 @@
...
+    }
 }
 
 impl UsedSpace {
-    pub fn new<T: AsRef<Path>>(
+    /// construct a new used space instance
+    /// NOTE: this constructs a new async-safe instance,
+    /// If you intend to create a new local `ChunkStore` tracking,
+    /// then use `clone()` and `add_local_store()` to ensure
+    /// consistentcy across local `ChunkStore`s
```
joshuef, line 41:
```suggestion
    /// consistency across local `ChunkStore`s
```

joshuef (changes requested):
Overall great PR with a lot docs and explanations. Thanks @Scorch-Dev  

Few wee things to tweak to stay consistent with how we're looking to approach tests. I've added some suggestions for those 👍 .

lionel1704 (commented):
Nice one @Scorch-Dev. Great work. 
Once @joshuef's comments are addressed, we can get this in.

lionel1704 (approved):

PR #1106, opened 2020-09-20
Async test suite atop master

joshuef:
Supersedes #1103


joshuef:
Only thing left on this is _I think_ to convert the Mutex to use futures mutex and remove all the unwraps there.

Oh, and a clippy fix I was not sure about: 
```
error: unsequenced read of a variable
   --> src/node/elder_duties/mod.rs:110:13
    |
110 |             self.key_section.elders_changed().await,
    |             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
```

lionel1704 (approved):

PR #1109, opened 2020-09-21
fix(process_while_any): don't drop any from `ops`

oetyng:
- It seems that previous impl simply drops remaining items in `ops` as soon
as any of them result in subsequent operations.

### DIFF
--- a/src/node/mod.rs
+++ b/src/node/mod.rs
@@ -156,14 +156,13 @@ impl<R: CryptoRng + Rng> Node<R> {
             next_op = match op {
                 Single(operation) => self.process(operation).await,
                 Multiple(ops) => {
-                    let mut node_op = None;
+                    let mut node_ops = Vec::new();
                     for c in ops.into_iter() {
                         if let Some(op) = self.process(c).await {
-                            node_op = op;
-                            break;
+                            node_ops.push(op);
                         }
                     }
-                    Some(node_op)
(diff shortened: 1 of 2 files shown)


src/node/mod.rs
```diff
@@ -155,16 +155,12 @@ impl<R: CryptoRng + Rng> Node<R> {
...
-                        }
-                    }
-                    Some(node_op)
-                }
+                Multiple(ops) => Some(
+                    ops.into_iter()
+                        .filter_map(|c| futures::executor::block_on(self.process(c)))
+                        .collect::<Vec<_>>()
+                        .into(),
+                ),
```
lionel1704, line 163: We are avoiding the use of `futures::executor::block_on()`. 
Does this make sense?
```suggestion
                    let mut node_ops = Vec::new();
                    for c in ops.into_iter() {
                        if let Some(op) = self.process(c).await {
                            node_ops.push(op);
                        }
                    }
                    Some(node_ops)
```

oetyng: Yeah sure, absolutely. I agree blocking on the tasks is not desired, was just looking for a simple fix.

oetyng: Woops, tried to use the commit suggestion feature. Will force push.

lionel1704 (approved):

PR #1124, opened 2020-09-28
Ci add chaos

joshuef:
Add `with_chaos` macro to randomly perform actions (such as drop messages), to test for resilience.

Provides `SAFE_CHAOS_LEVEL` env var to configure the likelihood of chaos occuring. Where you can provide a value 1-100 to add more chaos.

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -120,7 +120,7 @@ jobs:
         uses: actions-rs/cargo@v1
         with:
           command: test
-          args: --release  --features=simulated-payouts
+          args: --release  --features=simulated-payouts,chaos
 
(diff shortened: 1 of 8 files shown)


S-Coyle (approved):

Yoga07 (approved):

joshuef:
Looks like our client or vaults are not stable enough for this yet. Going to dig in to what the issue is here

joshuef:
Hmm, i'm updating the client with some retries, though atm with the current section startup, we're not getting more than 3 elders, which makes it trickier. This may have to wait until the section startup is more solid.

S-Coyle (approved):

Yoga07 (approved):

PR #1153, opened 2020-12-02
Monitior Section storage and use joins_allowed switch

Yoga07:


src/network.rs
```diff
@@ -117,6 +117,15 @@ impl Network {
         // Ok(())
     }
 
+    pub async fn set_node_join(&mut self, joins_allowed: bool) -> Result<()> {
```
lionel1704, line 120: Perhaps a rename will do good here?

src/node/elder_duties/key_section/mod.rs
```diff
@@ -86,6 +94,13 @@ impl<R: CryptoRng + Rng> KeySection<R> {
         self.transfers.catchup_with_replicas().await
     }
 
+    pub async fn set_node_join_flag(&mut self, joins_allowed: bool) -> Outcome<NodeOperation> {
```
lionel1704, line 97: Here too. Perhaps rename this to `set_joins_allowed` ?

lionel1704:
Left a few small comments. Once those are addressed and this is rebased on top of the Dev branch we can merge this.

Yoga07:
Closing in favour of https://github.com/maidsafe/sn_node/pull/1172

PR #1158, opened 2020-12-07
Re-enable chunk duplication only at adults

Yoga07:


src/node/adult_duties/chunks/chunk_storage.rs
```diff
@@ -127,6 +127,34 @@ impl ChunkStorage {
...
+
+        if let Err(e) = self
+            .chunks
+            .put(&blob)
+            .await
+            .map_err(|error| error.to_string().into())
+        {
+            Outcome::error(e)
+        } else {
+            Outcome::oki_no_change()
```
joshuef, line 154: is this `no_change`? or just `ok`? There's a diff in 143, and 154 outcomes. The latter has written something. Might be worth reflecting that?

src/node/adult_duties/mod.rs
```diff
@@ -9,31 +9,95 @@
...
+        let result = match duty {
+            RunAsChunks(chunk_duty) => match chunk_duty {
+                ReadChunk(msg) | WriteChunk(msg) => self.chunks.receive_msg(msg).await,
+            },
+            RequestForChunk {
+                section_authority,
+                address,
+                targets,
+            } => {
+                info!("Creating new MsgEnvelope for getting duplicate Chunk at NewHolder");
```
joshuef, line 53: Getting it at newHolder?

I'm confused by this message.

src/node/adult_duties/mod.rs
```diff
@@ -9,31 +9,95 @@
...
+                let origin = MsgSender::adult(pk, ChunkStorage, sign)?;
+                let env = MsgEnvelope {
+                    message: msg,
+                    origin,
+                    proxies: vec![],
+                };
+                info!("Sending to existing Holders");
+                self.msg_wrapping.send_to_adults(&env, targets).await
+            }
+            ReplyForDuplication {
```
joshuef, line 73: Should this duty perhaps not be, `ReplyWithChunk` or some such?

(also: I am doubting the use of `chunk` so liberally here... it's correct. but i'm worried about confusion with `Blob` etc... I know this is node duties, but how does this compare against the data fetch paths eg... `GetChunk` seems like it should be `Blob` related, but it's actually duplication related eg... I'm not sure how to solve this (if it's needed at all), but raising in case it chimes with anyone, or they have ideas to sort it.

src/node/adult_duties/mod.rs
```diff
@@ -9,31 +9,95 @@
...
+                            }),
+                            id: MessageId::new(),
+                        };
+                        info!("Send blob for duplication to the NewHolder");
+                        self.msg_wrapping.send_to_node(msg).await
+                    }
+                    Err(e) => {
+                        // TODO: Penalise adult for this behaviour
+                        error!("Chunk doesn't exist at current holder for Duplication. But it should. Check Logic");
+                        Outcome::error(e)
```
joshuef, line 96: Do we handle this error? (I didn't see where). Should we?

src/node/node_duties/msg_analysis.rs
```diff
@@ -191,35 +193,43 @@ impl NetworkMsgAnalysis {
...
+        let is_chunk_msg = matches!(msg.message,
+        Message::Cmd {
+            cmd:
+                Cmd::Data {
+                    cmd: DataCmd::Blob(_),
+                    ..
+                },
+            ..
+        }
+        | Message::Query { // TODO: Should not accumulate queries, just pass them through.
```
joshuef, line 217: Does this TODO need handled ? Or it's okay to leave it for now?

src/node/node_duties/msg_analysis.rs
```diff
@@ -318,18 +328,109 @@ impl NetworkMsgAnalysis {
...
 
         let shall_process =
-            from_metadata_section() && self.is_dst_for(msg).await? && self.is_adult().await;
+            from_metadata_section() && self.is_dst_for(&msg).await? && self.is_adult().await;
 
         if !shall_process {
             return Ok(None);
         }
 
+        let dup = match &msg.message {
```
joshuef, line 337: `dup` ? Is there a better/clearer name for this?

src/node/adult_duties/chunks/chunk_storage.rs
Yoga07: It should be just `no_change` at L#154 as the blob is successfully written to db and the flow ends there. We don't notify that to anybody else as it is only a chunk duplication write.

src/node/adult_duties/mod.rs
Yoga07: We don't handle this error because the requester sends the `GetChunk` request to multiple current_holders whom are identified by the section to hold those chunks. We only need one of the many to reply w/this chunk as the worst-case. 

But this must be handled later on where we keep track of these duplication msgs and if such adults don't have the chunk(which the section thinks it has), then they'll have to be penalised for it.

src/node/node_duties/msg_analysis.rs
Yoga07: This is an older TODO and can be handled later.

src/node/adult_duties/mod.rs
Yoga07: I guess `ReplyForDuplication` fits here as it's an internal node cmd where the node tells itself to reply for the duplication request. `GiveChunk` is the inter-node message that might go well with `ReplyWithChunk` I guess?

src/node/adult_duties/chunks/chunk_storage.rs
joshuef: I'm trying to understand why `no_change` and not just `ok`? What's the difference here? There is a change in that it's been written no? Or does `ok` always involve returning something?

src/node/node_duties/msg_analysis.rs
```diff
@@ -327,14 +329,16 @@ impl NetworkMsgAnalysis {
...
+            && self.is_dst_for(&msg).await?
+            && self.is_adult().await;
 
         if !shall_process {
             return Ok(None);
         }
 
-        let dup = match &msg.message {
+        info!("Checking chunking duplication!");
+        let is_chunk_duplication = match &msg.message {
```
joshuef, line 340: `is_chunk_duplication` ->  `is_chunk_duplicated` assuming that's what we mean

Yoga07: Not exactly, we are checking if the incoming message is a chunk duplication message. A generic one where it covers all the variants(`DuplicateChunk`, `GetChunk`, `GiveChunk` etc.)

src/node/adult_duties/chunks/chunk_storage.rs
Yoga07: `ok` only `Option<T>`-ises a `Result<T,E>`, where `oki_no_change` from our `TernaryResult` returns `Ok(None)` to fit the return `Outcome<T>`.

joshuef (approved):

PR #1164, opened 2020-12-14
Initiate section actor

oetyng:
Solves the initiation of genesis section actor, with a small tweak to `sn_routing`.


joshuef (approved):

src/node/elder_duties/mod.rs
```diff
@@ -47,11 +47,14 @@ impl ElderDuties {
...
     /// start working properly in the group.
     pub async fn initiate(&mut self, first: bool) -> Result<NodeOperation> {
-        // currently only key section needs to catch up
+        let mut ops = vec![];
         if first {
-            let _ = self.key_section.init_first().await?;
+            ops.push(self.key_section.init_first().await?);
         }
-        self.key_section.catchup_with_section().await
+        ops.push(self.key_section.catchup_with_section().await?);
```
Yoga07, line 54: Better to put this in the `else` block because the geneis would also run `catch_up_with_section` when there is technically no section? If we want to update a value at genesis, we should do so by updating it directly, as this'd it induces traffic.

src/node/elder_duties/data_section/rewards/mod.rs
```diff
@@ -15,23 +15,24 @@ pub use self::{reward_calc::RewardCalc, validator::Validator};
 use crate::{
     node::keys::NodeSigningKeys,
     node::msg_wrapping::ElderMsgWrapping,
-    node::node_ops::{NodeMessagingDuty, NodeOperation, RewardDuty},
+    node::node_ops::{Blah, NodeMessagingDuty, NodeOperation, RewardCmd, RewardDuty, RewardQuery},
```
Yoga07, line 18: Can rename the `Blah` trait :)

Yoga07 (commented):
https://github.com/maidsafe/sn_node/blob/194a9a317b0ed0880ba74f136a3e3898db7a949c/src/node/elder_duties/key_section/transfers/replicas.rs#L76
Not a part of this PR, but this line is a bit confusing to me(and at other querying methods). We are loading the db everytime we get a new query. Why can we not hold the db instance and read from there?

src/node/elder_duties/data_section/rewards/mod.rs
joshuef: Ha, good catch

joshuef:
I'm not sure how these tests are passing here. Locally running my client against this, I see a whole heap of errors. 

(I had expected as much w/ PR, that shouldn't really be a blocker for this into Dev... but are they passing? 🤔 )

src/node/elder_duties/mod.rs
oetyng: It's generically just catching up, using same path regardless and it works fine. There is no traffic if we are genesis, so that special case would just be bloat code.

oetyng:
> https://github.com/maidsafe/sn_node/blob/194a9a317b0ed0880ba74f136a3e3898db7a949c/src/node/elder_duties/key_section/transfers/replicas.rs#L76
> 
> 
> Not a part of this PR, but this line is a bit confusing to me(and at other querying methods). We are loading the db everytime we get a new query. Why can we not hold the db instance and read from there?

We hold it for write operations, which is then locked. But we don't want to lock on read so we just bypass the lock, by reading directly from disk (where it all is anyway since the auto dump db is used).

There's a readonly instance available of pickledb, but since there's a concurrency management one layer above that's not a help,, as we just don't want to serialize read access.

src/node/elder_duties/data_section/rewards/mod.rs
oetyng: Yep, feel free to figure something out. Will address it eventually otherwise :+1:

oetyng:
> I'm not sure how these tests are passing here. Locally running my client against this, I see a whole heap of errors.
> 
> (I had expected as much w/ PR, that shouldn't really be a blocker for this into Dev... but are they passing? 🤔 )

Yeah, let's look into that. I could not run the tests properly because the log file grew so big that very soon all they did was waiting for the IO to the file.
But let's try it without logging and see.

joshuef:
Hmm, i dont have an issue w/ log file size myself (on ubuntu... though they do definitely grow , that's true)

It looks like only 15 tests are running w/ `cargo test --release --features=simulated-payouts`, and I'm not sure why / whats changed there

PR #1172, opened 2020-12-16
Monitior Section storage and use joins_allowed switch

Yoga07:


src/node/mod.rs
```diff
@@ -92,14 +92,32 @@ impl Node {
...
+        } else {
+            info!("We are Infant");
+            Infant
+        };
+
+        let mut duties = NodeDuties::new(node_info, network_api.clone()).await;
         let _ = match age_group {
-            Infant => Ok(NodeOperation::NoOp),
+            Infant => {
+                info!("Do nothing as Infant");
```
joshuef, line 111:
```suggestion
                info!("Becoming Infant, do nothing.");
```

src/node/node_duties/messaging/network_sender.rs
```diff
@@ -111,6 +111,7 @@ impl NetworkSender {
             Address::Node(xorname) => DstLocation::Node(xorname),
             Address::Client(xorname) | Address::Section(xorname) => DstLocation::Section(xorname),
         };
+        info!("Dst Location: {:?}", dst);
```
joshuef, line 114:
```suggestion
        info!("Destination: {:?}", dst);
```

src/node/node_duties/mod.rs
```diff
@@ -108,9 +108,29 @@ impl NodeDuties {
                     .await
             }
             NoOp => Ok(NodeOperation::NoOp),
+            StorageFull => self.notify_section_of_our_storage().await,
         }
     }
 
+    async fn notify_section_of_our_storage(&mut self) -> Result<NodeOperation> {
```
joshuef, line 115: If this is never used but to notify full storage, then I'd rename this to reflect that

joshuef (changes requested):
minor tweaks on some log messages and naming

src/node/elder_duties/key_section/payment/mod.rs
```diff
@@ -0,0 +1,148 @@
+// Copyright 2020 MaidSafe.net limited.
```
oetyng, line 1: This file can be removed, was merged with transfers.

src/node/elder_duties/key_section/transfers/replica_manager.rs
```diff
@@ -398,4 +398,4 @@
 //     pub fn debit_without_proof(&mut self, transfer: Transfer) -> NdResult<()> {
```
oetyng, line 398: This file can be removed as well.

src/node/node_duties/mod.rs
oetyng: I'd say, additionally, that this is an event, not a cmd. `StorageIsFull` is a notification about a fact. It's not a cmd to do something.
But that can be fixed in a follow up PR.

PR #1194, opened 2020-12-27
Improved logging

Thierry61:
Improved logging a bit:
- Display number of elders and adults also when a node has left (not just when a node has joined)
- Adapted length of surrounding lines to length of content. The result is like this:
```plain
[sn_node] INFO 2020-12-27T19:29:50.992580556+01:00 [src/node/node_duties/network_events.rs:46] -----------------------------------
[sn_node] INFO 2020-12-27T19:29:50.992648755+01:00 [src/node/node_duties/network_events.rs:47] | No. of Elders in our Section: 5 |
[sn_node] INFO 2020-12-27T19:29:50.992662555+01:00 [src/node/node_duties/network_events.rs:48] | No. of Adults in our Section: 4 |
[sn_node] INFO 2020-12-27T19:29:50.992678455+01:00 [src/node/node_duties/network_events.rs:49] -----------------------------------
```
(separators length depends on the max length of inner messages)

The only thing that is missing is to get this information as an adult. That would be great if Maidsafe allowed this. Not all users will be an elder, they might be disappointed to not have this information.

Other improvements (not related to logging):
- Renamed all "remaining space ratio" functions and messages to "used space ratio" which is the value that is really computed and displayed.
- Added a line in all Github Action jobs to execute them only when they are invoked from Maidsafe repository.  The aim is to avoid spurious errors and mails sent to people forking the repo. This is the same proposal as for qp2p [#189](https://github.com/maidsafe/qp2p/pull/189). By the way I received no news for this one.
- Corrected indentation in tag release workflow.
- Set version number in Cargo.lock in phase with Cargo.toml

### DIFF
--- a/.github/workflows/auto_merge_prs.yml
+++ b/.github/workflows/auto_merge_prs.yml
@@ -10,6 +10,7 @@ on: pull_request
 
 jobs:
   merge:
+    if: ${{ github.repository_owner == 'maidsafe' }}
     runs-on: ubuntu-20.04
     steps:
     - uses: actions/checkout@v2
--- a/.github/workflows/bump_version.yml
+++ b/.github/workflows/bump_version.yml
@@ -11,6 +11,7 @@ env:
 
 jobs:
   update_changelog:
+    if: ${{ github.repository_owner == 'maidsafe' }}
     runs-on: ubuntu-20.04
     # Dont run if we're on a release commit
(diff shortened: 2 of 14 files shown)


dirvine (approved):
LGTM, thanks so much @Thierry61  very nice work.

PR #1197, opened 2021-01-05
Simplify errors and update deps

joshuef:


lionel1704 (approved):
The changes look good. But I have a concern about the removal of the challenge. Will discuss on Slack.

PR #1198, opened 2021-01-07
feat: Remove stream handling

joshuef:
- removes storage of client stream in nodes
- uses qp2p connection pool to send messaages to clients via routing
- note: still uses `stream` for sending bootstrap response

### DIFF
--- a/src/node/elder_duties/key_section/client/client_msg_handling.rs
+++ b/src/node/elder_duties/key_section/client/client_msg_handling.rs
@@ -147,14 +147,8 @@ impl ClientMsgHandling {
 
         match self.tracked_incoming.remove(&correlation_id) {
             Some((_, client_address)) => {
-                trace!("&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&");
-
                 let bytes = utils::serialise(message)?;
 
-                trace!("&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&");
-                trace!("&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&");
-                trace!("&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&");
-                trace!("&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&");
                 trace!("will send message via qp2p");
                 self.onboarding.send_bytes_to(client_address, bytes).await
             }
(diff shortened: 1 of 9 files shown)


src/node/elder_duties/key_section/client/client_msg_handling.rs
```diff
@@ -143,99 +105,49 @@ impl ClientMsgHandling {
...
+            | Message::CmdError { correlation_id, .. }
+            | Message::QueryResponse { correlation_id, .. } => correlation_id,
             _ => {
                 error!(
                     "{} for message-id {:?}, Invalid message for client.",
-                    self, msg_id
+                    self,
+                    message.id()
                 );
                 return Err(Error::InvalidMessage);
```
bochaco, line 123: Perhaps we can make the msg id, msg type received and msg type expected args of the Error type so we can provide detailed information when handling the error?

src/node/elder_duties/key_section/client/mod.rs
```diff
@@ -79,37 +76,31 @@ impl ClientGateway {
...
-                if let Some(public_key) = existing_client {
-                    let msg = try_deserialize_msg(&content)?;
-                    info!("Deserialized client msg from {}", public_key);
-                    trace!("Deserialized client msg is {:?}", msg.message);
-                    if !validate_client_sig(&msg) {
-                        return Err(Error::NetworkData(DtError::InvalidSignature));
+            RoutingEvent::ClientMessageReceived { content, src, .. } => {
+                // This check was about checking we knew and client was valid... but even if we don't
+                // we should be handling it...
+                match try_deserialize_handshake(&content, src) {
```
bochaco, line 82: Why do we need this different function to try to deserialise the client msg? we can return an enum from `try_deserialize_msg` so we have a simgle matching statement here?

I think we are trying to deserialise every single client message twice for most of messages, since most of messages are not handshake reqs.?

bochaco: Also, I guess we'll need to move the `HandshakeRequest` to the `sn_messaging` crate?

joshuef: you are correct yep. We are currently deseriallizing twice.

Handshake's haven't lived inside of messages for a while (not sure if there's a root reason). It's been in discussion today actually to get these into sn_messages, which should simplify that (as well as get it handled by routing).

So I'd leave this out of this PR for now, I think

bochaco (approved):

PR #1199, opened 2021-01-08
Rate limit tests

iancoleman:
Some tidy up of duplicate calculations.

And some additional tests, currently set to `ignore` because they don't pass, but thought it might provide a good starting point for a conversation and a kind of place to explore and express the expectations for storecost / ratelimit.


dirvine:
Nice one @iancoleman  thanks for this, those additional tests will prove useful. I am merging this now though. Thanks a million for the help.

PR #1204, opened 2021-01-12
Feat more error enhancements

joshuef:
- Removes UnexpectedNodeError use
- Throws an internal node error when we have no messaging error mapping
- adds more mapping to messaging errors
- Removes client as a dep. Before this, client test running was removed, and the tests themselves were still in place. This removes them. (We should ASAP set up tests against a prebuilt CLI)


src/node/adult_duties/chunks/chunk_storage.rs
```diff
@@ -178,7 +169,10 @@ impl ChunkStorage {
         msg_id: MessageId,
         origin: Address,
     ) -> Result<NodeMessagingDuty> {
-        let result = self.chunks.get(&address).map_err(convert_to_error_message);
+        let result = match self.chunks.get(&address) {
```
Yoga07, line 172: Can use `.map_err(|err|)` here.

src/node/elder_duties/data_section/metadata/map_storage.rs
```diff
@@ -229,9 +229,10 @@ impl MapStorage {
         msg_id: MessageId,
         origin: &MsgSender,
     ) -> Result<NodeMessagingDuty> {
-        let result = self
-            .get_chunk(&address, origin, MapAction::Read)
-            .map_err(convert_to_error_message);
+        let result = match self.get_chunk(&address, origin, MapAction::Read) {
+            Ok(res) => Ok(res),
+            Err(error) => Err(convert_to_error_message(error)?),
```
Yoga07, line 234: Same as above

src/node/elder_duties/data_section/metadata/map_storage.rs
```diff
@@ -253,10 +254,13 @@ impl MapStorage {
...
         origin: &MsgSender,
     ) -> Result<NodeMessagingDuty> {
-        let result = self
+        let result = match self
             .get_chunk(&address, origin, MapAction::Read)
             .map(|data| data.shell())
-            .map_err(convert_to_error_message);
+        {
+            Ok(res) => Ok(res),
+            Err(error) => Err(convert_to_error_message(error)?),
```
Yoga07, line 262: Same as above

dirvine (dismissed):

src/node/elder_duties/data_section/metadata/map_storage.rs
```diff
@@ -278,10 +282,13 @@ impl MapStorage {
...
         origin: &MsgSender,
     ) -> Result<NodeMessagingDuty> {
-        let result = self
+        let result = match self
             .get_chunk(&address, origin, MapAction::Read)
             .map(|data| data.version())
-            .map_err(convert_to_error_message);
+        {
+            Ok(res) => Ok(res),
+            Err(error) => Err(convert_to_error_message(error)?),
```
Yoga07, line 290: Same as above.

Yoga07 (commented):
Just one comment to be applied at multiple places(there are more, other than the mentioned ones) if possible. LGTM otherwise :+1:

src/node/adult_duties/chunks/chunk_storage.rs
joshuef: nope. it was map_err, but we cannot `try` inside there. `?` stopped us :(

bochaco: I don't think we need those helpers to return `Result`, even for the `error => Err(Error::NoErrorMapping(error.to_string())),` we can just return `Error::NoErrorMapping(error.to_string())`, that is the error we will be returning from wherever this conversion is being attempted, isn't this correct?

bochaco: If that is correct, those helpers should return simply `sn_messaging::Error`

bochaco: I'm reviweing this now more carefully, so ignore my previous comments, I understand now what these helpers are for, I'll see if I find a nicer/succinct way for it

joshuef: @Yoga07 has suggest `if let(Err)` on the response to get out of the match.

I was thinking to just return error internally in node. returning the "no mapping" over messages could just lead it to being a new "unexpected" variant, no? but with less info

joshuef: looking at this, `if let Err(e) = res` doesn't relly make things any nicer, we'd be overwriting the `result`'s type anyway... So i think `match` may be simplest way for now

dirvine (approved):

PR #1244, opened 2021-01-25
Re-added missing if condition

Thierry61:
I got an error and an email when I pushed master branch to my fork because condition excluding non-maidsafe repositories wasn't re-added in tag_release.yml (see https://github.com/Thierry61/sn_node/actions/runs/510371179).

The PR corrects this omission.


dirvine (approved):
LGTM @S-Coyle can you check too please?

S-Coyle (approved):
All good from me.
I see `cargo fmt` check is failing after I pushed a change resolving Cargo.lock conflict but this will be because of the new Rust v released today.
I'll merge this and we can fix that in-house.

Thanks @Thierry61 !

PR #1257, opened 2021-01-30
Refactor signing

oetyng:
- Adapts to `sn_routing` changed API where we can no longer access secret keys, and instead pass in data to be signed by `sn_routing` layer.
- Adapts to `sn_routing` API and feature change, where a small number of section keys are cached, and we thus ask to sign data using a secret correspondning to the passed in public key.


joshuef:
Added do not merge until we have client tests stabilised against this

src/node/elder_duties/data_section/mod.rs
```diff
@@ -41,6 +40,66 @@ pub struct DataSection {
...
+        }
+    }
+}
+
+impl Signing for ElderSigning {
+    fn id(&self) -> OwnerType {
+        self.id.clone()
+    }
+
+    fn sign<T: serde::Serialize>(&self, data: &T) -> DtResult<sn_data_types::Signature> {
```
joshuef, line 62: I guess we should probably make this trait `async` for each fn, to have the flexibility to avoid `block_on`.

Not something we need to do for this PR, but to keep in mind for later I think

src/node/elder_duties/key_section/transfers/genesis.rs
```diff
@@ -0,0 +1,63 @@
...
+        msg: "genesis".to_string(),
+    };
+
+    // actor instances' signatures over > credit <
+
+    let mut credit_sig_shares = BTreeMap::new();
+    let credit_sig_share = elder_state.sign_as_elder(&credit).await?;
+    let _ = credit_sig_shares.insert(credit_sig_share.index, credit_sig_share.share);
+
+    println!("Aggregating actor signature..");
```
joshuef, line 28: do we want to keep these? perhaps switch to info?
```suggestion
    info!("Aggregating actor signature..");
```

src/node/elder_duties/key_section/transfers/genesis.rs
```diff
@@ -0,0 +1,63 @@
...
+        actor_signature,
+    };
+
+    // replicas signatures over > signed_credit <
+
+    let mut credit_sig_shares = BTreeMap::new();
+    let credit_sig_share = elder_state.sign_as_elder(&signed_credit).await?;
+    let _ = credit_sig_shares.insert(credit_sig_share.index, credit_sig_share.share);
+
+    println!("Aggregating replica signature..");
```
joshuef, line 49:
```suggestion
    info!("Aggregating replica signature..");
```

joshuef (changes requested):

Yoga07 (dismissed):

joshuef:
looks like cache problems. we need to negate cache on mac I think is the way
`if: matrix.os != 'macOS-latest'`

joshuef (approved):

PR #1261, opened 2021-02-01
feat: give Config public interface

b-zee:
I was looking into running a node from Rust code, but I could not change the member fields of the Config instance.

Perhaps there is a different way to solve this problem, but I'd like to do something similar to [this code](https://github.com/maidsafe/sn_node/blob/c8b60370e3b03b152f85bd6847e3093be1633057/src/tests/mod.rs#L43).


b-zee:
It occurred to me I should have made `set_flag` and `set_value` public. But, these functions do not abstract away from the internals. Actually, with the `set_flag`, code will compile even if the flag you passed has changed out underneath. (But it will error during runtime.)

And so I believe making the struct member fields public is better. If these are used, we ensure code to not compile at all if the (internal) API changes.

joshuef (approved):
seems fair to me

oetyng (approved):
Same here, just a rebase on top of master needed :+1:

PR #1264, opened 2021-02-03
Feat messaging updates

joshuef:


src/node/adult_duties/chunks/mod.rs
```diff
@@ -17,7 +17,7 @@ use crate::{
 use chunk_storage::ChunkStorage;
 use log::{info, trace};
 use sn_data_types::{Blob, BlobAddress};
-use sn_messaging::{
+use sn_messaging::client::{
     Address, DataQuery, Message, MessageId, MsgEnvelope, MsgSender, NodeCmd, NodeDataCmd, Query,
```
oetyng, line 21: I haven't looked at sn_messaging yet, but these have not previously at least been client specific:
`Address, Message, MessageId, MsgEnvelope, MsgSender`

src/node/node_ops.rs
```diff
@@ -488,32 +488,32 @@ pub enum ChunkReplicationCmd {
     StoreReplicatedBlob(Blob),
 }
 
-// impl From<sn_messaging::TransferCmd> for ChunkReplicationCmd {
-//     fn from(cmd: sn_messaging::TransferCmd) -> Self {
+// impl From<sn_messaging::client::TransferCmd> for ChunkReplicationCmd {
```
oetyng, line 491: Maybe this code can be deleted, IDK if it's supposed to come back in? cc: Yogesh

src/node/adult_duties/chunks/mod.rs
joshuef: Yeh, I agree the organisation there could be improved. Atm `client` is a bit catchall for "not routing" i think. @bochaco was working ona  larger refactor there w/ routing stuff, but had to curtaill that to get this stuff in. So I think we can address that as that PR comes in 👍

oetyng: Thought so. Grand!

oetyng (approved):

PR #1283, opened 2021-02-10
Remove msgenvelope

oetyng:


dirvine:
definitely over the 200 line limit, but in this case, there is no choice.

oetyng:
> definitely over the 200 line limit, but in this case, there is no choice.

only by a bit :)

oetyng:
Superseded by #1300

PR #1284, opened 2021-02-10
Remove tiny-keccak unused dependency

iancoleman:


bochaco (approved):
Thanks @iancoleman !, please use https://www.conventionalcommits.org/en/v1.0.0-beta.3/ for upcoming PRs, we are following it in all our repos.

PR #1291, opened 2021-02-11
Reduce Redundant Storage of NodeInfo & fix Incorrect Tracking on Global Used Space

### LINKED ISSUE #1288: Elder Stores Don't Track Global Used Space Properly
**Describe the bug**

`UsedSpace` holds a single internal `Arc` reference to the total used space. The construction `UsedSpace::new()` should only ever be called once because they all need to reference the same total space count. In the construction of `ElderStores` we call `UsedSpace::new()` at least twice (e.g. `Metadata::new()` constructs multiple `UsedSpace`s instead of distributing a single one)., which means the node could potentially consume up to twice the disk space it's allotted.

**Expected behavior**
The total chunk storage should be shared between across chunk stores.

**Sugested Enhancement**

Use `UsedSpace.clone()` when constructing new chunk stores and reset the used space amount on aging up. The later part about resetting on aging up is based on assumption, but if that behavior is incorrect, I can certainly revisit it.

### LINKED ISSUE #1290: `NodeInfo` Stored Redundantly Means it Can't be Modified At Runtime
Here's the [thread](https://forum.safedev.org/t/question-about-large-number-of-copies-of-nodeinfo-struct-in-sn-node/2956) I started originally to address this. Basically, I want to update dynamically something in the `NodeInfo` struct, but it's currently not possible because there's no centralized access pattern for it. Several copies exist around the node, so once the first copy is constructed from the config at node construction, there's no hope of changing any of the fields.

Separate responsibility of `NodeInfo` and `Adult/ElderDuties`. The `Adult/ElderDuties` can hold only information relevent to adults and elders respectively. Then `NodeInfo` should held the non-routing information relevent to all nodes. That way the `NodeDuties` turns into:

```
struct NodeDuties {
    node_info: NodeInfo,
    stage: Stage,
}
```

where stage holds the age-specific information (e.g. either `AdultDuties`, `ElderDuties`, or nothing more for infants). If the `AdultDuties` or `ElderDuties` ever need access to something from `NodeInfo` then it is passed in as a parameter, as they should not have owned copies. The key point is that there is no overlap in state or responsiblity of the `stage` and `node_info` fields.

**Additional context**
As mentioned above, I'm attempting to embed a remote procedure call interface into my node, and it's not possible to change any fields at runtime because the `NodeInfo` is stored in the manner described.

Scorch-Dev:
closes #1290 
fixes #1288 

There's a little bit of interleaving in fixing these two because `UsedSpace` was made a part of `NodeInfo`, so the one issue ended up depending in part on the other.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2393,7 +2393,11 @@ dependencies = [
 
 [[package]]
 name = "sn_node"
+<<<<<<< HEAD
 version = "0.26.8"
+=======
+version = "0.26.9"
+>>>>>>> a39b0043 (fix: unifies used space across node)
 dependencies = [
  "async-log",
  "async-trait",
@@ -2432,7 +2436,6 @@ dependencies = [
  "tempdir",
  "thiserror",
  "threshold_crypto",
- "tiny-keccak 1.5.0",
  "tokio",
  "xor_name",
 ]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1338,9 +1338,9 @@ dependencies = [
 
 [[package]]
 name = "lru_time_cache"
-version = "0.11.6"
+version = "0.11.7"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "a559ce34615abd3145e6ca99df0e33ac069dc5298d1bfecfcc1ca821803f0cb2"
+checksum = "7ce7a913007f8d825242cd5017962f09bc3f19f91a382498c1e6756c842e3908"
 
 [[package]]
 name = "maplit"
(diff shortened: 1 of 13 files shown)


Scorch-Dev:
Agh, cargo just updated to 0.51.0 which changed some behavior of `cargo fmt`, so this PR fails the format check. To not go over the 200 line limit though, I'll wait until the change makes it into upstream and rebase out the formatting issue from there.

joshuef (commented):
LGTM 👍 thanks @Scorch-Dev ! 🙇

joshuef:
Okay, we've ended up moving ahead w/ the clippy prs, so we can get this in w/ a wee rebase. 👍

joshuef (approved):

PR #1292, opened 2021-02-12
chore(clippy): Removes unused Result<T> & applies formatting

### LINKED ISSUE #1293: New Version of `rust-fmt` and `clippy` cause CI to fail
**Describe the bug**
CI currently fails for PRs due to issues caused by formatting and clippy in unrelated Files. This is because the new version of clippy adds lints for unused `Result<T>` and changes the behavior of `cargo fmt`.

**Proposed Solution**
I linked a PR that fixes up clippy and applies the new formatting. It's technically over the line limit, but it's all formatting and lints so I'll leave it there anyway and somebody can always choose to accept or reject it.

Scorch-Dev:
fixes #1293 

There's a new Cargo version, which changes the behavior of `rust-fmt` and adds some lints to `clippy` which cause CI to fail. This patch removes some unused `Result<T>` as well as fixing some general formatting.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2393,7 +2393,7 @@ dependencies = [
 
 [[package]]
 name = "sn_node"
-version = "0.26.8"
+version = "0.26.11"
 dependencies = [
  "async-log",
  "async-trait",
@@ -2432,7 +2432,6 @@ dependencies = [
  "tempdir",
  "thiserror",
  "threshold_crypto",
- "tiny-keccak 1.5.0",
  "tokio",
  "xor_name",
 ]
--- a/src/capacity/rate_limit.rs
+++ b/src/capacity/rate_limit.rs
@@ -80,7 +80,7 @@ impl RateLimit {
 #[cfg(test)]
 mod test {
     use super::*;
-    use crate::Result;
+    use crate::{Error, Result};
     use sn_messaging::client::DataCmd;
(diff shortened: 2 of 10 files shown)


Scorch-Dev:
Strike that, the new clippy version warrants some different changes because of things like unnecessarily wrapping return values in `Result`, even for unit tests. Seems like it's more of a question of modifying the CI, modifying unit tests, or deciding to explicitly ignore it for certain modules. I don't have enough info and if I tried to change all of the unit tests, this would probably go over the 200 line limit, so I'll close this for now.

Scorch-Dev:
Well, it ends up being about ~300 lines of changed code because clippy is now not so happy about absurd test cases some other things that needed to be cleaned up like wrapping infallible functions in `Result`. I got mine to work anyway, so I'll reopen this and somebody can decide whether to accept or reject the fix🤷‍♂️

oetyng:
Thanks for the PR Alan!

Right now though there is a lot of work on messaging parts, things removed etc., and those are things that I see touched here as well. So risks being a small PITA to merge the changes.

I suggest maybe hold off a bit, we're looking to get that work merged asap.

Scorch-Dev:
Sure, that makes sense. Something similar was said on the forums but I saw it a little late. I’ll close this then and sit back until the messaging changes are pulled in and things settle down 👍

PR #1294, opened 2021-02-14
Flag to clear data of previous running node + code improvement

Thierry61:
- Added flag to clear data of previous node running on same PC

- Removed hard coded integral constants like `ARGS[18]` (hardly
maintainable, you have to count elements in ARGS array to check what
an index stands for)

  The situation is better now (no index values to manage anymore) but not
perfect. The problem is that I don't understand why arguments are
analyzed this way. This forces the creation of an array that duplicates
the structure used to define the arguments.

  Normally with structopt we don't need to do that.

- Corrected some clippy and fmt errors.

### DIFF
--- a/src/config_handler.rs
+++ b/src/config_handler.rs
@@ -28,7 +28,7 @@ const CONFIG_FILE: &str = "node.config";
 const CONNECTION_INFO_FILE: &str = "node_connection_info.config";
 const DEFAULT_ROOT_DIR_NAME: &str = "root_dir";
 const DEFAULT_MAX_CAPACITY: u64 = 2 * 1024 * 1024 * 1024;
-const ARGS: [&str; 19] = [
+const ARGS: [&str; 20] = [
     "wallet-id",
     "max-capacity",
     "root-dir",
@@ -48,6 +48,7 @@ const ARGS: [&str; 19] = [
     "local",
     "fresh",
     "clean",
+    "clear-data",
 ];
 
 /// Node configuration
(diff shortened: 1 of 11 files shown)


Thierry61:
What’s funny is that the PR passed the number of lines limit, but some clippy errors remained (in test configuration). So, I corrected them but now the PR exceeds the limit (213/200).

src/config_handler.rs
```diff
@@ -312,6 +322,17 @@ impl Config {
...
+    // Clear data from of a previous node running on the same PC
+    fn clear_data_from_disk(&self) -> Result<()> {
+        if self.clear_data {
+            let path = project_dirs()?.join(self.root_dir()?);
+            if path.exists() {
+                std::fs::remove_dir_all(&path)?;
+            }
+        }
+        Ok(())
+    }
```
lionel1704, line 334: Thus function seems redundant to the one above it.

lionel1704: Ah but I see that this is for the vault data alone. Approving :heavy_check_mark:

lionel1704 (approved):

Yoga07 (approved):

Thierry61:
Yes, this PR is an implementation of what I proposed here: https://safenetforum.org/t/can-anyone-check-if-my-safe-network-is-running/33565/113?u=tfa

joshuef:
Nice one @Thierry61 . That config stuff definitely needs cleaned up, so this is a great start. (I don't think there's any reason it's done that way beyon being legacy code... struct opt is definitely where we should be going with that.)

Thanks! 🙇

PR #1309, opened 2021-02-24
Split section funds

oetyng:


src/network.rs
```diff
@@ -242,17 +241,27 @@ impl Network {
...
+            is_elder = self.routing.lock().await.is_elder().await;
+        }
+
+        let is_adult;
         {
+            is_adult = self
+                .routing
+                .lock()
+                .await
+                .our_adults()
```
maqi, line 256: only an elder knows about all `adults` of the section.
if you are an Adult, `our_adults` returns with empty, and this check will fail.

src/network_state.rs
```diff
@@ -140,6 +141,11 @@ impl ElderState {
     /// https://github.com/rust-lang/rust-clippy/issues?q=is%3Aissue+is%3Aopen+eval_order_dependence
     #[allow(clippy::eval_order_dependence)]
     pub async fn new(network: Network) -> Result<Self> {
+        debug!(
+            ">> setting up elderstate, PK from routing is: {:?}",
+            PublicKey::Bls(network.public_key_set().await?.public_key())
```
maqi, line 146: `network.public_key_set().await?.public_key()` doesn't have Debug implemented?

src/node/adult_duties/chunks/chunk_storage.rs
```diff
@@ -61,7 +61,7 @@ impl ChunkStorage {
                     target_section_pk: None,
                 },
                 dst: DstLocation::EndUser(origin),
-                to_be_aggregated: false, // TODO: to_be_aggregated: true,
+                aggregation: Aggregation::None, // TODO: to_be_aggregated: Aggregation::AtDestination,
```
maqi, line 64: `TODO:  aggregation: Aggregation::AtDestination,` ?
same to do belows.

src/node/elder_duties/data_section/mod.rs
```diff
@@ -148,16 +148,17 @@ impl DataSection {
     pub async fn initiate_elder_change(
         &mut self,
         elder_state: ElderState,
+        sibling_key: Option<PublicKey>,
     ) -> Result<NetworkDuties> {
-        info!("Processing Elder change in data section");
+        info!(">> Processing Elder change in data section");
```
maqi, line 153: just wondering, is this `>>` a new guide line of log formatting?

src/node/elder_duties/data_section/rewards/mod.rs
```diff
@@ -101,16 +101,22 @@ impl Rewards {
...
             },
             dst: DstLocation::Section(section),
-            to_be_aggregated: false,
+            aggregation: Aggregation::None, // TODO: to_be_aggregated: Aggregation::AtDestination,
         })
         .into())
     }
 
     /// After Elder change, we transition to a new
     /// transfer actor, as there is now a new keypair for it.
```
maqi, line 110: better explain what the `sibling_key` is for in the comment ?

src/node/elder_duties/key_section/transfers/mod.rs
```diff
@@ -462,41 +464,30 @@ impl Transfers {
...
-        wallet_id: PublicKey,
+        _wallet_key: PublicKey,
         msg_id: MessageId,
         origin: SrcLocation,
     ) -> Result<NodeMessagingDuty> {
-        info!("Handling GetNewSectionWallet query");
+        info!(">>> Handling GetSectionWalletReplicas query");
         use NodeQueryResponse::*;
         use NodeTransferQueryResponse::*;
         // todo: validate signature
```
maqi, line 480: TODO

src/node/elder_duties/key_section/transfers/mod.rs
```diff
@@ -462,41 +464,30 @@ impl Transfers {
...
-        let result = match self.replicas.history(wallet_id).await {
```
maqi, line 479: so this check is no longer necesssary due to the refactory?

src/node/elder_duties/key_section/transfers/mod.rs
```diff
@@ -575,18 +566,22 @@ impl Transfers {
...
-                    event: NodeEvent::SectionPayoutValidated(event),
-                    id: MessageId::new(),
-                    correlation_id: msg_id,
-                    target_section_pk: None,
-                },
-                dst: origin.to_dst(),
-                to_be_aggregated: false,
-            })),
+            Ok(Some(event)) => {
+                debug!(">>>>> is valid!!!!!!!!!!!!!!!!!!!!!!!!!!!");
```
maqi, line 573: better remove this log in the production code?
or at least change the format?

src/node/elder_duties/key_section/transfers/mod.rs
```diff
@@ -739,12 +737,27 @@ impl Transfers {
                     target_section_pk: None,
                 }
             }
+            Err(Error::InvalidPropagatedTransfer(error)) => {
+                error!(
+                    ">> Error now being handled w/ fake Nosuchkeymsg at receive_propagated, {:?}",
```
maqi, line 742: not quite clear of this log message
`w/` means `with` ?
NoSuchKeyMsg 
what does `at receive_propagated` mean?

src/node/node_duties/elder_constellation.rs
```diff
@@ -67,23 +71,33 @@ impl ElderConstellation {
             return Ok(vec![]);
         }
 
-        info!("Elder change updates initiated");
-
+        info!(">>Elder change updates initiated");
```
maqi, line 74: missing a space after the `>>` ?

src/node/node_duties/elder_constellation.rs
```diff
@@ -55,8 +55,12 @@ impl ElderConstellation {
         &mut self,
         prefix: Prefix,
         new_section_key: PublicKey,
+        sibling_key: Option<PublicKey>,
     ) -> Result<NetworkDuties> {
         let elder_state = self.duties.state();
+        debug!(">> Prefix we have w/ elder change: {:?}", prefix);
+        debug!(">> New section key w/ change {:?}", new_section_key);
```
maqi, line 62: better avoid using `w/` in production code ?

src/node/node_duties/genesis.rs
```diff
@@ -0,0 +1,80 @@
...
+}
+
+impl GenesisAccumulation {
+    pub(crate) fn add(&mut self, sig: SignatureShare) -> Result<()> {
+        let _ = self.signatures.insert(sig.index, sig.share);
+        let min_count = 1 + self.elder_state.public_key_set().threshold();
+        if self.signatures.len() >= min_count {
+            info!("Aggregating replica signature..");
+            // Combine shares to produce the main signature.
+            let debiting_replicas_sig = sn_data_types::Signature::Bls(
```
maqi, line 64: better make signature related accumulation/combination into a separate function to avoid duplication of this block of code?

src/node/node_ops.rs
```diff
@@ -467,9 +469,15 @@ pub enum RewardDuty {
 #[derive(Debug)]
 #[allow(clippy::large_enum_variant)]
 pub enum RewardCmd {
+    // /// Initiates a new SectionActor with the
```
maqi, line 472: comment out the comment ?

src/node/elder_duties/data_section/mod.rs
oetyng: These are debug logs that Josh's been putting in. I don't think they are supposed to be kept when the PR is made ready for review/merge. He's usually very creative with the debug messages.. :)

src/node/elder_duties/key_section/transfers/mod.rs
oetyng: Yeah, you could say that. We're only asking for the pk set of the Replicas of this wallet, and it is not sure there is an history for it (actually, right now this query is sent at a time when we know there is no history).
There's no way to validate at this layer that these are the replicas of this wallet, so what upper layer should really do is to just ask for the replica pk set, and construct the message there.

src/network_state.rs
oetyng: I think this is one of those temporary debug messages that won't follow into ready for review/merge of this PR.

src/node/elder_duties/data_section/mod.rs
oetyng: But actually, IDK if that is some new guide line of log formatting.

src/node/node_duties/genesis.rs
oetyng: Yeah maybe so. Not sure how many occurrences we have of this, and where. Could be a task worthy of a raised issue if you think so.

src/node/node_ops.rs
oetyng: Yeah all of that is commented out as a whole. At any rate, before this is ready for review/merge I'd expect that thing to be deleted if not needed anymore (this specific one is not).

PR #1315, opened 2021-02-26
Add expected aggregation scheme, and use an itinerary

oetyng:
Refactor of messaging to separate aggregation and location scheme, and combine them in an itinerary,


src/node/adult_duties/chunks/chunk_storage.rs
```diff
@@ -61,7 +61,7 @@ impl ChunkStorage {
                     target_section_pk: None,
                 },
                 dst: DstLocation::EndUser(origin),
-                to_be_aggregated: false, // TODO: to_be_aggregated: true,
+                aggregation: Aggregation::None, // TODO: to_be_aggregated: Aggregation::AtDestination,
```
maqi, line 64: `TODO:   aggregation: Aggregation::AtDestination`
same comments as belows.

oetyng: It's intended, because it's a marker for each of these. Not all are to be aggregated at destination, so just the ones that should be, have this comment.

lionel1704:
Superseded by #1317

PR #1317, opened 2021-03-02
feat(Seq): upgrading sn_data_types which makes the Policy of a Sequence data type immutable

bochaco:
BREAKING CHANGE: Policy mutation operations are removed.


davidrusu (dismissed):
LGTM :boat:

joshuef (approved):

PR #1339, opened 2021-03-31
Reward on split

oetyng:
# Rewards and SectionWallet refactor

### Basics

- Payments as usual to a section key using `AT2`.
- Sections collect the `CreditAgreementProofs`, even cross Elder changes.
- At split: Sums the payments at split, tops up with equal amount, distributes to nodes proportional to age.

### Simplifications
- Removes genesis stage.
- Removes the large amounts held at sections.
- Removes the need to transfer amounts between section keys.
- Removes the queueing of reward payouts (as we removed wallet transition).

#### Genesis
The genesis can for example be a hardcoded entry in the network (to check agains when minting), but we do not need to infuse the first section key with that amount.

#### Reward amounts

All payments made to our section are kept until there is a split. At that point we simply distribute the sum to the registered node wallets proportional to their age.
Also, new tokens are minted here by simply doubling the amount credited to the wallets.

#### Payout

The payout happens in the same way as the section wallet transition was supposed to: new elders simply create the credits.

#### Limiting supply

A solution for this has already been designed after discussions in the team, but will come in a follow up PR.


src/section_funds/reward_calc.rs
```diff
@@ -6,57 +6,155 @@
...
+    let mut reward_buckets = BTreeMap::new();
+    for (node_name, (age, wallet)) in nodes {
+        if age >= MIN_REWARD_AGE {
+            let _ = reward_buckets
+                .entry(age)
+                .or_insert_with(BTreeMap::new)
+                .insert(node_name, wallet);
+        }
+    }
+    println!("reward_buckets: {}", reward_buckets.len());
```
lionel1704, line 39: :black_flag:

src/section_funds/reward_calc.rs
```diff
@@ -6,57 +6,155 @@
...
+            } else {
+                let _ = to_return.insert(
+                    *node_name,
+                    (age, *wallet, Token::from_nano(reward_per_wallet)),
+                );
+            }
+        }
     }
+
+    println!("remaining_amount: {}", remaining_amount);
```
lionel1704, line 103: :black_flag:

src/section_funds/reward_calc.rs
```diff
@@ -6,57 +6,155 @@
...
+
+        // calc
+        let rewards = distribute_rewards(amount, nodes);
+
+        // stop timer
+        let duration = now.elapsed();
+
+        println!();
+        println!("Elapsed: {:?} ms", duration.as_millis());
+        println!();
```
lionel1704, line 143: :black_flag:

src/section_funds/reward_stage.rs
```diff
@@ -0,0 +1,229 @@
...
+// permissions and limitations relating to use of the SAFE Network Software.
+
+use crate::{Error, Result};
+use log::{debug, info, warn};
+use sn_data_types::{
+    Credit, CreditAgreementProof, CreditId, PublicKey, ReplicaPublicKeySet, SignatureShare,
+    SignedCredit, SignedCreditShare, Token, TransferPropagated,
+};
+use std::collections::{BTreeMap, BTreeSet};
+
```
lionel1704, line 16: Seems like we are duplicating logic here from the bls_signature_aggregator crate. But it might be required, so I'm not too concerned about it.

lionel1704 (approved):

src/section_funds/reward_stage.rs
oetyng: Yeah, we could try get it in at some point  👍

oetyng:
> 🏴 

Noted! Will address in follow up PR!

PR #1343, opened 2021-04-01
fix(transfers): update transfers on network churn

Yoga07:


oetyng:
Superseeded by #1345 1345

PR #1352, opened 2021-04-01
fix(rewards): sync wallets in a better way

oetyng:
- Only push state belonging to dst.
- Many node wallets were not synced. Bit of sledgehammer now, but syncs a bit better..
- Also removes empty runs of rewarding.


oetyng:
Superseeded by #1356

PR #1354, opened 2021-04-02
fix: use routing 0.57 for qp2p wan fixes

joshuef:


joshuef:
Markign as do not merge until tested and confirmed

Yoga07 (approved):

PR #1366, opened 2021-04-07
fix(config): set connection info when genesis

oetyng:


src/bin/sn_node.rs
```diff
@@ -127,8 +127,12 @@ async fn run_node() {
     );
 
     if config.is_first() {
+        let _ = purge_connection_info().unwrap_or_else(|err| {
+            log::error!("Unable to purge connection info from disk: {}", err);
+            Default::default()
+        });
         let _ = add_connection_info(our_conn_info).unwrap_or_else(|err| {
-            log::error!("Unable to write config to disk: {}", err);
+            log::error!("Unable to write connection info to disk: {}", err);
```
lionel1704, line 135: Can be simplified by calling:
`write_file(CONNECTION_INFO_FILE, vec![our_address]);`
So one write operation instead of two :smile:

lionel1704 (approved):

PR #1370, opened 2021-04-07
feat(joins): enable limiting joins again

oetyng:


src/capacity/rate_limit.rs
```diff
@@ -37,20 +37,20 @@ impl RateLimit {
         RateLimit::rate_limit(bytes, full_nodes, all_nodes, prefix_len)
     }
 
-    ///
+    /// Adds this node to the list o full nodes.
```
maqi, line 40: of

src/capacity/rate_limit.rs
```diff
@@ -37,20 +37,20 @@ impl RateLimit {
...
     pub async fn increase_full_node_count(&mut self, node_id: PublicKey) -> Result<()> {
         self.capacity.increase_full_node_count(node_id).await
     }
 
-    ///
-    #[allow(unused)]
-    pub async fn check_network_storage(&self) -> bool {
-        info!("Checking network storage");
+    /// Returns `true` if more storage capacity
+    /// is required, otherwise `false`.
```
maqi, line 46: no need to line break?

lionel1704 (approved):
Looks good to me

src/event_mapping/mod.rs
```diff
@@ -184,44 +184,23 @@ pub async fn map_routing_event(event: RoutingEvent, network_api: &Network) -> Ma
...
+            ..
+        } => {
+            //self.log_node_counts().await;
+            let op = if let Some(prev_name) = previous_name {
+                trace!("A relocated node has joined the section.");
+                // Switch joins_allowed off a new adult joining.
+                NodeDuty::SetNodeJoinsAllowed(false)
+            } else {
+                //trace!("New node has just joined the network and is a fresh node.",);
+                NodeDuty::NoOp
```
dirvine, line 200: Does this also need to be `NodeDuty::SetNodeJoinsAllowed(false)` unless we are in section null (which routing already handles as it adds till section null is recommended section size)

oetyng: Yeah maybe 🤔

src/transfers/mod.rs
```diff
@@ -140,11 +141,14 @@ impl Transfers {
...
-        })
-        //let mut ops = vec![];
-        //ops.push();
-        //ops.push(Ok(ElderDuty::SwitchNodeJoin(self.rate_limit.check_network_storage().await).into()));
-        //ops
+        });
+
+        ops.push(response);
+        ops.push(NodeDuty::SetNodeJoinsAllowed(
+            self.rate_limit.more_nodes_required().await,
```
maqi, line 148: is this the place that will set the flag back to true ?

dirvine (dismissed):

src/transfers/mod.rs
oetyng: Yes, and would only do so if we are at >50% storage capacity.

oetyng: Not great for spinning up multiple sections fast..

dirvine (approved):

PR #1374, opened 2021-04-07
fix(rewards): improve distribution

oetyng:
- Stretches out the band of distribution further.

### DIFF
--- a/src/section_funds/reward_calc.rs
+++ b/src/section_funds/reward_calc.rs
@@ -6,12 +6,13 @@
 // KIND, either express or implied. Please review the Licences for the specific language governing
 // permissions and limitations relating to use of the SAFE Network Software.
 
+use crate::capacity::MAX_SUPPLY;
 use log::debug;
 use sn_data_types::{NodeAge, PublicKey, Token};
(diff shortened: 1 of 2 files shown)


src/section_funds/reward_calc.rs
```diff
@@ -142,6 +145,48 @@ fn distribute(
...
+    } else if 100_000 >= number {
+        5
+    } else if 10_000 >= number {
+        4
+    } else if 1_000 >= number {
+        3
+    } else if 100 >= number {
+        2
+    } else if 10 >= number {
+        1
```
lionel1704, line 184: I think this can be replaced by `number % 10` ?

lionel1704: I found [this function](https://docs.rs/num-integer/0.1.44/num_integer/fn.nth_root.html) that might be useful

lionel1704: Just realized that 10th root is very wrong :smile: :facepalm: 
Right thing would be `(first multiple of 10 > number) % 10`

lionel1704 (approved):

PR #1383, opened 2021-04-08
chore(deps): update sn_messaging

oetyng:


Cargo.toml
```diff
@@ -31,9 +31,9 @@ serde_json = "1.0.53"
 structopt = "~0.3.17"
 crdts = "4.3.0"
 ed25519-dalek = "1.0.0-pre.4"
-sn_routing = "~0.58.0"
+sn_routing = "~0.58.1"
```
oetyng, line 34: Note: This should not be needed when setting "~0.58".

S-Coyle (approved):

PR #1391, opened 2021-04-12
fix(replication): update holders

oetyng:
- Assumes cmds go through, which follows the general pattern.
- Needs a path for an error returned by adults, whereby the replication needs to be handed over to other nodes.
- Also adds some clarifying errors and simplifies some code.

### DIFF
--- a/src/metadata/blob_register.rs
+++ b/src/metadata/blob_register.rs
@@ -73,40 +73,23 @@ impl BlobRegister {
         // If the data already exist, check the existing no of copies.
         // If no of copies are less then required, then continue with the put request.
         let target_holders = if let Ok(metadata) = self.get_metadata_for(*data.address()).await {
-            if metadata.holders.len() == CHUNK_COPY_COUNT {
-                if data.is_public() {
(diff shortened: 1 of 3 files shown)


src/metadata/blob_register.rs
```diff
@@ -73,40 +73,23 @@ impl BlobRegister {
...
-                            id: MessageId::in_response_to(&msg_id),
-                            correlation_id: msg_id,
-                            target_section_pk: None,
-                        },
-                        section_source: false, // strictly this is not correct, but we don't expect responses to an error..
-                        dst: DstLocation::EndUser(origin),
-                        aggregation: Aggregation::None, // TODO: to_be_aggregated: Aggregation::AtDestination,
-                    }));
-                }
+            if metadata.holders.len() != CHUNK_COPY_COUNT {
```
Yoga07, line 76: It should be `<=` no? 
If holders are lesser than the copy count, then we have top up right?(or that is what the comment above says :) )

oetyng: Yeah, absolutely

Yoga07 (approved):

PR #1399, opened 2021-04-13
fix(sn_node): retry bootstrapping on TryJoinLater error

Yoga07:


lionel1704:
Done in #1398 (killing to save CI time)

PR #1401, opened 2021-04-13
Split Node state between Adult/Elder

davidrusu:
The Node struct was using `Option`s over the fields that were only used when the node was behaving as an Elder vs. behaving as an Adult. This could lead us into inconsistent states where Adults who used to be Elders might still have some Elder state that was mistakenly not cleared (or vice-versa, Elder still behaving as Adult).

The node struct as it was prior to this PR:
```rust
/// Main node struct.
pub struct Node {
    network_api: Network,
    network_events: EventStream,
    node_info: NodeInfo,
    used_space: UsedSpace,
    prefix: Prefix,
    chunks: Option<Chunks>,  // <--- only used by Adults
    meta_data: Option<Metadata>, // <--- only used as Elder
    transfers: Option<Transfers>, // <--- only used as Elder
    section_funds: Option<SectionFunds>, // <--- only used as Elder
}
```
This is changed to the following:
```rust
pub struct Node {
    network_api: Network,
    network_events: EventStream,
    node_info: NodeInfo,
    used_space: UsedSpace,
    prefix: Prefix,
    role: Role, // <-- Role Enum
}
```
Where `Role` is defined as follows:
```rust
struct AdultRole {
    // immutable chunks
    chunks: Chunks,
}

struct ElderRole {
    // data operations
    meta_data: Metadata,
    // transfers
    transfers: Transfers,
    // reward payouts
    section_funds: SectionFunds,
}

enum Role {
    Adult(AdultRole),
    Elder(ElderRole),
}
```

The rest of this PR is just propagating this change throughout the codebase.

### DIFF
--- a/src/node/member_churn.rs
+++ b/src/node/member_churn.rs
@@ -89,13 +89,8 @@ impl Node {
         elder.transfers.merge(user_wallets);
 
         //  merge in provided node reward stages
-        match &mut elder.section_funds {
-            SectionFunds::KeepingNodeWallets { wallets, .. }
-            | SectionFunds::Churning { wallets, .. } => {
-                for (key, (age, wallet)) in &node_wallets {
-                    wallets.set_node_wallet(*key, *age, *wallet);
-                }
-            }
+        for (key, (age, wallet)) in &node_wallets {
+            elder.section_funds.set_node_wallet(*key, *wallet, *age)
         }
 
         let node_id = self.network_api.our_name().await;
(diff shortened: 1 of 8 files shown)


src/node/member_churn.rs
```diff
@@ -34,43 +35,45 @@ impl Node {
...
         Ok(())
     }
 
     /// Level up a newbie to an oldie on promotion
     pub async fn level_up(&mut self) -> Result<()> {
-        //
-        // do not hande immutable chunks anymore
-        self.chunks = None;
-        self.used_space.reset().await;
+        self.used_space.reset().await; // TODO(drusu): should this be part of adult_state?
```
oetyng, line 46: We were talking o this just the other day. Also Elders store data and can fill up. But the whole circuit for measuring and reporting (like adults do) is not there for Elders yet.

oetyng (approved):

PR #1406, opened 2021-04-13
fix(msgs): add aggregation scheme to SendToNodes

oetyng:
The cmd assumed `Aggregation::AtDestination`, since it is mostly used by
Elders. But when replicating chunk, Adults query the holders for the chunk and use the `SendToNodes` cmd.
Therefore it has to contain aggregation scheme so that Adults can set it to None.


oetyng:
Superseded by #1407

PR #1407, opened 2021-04-13
Refactor replication flow

oetyng:
- Refactor of the replication flow:
Adult to Adult connectivity is not guaranteed,
so we have Elders collect the chunks and send them out to new holders.

- Also fixes the `MissingSecretKeyShare` bug:
The cmd assumed `Aggregation::AtDestination`, since it is mostly used by
Elders. But when replicating chunk, Adults query the holders for the chunk and use the `SendToNodes` cmd.
Therefore it has to contain aggregation scheme so that Adults can set it to None.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,9 +31,11 @@ serde_json = "1.0.53"
 structopt = "~0.3.17"
 crdts = "4.3.0"
 ed25519-dalek = "1.0.0-pre.4"
-sn_routing = "~0.59.0"
+# sn_routing = "~0.59.0"
+sn_routing = { git = "https://github.com/oetyng/sn_routing", branch = "refactor-replication-flow" }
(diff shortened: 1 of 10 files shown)


Cargo.toml
```diff
@@ -31,9 +31,11 @@ serde_json = "1.0.53"
 structopt = "~0.3.17"
 crdts = "4.3.0"
 ed25519-dalek = "1.0.0-pre.4"
-sn_routing = "~0.59.0"
+# sn_routing = "~0.59.0"
+sn_routing = { git = "https://github.com/oetyng/sn_routing", branch = "refactor-replication-flow" }
 sn_data_types = "~0.18.0"
-sn_messaging = "~13.1.0"
```
dirvine, line 36: Just update to 14 and we are good here

dirvine (approved):

PR #1415, opened 2021-04-14
refactor(node): processing network events concurrently by spawning to…

oetyng:
…kio tasks

### DIFF
--- a/src/node/interaction.rs
+++ b/src/node/interaction.rs
@@ -6,50 +6,25 @@
 // KIND, either express or implied. Please review the Licences for the specific language governing
 // permissions and limitations relating to use of the SAFE Network Software.
 
+use super::DutyHandler;
 use crate::{
-    node_ops::{NodeDuties, NodeDuty, OutgoingMsg},
-    Node, Result,
+    network::Network,
+    node_ops::{NodeDuty, OutgoingMsg},
+    state::State,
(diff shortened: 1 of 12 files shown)


src/node/interaction.rs
```diff
@@ -91,46 +66,13 @@ impl Node {
...
-            },
-            section_source: false, // sent as single node
-            dst: DstLocation::Section(address),
-            aggregation: Aggregation::None,
-        }
-    }
-
     /// Push our state to the given dst
-    pub fn push_state(&self, prefix: Prefix, msg_id: MessageId) -> NodeDuty {
+    pub(crate) async fn push_state(&self, prefix: Prefix, msg_id: MessageId) -> NodeDuty {
```
davidrusu, line 70: I have some doubts about the correctness of this function in a threaded context.

It takes snapshots of `user_wallets` and `section_wallets` but since each snapshot takes it's own lock, they may not be 100% in sync (say some event comes in that moves some funds from the section wallets to the user wallets before we have a chance to snapshot the section wallets).

I think it'd be safer to move this entire  function into `DutyHandler`

bochaco: yes, I agree

oetyng:
Found that this is not paying out rewards properly.

PR #1416, opened 2021-04-14
refactor(replica): minor changes to TransferStore::get_all,

Dylan-DPC:
conficts with #1415 

* load_wallet doesn't need to be async 
* refactored `get_all` hoping that the compiler optimises the code better and uses iterator that `PickleDB` exposes instead of getting each value separately from the store.


S-Coyle:

PR #1421, opened 2021-04-14
Misc. cleanups

davidrusu:
In this PR:
* UnusedSpace had a funky way of wrapping with Arc<Mutex<..>>, replace that with a more standard approach.
* NodeInfo also had some fields that were not being used, removing those as well
* a few misc. changes.


dirvine (approved):
Very nice clean up

PR #1424, opened 2021-04-14
Added better message to user to suggest why node failed to start

### LINKED ISSUE #1314: Missing --first flag for first node needs better error message. [Good first issue]
**Is your feature request related to a problem? Please describe.**
https://forum.safedev.org/t/latest-sn-node-versions-not-working-routing-network-emptyboostrapnodeslist/2973

A clear message that `--first` is missing.

clD11:
resolves #1314

It is difficult to know where to places this as the error comes from the lib qp2p whic has no notion of --first so at that point it is just an `Routing(Network(UnspecifiedLocalIp))`. 

The config handler also cannot validate as no IP is a valid value and the sn_routing again dosent seem like the correct place. 

This seems like a happy medium for a hint to the user


src/bin/sn_node.rs
```diff
@@ -120,7 +120,8 @@ async fn run_node() {
                     .await;
             }
             Err(e) => {
-                println!("Cannot start node due to error: {:?}. Exiting", e);
+                println!("Cannot start node due to error: {:?}. If this is the first node on the network \
```
lionel1704, line 123: Could you add the same message to the below `error!(....)` call as well?
Many users follow the node log so would be nice to have it there as well.

clD11: I have added that now, I was in two minds to add it last night on whether a 'user hint' should be in the error but your point makes sense

lionel1704 (approved):

clD11:
@lionel1704 Thanks for approving, does this get merged by an authorised user as i dont have authorization

lionel1704:
Was waiting on the CI. Merging now :)

clD11:
@lionel1704 I have looked through some older issues and commented to close where they are no longer relevant due to code changes. Is there anything suitable you would suggest to look at?

PR #1430, opened 2021-04-19
chore(cicd): stop publishing this crate on crates.io as not a lib

S-Coyle:
crates.io is intended for libs, not crates such as this which produce a bin.
Removing publish steps from ci/cd.

Separate to this PR I'll upload to crates.io so that the text there explains that this is no longer published on crates.io


lionel1704 (approved):

S-Coyle:
Decided to close as it was pointed out that there may be future scenarios where this could be used as a lib

PR #1432, opened 2021-04-20
chore(refactor): use atomics for UsedStorage

Dylan-DPC:
- changes mutex to rwlocks 
- some functions do not need to be async due to the change
- gives fixed capacity for the buffers used for bincode so it allocates memory once per need instead of per byte

### DIFF
--- a/src/chunk_store/used_space.rs
+++ b/src/chunk_store/used_space.rs
@@ -7,261 +7,208 @@
 // permissions and limitations relating to use of the SAFE Network Software.
 
 use crate::{Error, Result};
+use std::collections::HashMap;
+use std::sync::atomic::{AtomicU64, Ordering};
 use std::{path::Path, sync::Arc};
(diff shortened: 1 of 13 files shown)


Dylan-DPC:
there's one failing test, i'll get to that later

src/chunk_store/used_space.rs
```diff
@@ -7,261 +7,208 @@
...
+            max_capacity: AtomicU64::new(max_capacity),
+            total_value: AtomicU64::default(),
+            local_stores: Arc::new(RwLock::new(HashMap::<u64, LocalUsedSpace>::new())),
+            next_id: Default::default(),
         }
     }
 
-    /// Clears the entire storage and sets total_value back to zero
-    /// while removing all local stores
+    /// Clears the storage, setting total value ot zero
```
dirvine, line 58: Small type s/ot/to

PR #1433, opened 2021-04-21
feat(data_section): track responsiveness of Adults to data requests so that they can be voted off if necessary

lionel1704:


src/metadata/adult_ops.rs
```diff
@@ -0,0 +1,192 @@
...
+        origin: EndUser,
+        targets: BTreeSet<XorName>,
+    ) -> Result<(), Error> {
+        if let Entry::Vacant(entry) = self.ops.entry(msg_id) {
+            let _ = entry.insert(Operation::Write {
+                blob_write: Box::new(blob_write),
+                origin,
+                targets: targets.clone(),
+            });
+            Ok(())
```
oetyng, line 63: No need to introduce the `AdultOperationInProgress` error, because we have no use of it.

If the entry is vacant, call `self.increment_pending_op(&targets);` directly in that block, otherwise do nothing.

Return a simple `bool` to denote if op was registered or not.

src/chunks/chunk_storage.rs
```diff
@@ -105,7 +115,8 @@ impl ChunkStorage {
                 target_section_pk: None,
             },
             section_source: false, // sent as single node
-            dst: DstLocation::EndUser(origin),
+            // Respond to the chunk's metadata elders
+            dst: DstLocation::Section(*address.name()),
```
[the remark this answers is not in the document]
lionel1704: Yes. The Adults forward it to  Elders who take a note of the adult's response and then forwards it to the client.

src/metadata/adult_ops.rs
lionel1704: Yeah I initially thought of a bool as well. But adding the error seemed to give a bit more clarity as to why it didn't repeat. I guess a command and a log message should do :+1:

PR #1437, opened 2021-04-21
Update new elders on Churn and Splits

### LINKED ISSUE #1429: Reset full_node count on splits and churns
`full_node` count denotes the number of adult nodes in a section who have their storage filled up till `MAX_STORAGE_USAGE_RATIO`. This counter is not reset when a section undergoes split or churns, which flips the `JoinsAllowed` switch in routing. Fixing this will stop allowing more adults to join the network prematurely.

Yoga07:
Also resolves #1429

### DIFF
--- a/src/node/handle.rs
+++ b/src/node/handle.rs
@@ -20,6 +20,7 @@ use sn_messaging::{
     client::{Message, NodeQuery},
     Aggregation, DstLocation, MessageId,
 };
+use sn_routing::ELDER_SIZE;
 use xor_name::XorName;
 
 impl Node {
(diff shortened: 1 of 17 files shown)


src/node/handle.rs
```diff
@@ -15,19 +15,43 @@ use crate::{
...
 impl Node {
     ///
     pub async fn handle(&mut self, duty: NodeDuty) -> Result<NodeDuties> {
         info!("Handling NodeDuty: {:?}", duty);
         match duty {
-            NodeDuty::Genesis => {
-                self.level_up().await?;
+            NodeDuty::Genesis => self.level_up(true).await,
+            NodeDuty::RequestForUpdatingDataAsElder { name, msg_id, .. } => {
+                if name != self.our_name().await && self.role.as_elder()?.is_caught_up {
```
oetyng, line 33: `self.role.as_elder()?.is_caught_up`
A bit suspicious thing here. Will read more further down, but in an EC environment, you can't really say that you are ever "caught up". You'll be from time to time, and eventually and so on. But the moment you are, you could just as well not be.

oetyng: Ok, I see what the intention is. What it wants to say is something like "got the initial sync".

oetyng (approved):

PR #1455, opened 2021-04-22
chore(log): log info about current node role, section prefix, age, and name upon every new routing event

bochaco:
Minor refactor also moving Role to its own file.

Example log entry:
```
[sn_node] INFO 2021-04-22T20:57:51.648325846-03:00 [src/node/mod.rs:127] New RoutingEvent received. Current role: Adult, section prefix: Prefix(10), age: 5, node name: b0e5a6..
```

### DIFF
--- a/src/event_mapping/mod.rs
+++ b/src/event_mapping/mod.rs
@@ -66,6 +66,7 @@ pub async fn map_routing_event(event: RoutingEvent, network_api: &Network) -> Ma
             self_status_change,
             ..
         } => {
+            log_network_stats(network_api).await;
             let first_section = network_api.our_prefix().await.is_empty();
             let first_elder = network_api.our_elder_names().await.len() == 1;
             if first_section && first_elder {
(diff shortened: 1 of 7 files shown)


Yoga07 (commented):
It'd be nice to print the Network Stats whenever a node joins/leaves the network. Should be within the scope of this PR.

lionel1704:
@Yoga07 since Gab isn't in until later could you please add that in so we can merge and deploy this? :)

lionel1704 (dismissed):

src/event_mapping/mod.rs
```diff
@@ -200,3 +205,15 @@ pub async fn map_routing_event(event: RoutingEvent, network_api: &Network) -> Ma
         },
     }
 }
+
+pub async fn log_network_stats(network_api: &Network) {
```
happybeing, line 209: Too late this time, but I have a request that we keep info messages to a single line as this makes parsing logs much easier and means I could add data like the above to vdash in a couple of minutes. When it is spread over multiple lines I have to build a mini state machine to parse it, which takes more time (and allows execution) both to code and test. 

I'll add parsing of the role / age data today, but these network stats will have to wait - but let me know if you are likely to revert to a single line format because then I'll wait. Thanks.

Great work folks, amazing to see how fast you guys are operating.

happybeing: ^^ @Yoga07

Yoga07: Surething @happybeing! Wasn't satisfied with the box formatting myself :) Reverting to single line logging now. PR incoming!

PR #1459, opened 2021-04-23
chore(refactor): add methods to prepare for atomics pr

Dylan-DPC:
since we can't clone the used space anymore  this pr adds extra functions to it and other types and also allows other types to create the user space internally by accepting the max capacity instead of the entire used space.


src/chunk_store/used_space.rs
```diff
@@ -79,6 +84,13 @@ impl UsedSpace {
     pub async fn decrease(&self, id: StoreId, released: u64) -> Result<()> {
         self.inner.lock().await.decrease(id, released).await
     }
+
+    pub fn from_existing_used_space(used_space: &mut UsedSpace) -> UsedSpace {
```
lionel1704, line 88: Perhaps it's good to add a (doc?) comment here saying that the provided used_space will be reset to 0.

lionel1704 (dismissed):
Looks good. Just one comment.

src/chunk_store/used_space.rs
Dylan-DPC: it won't be reset though, it will be cloned from the used space that's passed to it

lionel1704: Right but wouldn't the value of the passed used space become 0?

Dylan-DPC: yeah ah that usage. yes. will add a comment there (still wondering if it's fine to pass an empty used_space for the second one or not)

src/metadata/mod.rs
```diff
@@ -46,16 +46,36 @@ pub struct Metadata {
...
+    pub async fn from_used_space(
+        path: &Path,
+        used_space: &mut UsedSpace,
+        dbs: ChunkHolderDbs,
+        reader: AdultReader,
+    ) -> Result<Self> {
+        let blob_records = BlobRecords::new(dbs, reader);
+        let map_storage = MapStorage::from_used_space(path, used_space).await?;
+        let sequence_storage = SequenceStorage::from_used_space(path, used_space).await?;
+        let register_storage = RegisterStorage::from_used_space(path, used_space).await?;
```
lionel1704, line 78: Related to the other comment. The value of used space would become in L76 so passing the same to L77 and L78 might not be right?

Dylan-DPC: yeah it is, i'm thinking of a way we can get the intended behaviour

Dylan-DPC: going through the code again, the only place it is used the usedspace is reset before creating the `Metadata`, so should be fine. So probably can remove from  the function from `Metadata` and leave it on the other types with a note.

Dylan-DPC: went through the code again, realised we could do it without these functions so removed them :D

lionel1704 (approved):

PR #1467, opened 2021-04-26
feat(chunks): restore reg of liveness with queryresponse

oetyng:
- This requires 1 less msg per query.
- Minor additional cleanup.


lionel1704 (approved):
Nice! :tada:

PR #1470, opened 2021-04-27
fix(e2e): add --for-cli flag to e2e test

Dylan-DPC:


.github/workflows/e2e.yml
```diff
@@ -74,7 +74,7 @@ jobs:
 
       # Our basic test: create an account
       - name: Create Account
-        run: ~/.safe/cli/safe auth create --test-coins
+        run: ~/.safe/cli/safe keys create --test-coins --for-cli
 
       - run: ~/.safe/cli/safe auth unlock --self-auth
```
lionel1704, line 79: This is the line that's failing. Should be okay to remove this since we don't need the auth to be unlocked anymore.

Dylan-DPC: thanks, removed it

lionel1704 (approved):

PR #1482, opened 2021-04-30
fix(error-handling): return a message to the sender when any error was encountered

bochaco:
Attempt to always return a message to the sender if there was an error, in any particular case that's not desired, that step of the logic should catch the error and explicitly return `NoOp`.

### DIFF
--- a/src/chunks/chunk_storage.rs
+++ b/src/chunks/chunk_storage.rs
@@ -46,7 +46,7 @@ impl ChunkStorage {
         origin: EndUser,
     ) -> Result<NodeDuty> {
         let result = if let Err(error) = self.try_store(data, origin).await {
-            Err(CmdError::Data(convert_to_error_message(error)?))
+            Err(CmdError::Data(convert_to_error_message(error)))
         } else {
             Ok(())
         };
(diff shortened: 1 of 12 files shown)


oetyng (approved):
Looks good.
One point about newing up msg ids though: The actual msg ids should be passed there instead, to get correlation. It's done in other places so not sure what is meant by the comment that it can't be done. 
For client there would be the causing msg passed as ctx. But with node msgs we don't do that, so correlation_id = causing msg id is what allows original sender derive context.

Might be worth considering having same pattern all over (i.e. either correlation_id or the actual causing msg sent back, the diff is if source shall keep sent msgs or not).

bochaco:
> One point about newing up msg ids though: The actual msg ids should be passed there instead, to get correlation. It's done in other places so not sure what is meant by the comment that it can't be done.

If you mean bout this comment: https://github.com/maidsafe/sn_node/pull/1482/files#diff-d0661d625e03c3c9a6cc5bb154c5fb69c53b2815eb994041a41ec983be53f899R47 
I have to double check, because that's when the `Message` cannot be parsed and since the msg id is within that struct that's what i mean it's not available. I think that's the only place I am no using the msg id sent for the correlation id, so we may even need to consider moving msg id to the header...

oetyng:
> I have to double check, because that's when the `Message` cannot be parsed and since the msg id is within that struct that's what i mean it's not available. 

Ah, I see.

> I think that's the only place I am no using the msg id sent for filling the correlation id, so we may even need to consider moving msg id to the header...

Yeah, I've thought that as well. I assume routing includes header when aggregating msgs? Because if not, then moving the msg id out of the aggregated payload would cause problems.

Anyway, would be good for AE to get this PR in. So whenever you are up for rebasing and making ready for review.

PR #1486, opened 2021-05-05
Improve logging

oetyng:
Main focus of this work now is to bring some order to the logging.

We want to provide terse and useful information to

**A**. Ourselves during testnets
**B**. Third party UX developers

Currently our logs are a mishmash, unstructured and with few or no conventions.
Our strategy now is to move all those to the `debug` level. (Covered by the commits so far in this PR).

Then we’ll focus on identifying the most appropriate information, format, and conventions for covering **A** on `trace` level, and **B** on `info` level. 
`error` and `warn` would be avoided as far as possible, as that should be covered by successful error handling, and only used as a an ultimate resort.

Additionally, this PR adds system logs. Note that the interval for the logs is perhaps a bit short and we might want it configurable.


happybeing:
Nice to see work on this and the extra system logging. The effect of collapsing everything into one level (if I read the changes correctly) is to increase the amount of information in the logfile and reduce the information available (and the ability to differentiate different kinds of message).

I expected rationalisation but not collapsing more things into one level. I'm not sure if INFO is still there for some things, but there do seem to be useful messages (e.g. about PUTS - "Writing...") which will require setting DEBUG level that were previously available at a lower level of logging.

Can you clarify the aim and effect of this PR? Thanks.

oetyng:
> Nice to see work on this and the extra system logging. The effect of collapsing everything into one level (if I read the changes correctly) is to increase the amount of information in the logfile and reduce the information available (and the ability to differentiate different kinds of message).
> 
> I expected rationalisation but not collapsing more things into one level. I'm not sure if INFO is still there for some things, but there do seem to be useful messages (e.g. about PUTS - "Writing...") which will require setting DEBUG level that were previously available at a lower level of logging.
> 
> Can you clarify the aim and effect of this PR? Thanks.

Sure!

We want to provide terse and useful information to

**A**. Ourselves during testnets
**B**. Third party UX developers

Currently our logs are a mishmash, unstructured and with few or no conventions.
Our strategy now is to move all those to the `debug` level.

Then we’ll focus on identifying the most appropriate information, format, and conventions for covering **A** on `trace` level, and **B** on `info` level. 
`error` and `warn` would be avoided as far as possible, as that should be covered by successful error handling, and only used as a an ultimate resort.

happybeing:
Thanks, makes sense. So this is step one!

Let me know if more input from me would help. I expect you know the kind of thing is like at the info level from my PM though. I was concerned to see useful stuff disappearing!

S-Coyle:

PR #1487, opened 2021-05-05
fix(error-handling): return a message to the sender when any error wa…

oetyng:
…s encountered


oetyng:
I'll merge this to AntiEntropy branch instead.

PR #1492, opened 2021-05-06
fix(section_funds): fix unfinished loop when dropping wallets

lionel1704:


oetyng (dismissed):

src/section_funds/reward_wallets.rs
```diff
@@ -57,7 +57,11 @@ impl RewardWallets {
     /// part of this section, after a split.
     pub fn keep_wallets_of(&self, prefix: Prefix) {
         // Removes keys that are no longer our section responsibility.
-        let keys = self.node_rewards.iter().map(|info| *info.key());
+        let keys = self
```
bochaco, line 60: I think the whole body of this function can simply be: `self.node_rewards.retain(|xorname, _| prefix.matches(&xorname));` ?

lionel1704: Sorry Gabriel. I was on a merge spree. I'll include this change in an upcoming PR :)

PR #1496, opened 2021-05-08
Remove redundant recording of payments

oetyng:
Removes reduntant recording of payments.
Includes some general cleanup.

### DIFF
--- a/src/event_mapping/mod.rs
+++ b/src/event_mapping/mod.rs
@@ -10,7 +10,7 @@ mod map_msg;
 
 use super::node_ops::NodeDuty;
 use crate::network::Network;
-use log::{info, trace};
+use log::{debug, info, trace};
 use map_msg::{map_node_msg, match_user_sent_msg};
 use sn_data_types::PublicKey;
 use sn_messaging::{client::Message, SrcLocation};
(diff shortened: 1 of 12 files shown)


src/node/mod.rs
```diff
@@ -121,13 +121,13 @@ impl Node {
...
-                self.role,
-                self.our_prefix().await,
-                self.network_api.age().await,
-                self.our_name().await
-            );
+            // info!(
+            //     "New RoutingEvent received. Current role: {}, section prefix: {:?}, age: {}, node name: {}",
+            //     self.role,
+            //     self.our_prefix().await,
+            //     self.network_api.age().await,
```
dirvine, line 128: Should we just delete this?

src/event_mapping/mod.rs
```diff
@@ -211,10 +211,18 @@ pub async fn map_routing_event(event: RoutingEvent, network_api: &Network) -> Ma
...
+            "{:?}: {:?} Elders, {:?} Adults.",
+            network_api.our_prefix().await,
+            network_api.our_elder_names().await.len(),
+            network_api.our_adults().await.len()
+        );
+    } else {
+        debug!(
+            "{:?}: {:?} Elders.",
+            network_api.our_prefix().await,
+            network_api.our_elder_names().await.len(),
```
maqi, line 225: Adults shall now knows about the section adult list as well, though may not connect to them.
Any reason why no log log it?

src/node/handle.rs
```diff
@@ -44,7 +46,6 @@ impl Node {
                 newbie,
             } => {
                 if newbie {
-                    info!("Promoted to Elder on Churn");
```
maqi, line 47: this is somehow useful, especially when routing log is switched off.
just to confirm that is really not-needed on sn_node level.
same to the below removals.

src/event_mapping/mod.rs
oetyng: So the list is visible now? Then it can be added back for sure :+1: . Was just that it was misleading when it showed 0.

maqi (approved):

PR #1515, opened 2021-05-18
ci: do not use safe authd

joshuef:


lionel1704:
The `jq` tool will need to be installed before the script is run.

S-Coyle:

PR #1517, opened 2021-05-19
ci: default to prerelease fot github releases

joshuef:
dont't bump  automatically


joshuef:
Closing as this isn't enough w/ self udpate. we need release channels

PR #1519, opened 2021-05-19
feat(client-msgs): adapt to changes to client messages to receive client signature in each message

bochaco:


src/metadata/map_storage.rs
```diff
@@ -181,35 +202,18 @@ impl MapStorage {
         self.ok_or_error(result, msg_id, origin).await
     }
 
-    /// Set Map user permissions.
-    async fn set_user_permissions(
```
Yoga07, line 185: Why are we removing this fn?

Yoga07 (approved):

PR #1530, opened 2021-05-24
chore: minor refactor to adapt to new sn_messaging v24.0.0

bochaco:
BREAKING CHANGE: removing the use of msg::Msg message type which breaks backward compatibility with clients.


bochaco:
Superseded by PR #2524

PR #1534, opened 2021-05-26
Process routing events in parallel while listening for other network events

lionel1704:
closes #1415

### DIFF
--- a/src/node/mod.rs
+++ b/src/node/mod.rs
@@ -23,7 +23,9 @@ use crate::{
     state_db::{get_reward_pk, store_new_reward_keypair},
     Config, Error, Result,
 };
-use crate::{event_mapping::Mapping, node_ops::NodeDuties};
+use crate::node_ops::NodeDuties;
+use futures::{future::BoxFuture, FutureExt};
+use handle::NodeTask;
 use log::{error, warn};
 use rand::rngs::OsRng;
 use role::{AdultRole, Role};
@@ -32,11 +34,13 @@ use sn_messaging::client::ClientMsg;
 use sn_routing::{
     EventStream, {Prefix, XorName},
 };
+use std::sync::Arc;
 use std::{
     fmt::{self, Display, Formatter},
     net::SocketAddr,
     path::{Path, PathBuf},
 };
+use tokio::sync::RwLock;
 use tokio::task::JoinHandle;
 
 /// Static info about the node.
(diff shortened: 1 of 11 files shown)


src/node/mod.rs
```diff
@@ -56,15 +61,14 @@ impl NodeInfo {
...
-    network_events: EventStream,
     node_info: NodeInfo,
     used_space: UsedSpace,
     role: Role,
 }
 
 impl Node {
     /// Initialize a new node.
-    pub async fn new(config: &Config) -> Result<Self> {
+    pub async fn new(config: &Config) -> Result<(Self, EventStream)> {
```
joshuef, line 71: what's the reaon to return this out instead of keep on the node?

src/node/mod.rs
```diff
@@ -116,33 +125,94 @@ impl Node {
         self.network_api.our_prefix().await
     }
 
+    async fn process_routing_event(
+        network_events: Arc<Mutex<EventStream>>,
+        network_api: Network,
+    ) -> Result<NodeTask> {
+        let node_task = if let Some(event) = network_events.lock().await.next().await {
+            // tokio spawn should only be needed around intensive tasks, ie sign/verify
```
joshuef, line 133: why not spawn for every event? 

This reads like we may have missed something we deem 'expensive', or could easily when one of the 'non intensive' tasks are refactored down the line.

src/node/mod.rs
```diff
@@ -116,33 +125,94 @@ impl Node {
...
+                    let tasks = self.handle_and_get_threads(duty, None).await;
+                    threads.extend(tasks.into_iter());
+                }
+                Err(err) => {
+                    error!("Error spawning task for task: {}", err);
+                }
+            }
+            // If we can attain the lock on the mutex it means the previous routing event
+            // has already been processed. So spawn it again
+            if event_lock.try_lock().is_some() {
```
joshuef, line 176: does this mean we're only multithreaded _after_ processing? So any holdup there will holdup all event processing?

(but thereafter the node handles the result of that in a multi-threaded manner?)

src/node/mod.rs
lionel1704: If this is kept in node, we have to borrow the entire node object in order to listen for routing events. Separating them allows us to process node operations and listen for routing events in parallel.

src/node/mod.rs
lionel1704: This comment is from older code :smile: Let me see if it's still relevant

src/node/mod.rs
lionel1704: Not really. This check is to ensure that we have only one thread listening for routing events. So if the completed task is the one that was spawned to listen for routing events (i.e. lock on the Mutex is released) we spawn a task to listen for other routing events.
If the lock on the mutex is still held, it means that there is already a running task listening for a routing event so we don't have to spawn another task to do the same.

src/node/mod.rs
joshuef: Do we want to return it out of making a node though? Is that realistically useful?

I guess what I'msaying is, return it out for node consumer? (to do something with?) Or have internal structs to hold those (and anything we want to pass around on _another_ struct)

joshuef (approved):

joshuef (approved):

joshuef:
Adding DONOTMERGE while investigating some ubuntu oddness 🤔

joshuef:
Oddness was my local-node-folder refusing to be cleaned. leading to different storage space at nodes

PR #1537, opened 2021-05-27
chore(cicd): remove references to auth from e2e.yml

S-Coyle:
This PR also updates the e2e check to not run on auto generated PRs


S-Coyle:
A CI run shows that this PR resolves the auth errors. The actual e2e tests still won't run because of incompatible versions, which will be resolved elsewhere.
Merging.

PR #1589, opened 2021-06-17
fix: node now uses contact config file when no HCC supplied

joshuef:


S-Coyle (approved):

S-Coyle:

PR #1595, opened 2021-06-18
feat(network): add timeout on initializing routing

Yoga07:


src/node/mod.rs
```diff
@@ -42,6 +43,8 @@ use std::{
 use tokio::sync::RwLock;
 use tokio::task::JoinHandle;
 
+const JOINING_TIMEOUT: u64 = 60; // 60 seconds
```
lionel1704, line 46: I'd increase this to about 3 minutes. Since this function will return after completing the resource proof etc.

src/error.rs
```diff
@@ -169,6 +169,9 @@ pub enum Error {
     /// Logic error.
     #[error("Logic error: {0}")]
     Logic(String),
+    /// Timeout errors
```
lionel1704, line 172: Instead of a generic timeout error, I'd make this a specific 
```
#[error("Timeout when trying to join the network")]
JoinTimeout
```
And this can be handled as another special case in `bin/sn_node` with a meaningful message like:
```
Encountered a timeout while trying to join the network. Please try again later.
```

joshuef (approved):

